PR #152, opened 2020-04-01
Removed snapshot dependency on model

### LINKED ISSUE #130: Avoid redundant copies of Model members
[Snapshot Feature](https://github.com/pytorch/serve/pull/111/files/a2a8d17552664161a31b7fa2a54abc9607215039#) introduces a new class in the following file.
`frontend/server/src/main/java/org/pytorch/serve/snapshot/ModelInfo.java`

It is unclear how this is different from `class Model`, it seems redundant? If it's redundant, we should remove it. If it's not redundant, then why are moving `class ModelInfo` under snapshot. Since `class Snaphot` should, in theory, not be concerned about specifics of the model as per the documentation.

This results in duplication of members in memory. This implementation is OK for as long as the number of models is low. The moment we start scaling things the larger memory occupant members will also be duplicated. AFAIK, currently we are not using any data members that occupy large amounts of memory in ModelInfo that may not always be the case.

Instead, if we can use a class that only references members which are already loaded into memory of, say `class Model`, in memory that might be efficient.  The urgency of this problem is unclear, depending on a "typical" usage scenario the impact of this will vary.

dhaniram-kshirsagar:
## Description
This PR takes care of concern raised in #130 i.e. remove snapshot dependency on model object.

Fixes #130 

## Type of change

The changes are internal in nature hence already covered by the existing set of snapshot related UT/ITs.
- Logs
[sanity_test.log](https://github.com/pytorch/serve/files/4588684/sanity_test.log)

## Checklist:

NA
NA
NA


frontend/server/src/main/java/org/pytorch/serve/snapshot/SnapshotManager.java
```diff
@@ -170,12 +144,12 @@ private boolean validate(Snapshot snapshot) throws IOException, InvalidSnapshotE
...
-        Map<String, Map<String, ModelSnapshot>> models = snapshot.getModels();
-        for (Map.Entry<String, Map<String, ModelSnapshot>> modelMap : models.entrySet()) {
+        Map<String, Map<String, JsonObject>> models = snapshot.getModels();
+        for (Map.Entry<String, Map<String, JsonObject>> modelMap : models.entrySet()) {
             String modelName = modelMap.getKey();
-            for (Map.Entry<String, ModelSnapshot> versionModel : modelMap.getValue().entrySet()) {
+            for (Map.Entry<String, JsonObject> versionModel : modelMap.getValue().entrySet()) {
                 String versionId = versionModel.getKey();
-                String marName = versionModel.getValue().getMarName();
+                String marName = versionModel.getValue().get("marName").getAsString();
```
mycpuorg, line 152: move "marName" to a global name

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
```diff
@@ -48,6 +50,28 @@ public Model(ModelArchive modelArchive, int queueSize) {
                         this.modelArchive.getModelName(), this.modelArchive.getModelVersion());
     }
 
+    public JsonObject getModelState(boolean isDefaultVersion) {
+
+        JsonObject modelInfo = new JsonObject();
+        modelInfo.addProperty("defaultVersion", isDefaultVersion);
```
mycpuorg, line 56: In general, keep all these const string names to a global location.
Same here. const strings / prop names to either global location or use getter methods.

frontend/server/src/main/java/org/pytorch/serve/wlm/ModelManager.java
```diff
@@ -71,6 +73,29 @@ public ModelArchive registerModel(String url, String defaultModelName)
                 defaultModelName);
     }
 
+    public void registerAndUpdateModel(String modelName, JsonObject modelInfo)
+            throws ModelException, IOException {
+
+        boolean defaultVersion = modelInfo.get("defaultVersion").getAsBoolean();
```
mycpuorg, line 79: Same here. const strings / prop names to either global location or use getter methods.

frontend/server/src/test/java/org/pytorch/serve/SnapshotTest.java
```diff
@@ -368,11 +368,11 @@ private void updateSnapshot(Properties prop) {
...
         snapshot.setName("snapshot");
         snapshot.setCreated(123456);
-        for (Map.Entry<String, Map<String, ModelSnapshot>> modelMap :
+        for (Map.Entry<String, Map<String, JsonObject>> modelMap :
                 snapshot.getModels().entrySet()) {
-            for (Map.Entry<String, ModelSnapshot> versionModel : modelMap.getValue().entrySet()) {
-                versionModel.getValue().setMinWorkers(4);
-                versionModel.getValue().setMaxWorkers(4);
+            for (Map.Entry<String, JsonObject> versionModel : modelMap.getValue().entrySet()) {
+                versionModel.getValue().addProperty("minWorkers", 4);
```
mycpuorg, line 374: Same here. const strings / prop names to either global location or use getter methods.

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
```diff
@@ -17,6 +19,15 @@
 public class Model {
 
     public static final String DEFAULT_DATA_QUEUE = "DATA_QUEUE";
+
+    public static final String MIN_WORKERS = "minWorkers";
```
mycpuorg, line 23: thanks!

dhaniram-kshirsagar:
> This change looks good to me. It should actually reduce the footprint of memory for each model. Can you please change your commit text to include what was tested etc. If this commit was tested for snapshot test cases (please indicate them if you have done so) I'd like to merge this soon.
> 
> @dhaniram-kshirsagar

Done. Added details per PR template and attached logs.

Issue #118, opened 2020-03-23
Python linting


alexwong:
Python linting fails (which I have as a command in the CI buildspec file). Do we want python linting using pylint, or something else, or just remove it entirely?

[Container] 2020/03/23 19:00:23 Running command pylint -rn --rcfile=./ts/tests/pylintrc ts/.
--
6853 | Using config file /codebuild/output/src833252683/src/github.com/pytorch/serve/ts/tests/pylintrc
6854 | ************* Module ts.protocol.__init__
6855 | C:  1, 0: Trailing newlines (trailing-newlines)
6856 | ************* Module ts.model_loader
6857 | W: 91,15: Catching too general exception Exception (broad-except)
6858 | W: 91, 8: Unused variable 'e' (unused-variable)
6859 | ************* Module ts.torch_handler.text_handler
6860 | C:  1, 0: Missing module docstring (missing-docstring)
6861 | C: 11, 0: Missing class docstring (missing-docstring)
6862 | W: 11, 0: Method 'inference' is abstract in class 'BaseHandler' but is not overridden (abstract-method)
6863 | W: 11, 0: Method 'postprocess' is abstract in class 'BaseHandler' but is not overridden (abstract-method)
6864 | W: 11, 0: Method 'preprocess' is abstract in class 'BaseHandler' but is not overridden (abstract-method)
6865 | C:  4, 0: first party import "from torchtext.data.utils import get_tokenizer" should be placed before "from .base_handler import BaseHandler" (wrong-import-order)
6866 | C:  5, 0: standard import "import re" should be placed before "from torchtext.data.utils import get_tokenizer" (wrong-import-order)
6867 | C:  6, 0: standard import "import string" should be placed before "from torchtext.data.utils import get_tokenizer" (wrong-import-order)
6868 | C:  7, 0: third party import "import torch" should be placed before "from torchtext.data.utils import get_tokenizer" (wrong-import-order)
6869 | C:  8, 0: standard import "import unicodedata" should be placed before "import torch" (wrong-import-order)
6870 | ************* Module ts.torch_handler.vision_handler
6871 | C:  1, 0: Missing module docstring (missing-docstring)
6872 | W:  6, 0: Method 'inference' is abstract in class 'BaseHandler' but is not overridden (abstract-method)
6873 | W:  6, 0: Method 'postprocess' is abstract in class 'BaseHandler' but is not overridden (abstract-method)
[...]

PR #163, opened 2020-04-06
Fix pylint issue

harshbafna:
Address #118 

This PR consists of following

- Changes for errors reported by pylint.
- Updated torchserve_sanity script to use latest version of pylint and execute pylint command
- Unused code cleanup
- UT fixes


ts/model_loader.py
```diff
@@ -23,14 +24,8 @@ class ModelLoaderFactory(object):
     """
 
     @staticmethod
-    def get_model_loader(model_dir):
-        manifest_file = os.path.join(model_dir, "MAR-INF/MANIFEST.json")
-        if os.path.exists(manifest_file):
-            return TsModelLoader()
-        elif os.path.exists(os.path.join(model_dir, "MANIFEST.json")):
-            return LegacyModelLoader()
```
mycpuorg, line 31: Not sure if this is a lint rule we set up.
Not clear why this isn't necessary. Could you please explain?

mycpuorg:
Also this branch is out of date

ts/model_loader.py
harshbafna: The LegacyModelLoader was ported from MMS and is no more in use in TorchServe.

mycpuorg:
The code changes here are recommended from pylint checks. I have tested with a basic run [here](https://github.com/pytorch/serve/pull/256#issuecomment-635011730) with these changes on `staging_0_1_1` . Adding `v0.1.1_tested` label based on this since there isn't specific case to be tested here.

PR #169, opened 2020-04-07
Remove model registered through URL from model store on unregister

### LINKED ISSUE #160: Delete model does not delete .mar file in model store
On deleting a model via the API, the .mar file for it is not removed in the model store folder. As a result if one tries to register the same model again, it fails with an error that file already exists.

harshbafna:
## Description

Remove model registered through URL from model store on unregister.

Fixes #160 

## Type of change

- [x] Bug fix

1) 
 - check the squeezenet1_1.mar file is not already available in model store

 - Register squeezenet model from S3 URL :
https://torchserve.s3.amazonaws.com/mar_files/squeezenet1_1.mar

 - Verify the model archive is downloaded in model_store directory

 - Unregister the squeezenet model
   
 - Verify the model archive is removed from model_store directory
 
2) Register and unregister a model from a mar file already present in model store. Validate that the mar file is not removed from the model store on unregistering the model.

- UT/IT execution results
[frontend_build_log_issue_160.txt](https://github.com/pytorch/serve/files/4444475/frontend_build_log_issue_160.txt)

## Checklist:


frontend/modelarchive/src/main/java/org/pytorch/serve/archive/ModelArchive.java
```diff
@@ -176,6 +176,14 @@ public void validate() throws InvalidModelException {
         }
     }
 
+    public static void removeModel(String modelStore, String marURL) {
```
mycpuorg, line 179: looks good to me

fbbradheintz:
Passed initial test on CPU & GPU. Waiting to move the card to "Tested" column until we're at code freeze for `staging_0_1_1` and I've re-run my test script.

PR #265, opened 2020-04-24
updated object detector handler to use only default cuda device for torchvision version less then 0.6

### LINKED ISSUE #104: Default object detection handler works only on cuda:0 device in GPU machine
Executing pre-trained FasterRCNN/MaskRCNN object detection modes using default object detection handler fails if GPU device other then `cuda:0` is used.

Cuda:0 and CPU devices return similar response with minor floating point differences in the BB. However executing on different GPU devices like `cuda:1`, `cuda:2` etc returns tensor/label/score for a single BB.

We created following ticket on PyTorch forum and still waiting for some guidelines to address this difference in behavior in multi-GPU environment for OD 

https://discuss.pytorch.org/t/pytorch-different-output-on-different-cuda-device-for-fasterrcnn-maskrcnn/71867/3

For the time being the OD default handler is hard coded to use cuda:0 device in GPU environment

harshbafna:
## Description

There was bug in torchvision "0.5.0" due to which the object detector models like faster-rcnn, mask-rcnn etc. did not work on multiple GPU devices. Thus the object detector default handler used "cuda:0" device for execution on GPU instances.

With torchvision "0.6.0" release this issue has been fixed. This PR makes the device selection logic conditional for version "0.5.0" and below. Also made changes to use default cuda device instead of "cuda:0"

Also enhanced sanity script to validate all default handlers.

Fixes #104 

## Type of change

 - Install torchvision 0.5.0
 - register fastrcnn model with 4 workers [should be successful]
 - run inference 4 times [should be successful]
 - check nvidi-smi output, all workers should be created on default cuda device.

 - upgrade torchvision to 0.6.0
 - register fastrcnn model with 4 workers [should be successful]
 - run inference 4 times [should be successful]
 - check nvidi-smi output, all workers should be created on different cuda device.

Logs :

Install from source : 
[install_from_src_104.txt](https://github.com/pytorch/serve/files/4668710/install_from_src_104.txt)

Test Object detector on multiple GPUs (p3.8xlarge) : 
[test_object_detector_gpu_104.log](https://github.com/pytorch/serve/files/4668714/test_object_detector_gpu_104.log)

## Checklist:


ts/torch_handler/object_detector.py
```diff
@@ -17,11 +18,12 @@ def __init__(self):
 
     def initialize(self, ctx):
         super(VisionHandler, self).initialize(ctx)
-        self.initialized = False
-        self.device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
-        self.model.to(self.device)
-        self.model.eval()
-        self.initialized = True
+        if torchvision_version == "0.5.0":
```
fbbradheintz, line 21: My understanding is that this bug will also strike users on TorchVision 0.4. Could we change this condition to demand a version < 0.6?

Thoughts:
* We don't want to do a string comparison, because that would fail if we get to version 0.10.0.
* I've done some digging, and PyTorch & TorchVision version strings are consistently three integers separated by periods. So if we were to split the version string on the periods, and treat the components like ints, that would look something like:

```
if major_version <= 0 and minor_version < 6:
    ...
```

fbbradheintz: (NB: I don't encourage treating versions strings numerically in the general case, but this is a narrow, bounded case with low risk that serves our need here.)

harshbafna: Done.

harshbafna:
> This is great, but as I discussed with @mycpuorg, we need a note in the documentation of the `object_detector` example that TorchVision versions < 0.6.0 won't work on multiple GPUs as expected.
> 
> See also my inline comment on the `torchvision.__version__` condition.

Done.

codebuild/Dockerfile.cpu
```diff
@@ -20,7 +20,7 @@
   RUN update-alternatives --install /usr/bin/python python /usr/bin/python3 1
   RUN update-alternatives --install /usr/local/bin/pip pip /usr/local/bin/pip3 1
   
-  RUN pip install torch torchvision -f https://download.pytorch.org/whl/torch_stable.html
+  RUN pip install torch torchvision torchtext -f https://download.pytorch.org/whl/torch_stable.html
```
mycpuorg, line 23: curious, why is this change here?

codebuild/Dockerfile.gpu
```diff
@@ -20,7 +20,7 @@
   RUN update-alternatives --install /usr/bin/python python /usr/bin/python3 1
   RUN update-alternatives --install /usr/local/bin/pip pip /usr/local/bin/pip3 1
   
-  RUN pip install torch torchvision 
+  RUN pip install torch torchvision torchtext
```
mycpuorg, line 23: Same comment. Docker file changes should not come with this.

torchserve_sanity.sh
```diff
@@ -135,7 +154,43 @@ mkdir -p model_store
 
 start_torchserve
 
-register_model
+# run object detection example
+
+register_model "fastrcnn"
```
mycpuorg, line 159:
```bash
# models
models=("fastrcnn" "vgg11" "resnet18" "waveglow")

# inputs - ensure length matches models
model_inputs=("persons.jpg" "television.jpg" "kitten.jpg" "voice.wav")

for i in ${!models[@]};
do
  m=${models[$i]}
  i=${model_inputs[$i]}
  register_model "$m"
  run_inference "$m" "$i"
  unregister_model "$m"
done


```

maaquib:
@harshbafna When you respond to @mycpuorg questions, please resolve `torchserve_sanity.sh` conflicts as well. Thanks.

codebuild/Dockerfile.cpu
harshbafna: As a part of this PR we added changes in the sanity script for validating the object_detector default handler. Later we thought of adding the changes to validate the remaining default handlers as well. For validating the text_classification default handler we needed torchtext installed in the codebuild docker image.

harshbafna: @mycpuorg 
Coming to think of it, the new regression suite also validates all the default handlers. This looks like a repetition of work and managing the same regression in two different places.

torchserve_sanity.sh
harshbafna: Done.

torchserve_sanity.sh
```diff
@@ -35,7 +53,7 @@ stop_torchserve
 
 start_torchserve
 
-run_inference resnet-18 examples/image_classifier/kitten.jpg
```
mycpuorg, line 38: do you really want to remove this?
Wasn't this to verify snapshot resume?

harshbafna: Done.

PR #283, opened 2020-04-29
Transformers custom handler example (CoLA Dataset)

MFreidank:
## Description

Adding an example for integration of custom handler as described in my blogpost:
https://medium.com/analytics-vidhya/deploy-huggingface-s-bert-to-production-with-pytorch-serve-27b068026d18

with the run_glue example (I'm restricting to the CoLA dataset in this example) given by transformers:
https://github.com/huggingface/transformers/blob/master/examples/run_glue.py

This example runs through in <30 minutes on a single Tesla K80 GPU (this can be further tweaked by training for less epochs, if needed).

Fixes #(issue)
Addresses issues #267 and #249

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Feature/Issue validation/testing

Please describe the tests [UT/IT] that you ran to verify your changes and relevent result summary. Provide instructions so it can be reproduced. 
Please also list any relevant details for your test configuration.

- [ ] Test A
- [ ] Test B

- UT/IT execution results

- Logs

## Checklist:

- [ ] Have you added tests that prove your fix is effective or that this feature works?
- [x] New and existing unit tests pass locally with these changes?
- [x] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the documentation?

PR #302, opened 2020-05-04
Huggingface Transformers generalization

HamidShojanazeri:
## Description
Generalizing the Huggingface transformers handler, to handle different models such as BERT, RoBERTa, XLM, etc. for use cases defined in the AutoModel class such as AutoModelForSequenceClassification, AutoModelForQuestionAnswering, AutoModelForTokenClassification, and AutoModelWithLMHead will be added later. 
Fixes #(issue)
Adding a generalized handler for  Huggingface transformers, issue_267.

## Type of change

Following the procedure in the REAMDE, can reproduce the results. 

## Checklist:


examples/Huggingface_Transformers/Download_Transformer_models.py
```diff
@@ -0,0 +1,103 @@
...
+        model = AutoModelForSequenceClassification.from_pretrained(pretrained_model_name, config=config)
+        tokenizer = AutoTokenizer.from_pretrained(pretrained_model_name,do_lower_case=do_lower_case)
+    elif mode== "question_answering":
+        model = AutoModelForQuestionAnswering.from_pretrained(pretrained_model_name)
+        tokenizer = AutoTokenizer.from_pretrained(pretrained_model_name,do_lower_case=do_lower_case)
+    elif mode== "token_classification":
+        config = AutoConfig.from_pretrained(pretrained_model_name,num_labels=num_labels)
+        model = AutoModelForTokenClassification.from_pretrained(pretrained_model_name)
+        tokenizer = AutoTokenizer.from_pretrained(pretrained_model_name,do_lower_case=do_lower_case)
+
```
zhangguanheng66, line 27: Should we add a else warning here if the mode is not supported?

examples/Huggingface_Transformers/Transformers_handler.py
```diff
@@ -0,0 +1,112 @@
...
+class TransformersSeqClassifierHandler(BaseHandler, ABC):
+    """
+    Transformers text classifier handler class. This handler takes a text (string) and
+    as input and returns the classification text based on the serialized transformers checkpoint.
+    """
+    def __init__(self):
+        super(TransformersSeqClassifierHandler, self).__init__()
+        self.initialized = False
+
+    def initialize(self, ctx):
```
zhangguanheng66, line 23: why not combine `self.initialize` func and constructor together? Do you expect user to initialize the class multiple times?

examples/Huggingface_Transformers/Download_Transformer_models.py
HamidShojanazeri: Thanks for your suggestion, the warning is added now.

examples/Huggingface_Transformers/Transformers_handler.py
HamidShojanazeri: Constructor and self.initialize func are combined.

zhangguanheng66: How about this one?

HamidShojanazeri: Sure, modified as well. Thanks.

chauhang: @zhangguanheng66 The `initialize` method has a special purpose in TorchServe base model handlers. @mycpuorg Can you please describe the details for why it needs to be separate?

examples/Huggingface_Transformers/Transformer_handler_generalized.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class TransformersSeqClassifierHandler(BaseHandler, ABC):
+    """
+    Transformers handler class for sequence, token classification and question answering.
+    """
+    def __init__(self,ctx):
+        super(TransformersSeqClassifierHandler, self).__init__()
+        self.manifest = ctx.manifest
+        properties = ctx.system_properties
```
mycpuorg, line 21: We could be doing all these inside the initialize method instead of the constructor. `initialize` is called with service context which can be very helpful.

examples/Huggingface_Transformers/Transformers_handler.py
```diff
@@ -0,0 +1,106 @@
...
+logger = logging.getLogger(__name__)
+
+
+class TransformersSeqClassifierHandler(BaseHandler, ABC):
+    """
+    Transformers text classifier handler class. This handler takes a text (string) and
+    as input and returns the classification text based on the serialized transformers checkpoint.
+    """
+    def __init__(self, ctx):
+        super(TransformersSeqClassifierHandler, self).__init__()
```
mycpuorg, line 20: Same comment as before.
In general, both initialize and instantiation happen around the same time during model load stage. But the initialize function is called with service context  which gives the handler information provided about the model during registration such as:

- model_name
- model_dir
- manifest
- batch_size
- gpu
- ts.version

So the handlers can make better decisions during initialize than in constructor.

chauhang:
Thanks @HamidShojanazeri. Please work with @MFreidank to integrate PR https://github.com/pytorch/serve/pull/283 so that the combined sol handles the general scenario for BERT better

HamidShojanazeri:
Thanks @chauhang , @zhangguanheng66 and @mycpuorg for helping with this PR.  Sure, thanks  @MFreidank for your great post and I am happy to work together to integrate PR #283.

PR #318, opened 2020-05-08
Removed duplicate snapshot server property

### LINKED ISSUE #154: Spurious NUM_WORKERS entry in snapshots
Repro: Start TorchServe, register models, and scale up with 3 separate commands:

```
torchserve --start --model-store modelstore
curl -X POST "http://localhost:8081/models?url=r18.mar"
curl -v -X PUT "http://localhost:8081/models/r18?min_worker=4"
```

There are (correctly) three config files emitted - one for startup, and one after each Management API call. *All three* have a confusing entry:

```
NUM_WORKERS=8
```

It appears (from looking at the source code) that the `NUM_WORKERS` entry is actually the *default* number of workers per model (when the model is specified on the command line).

This doesn't block launch, but the human-readable labels in the snapshot files should be clearer.

shivamshriwas:
## Description

Addresses concern about snapshot server properties "NUM_WORKERS" which was confusing.
- Removed "NUM_WORKERS" which was unnecessary property duplicate of "default_workers_per_model"

 - "default_workers_per_model"  is the actual config property to set default workers for models which are loaded at startup of torchserve.

Fixes #154 

## Type of change

- [x] Test A
repro steps:
torchserve --start --model-store modelstore
curl -X POST "http://localhost:8081/models?url=r18.mar"
curl -v -X PUT "http://localhost:8081/models/r18?min_worker=4"

Observe Snapshot files generated, has a entry for "default_workers_per_init_model".

- Logs
[issue_154_torchserve_sanity_log.txt](https://github.com/pytorch/serve/files/4600283/issue_154_torchserve_sanity_log.txt)

## Checklist:


docs/configuration.md
```diff
@@ -183,7 +183,7 @@ Most of the following properties are designed for performance tuning. Adjusting
 * `enable_envvars_config`: Enable configuring TorchServe through environment variables. When this option is set to "true", all the static configurations of TorchServe can come through environment variables as well. Default: false
 * `number_of_netty_threads`: number frontend netty thread. Default: number of logical processors available to the JVM.
 * `netty_client_threads`: number of backend netty thread. Default: number of logical processors available to the JVM.
-* `default_workers_per_model`: number of workers to create for each model that loaded at startup time. Default: available GPUs in system or number of logical processors available to the JVM.
+* `default_workers_per_init_model`: number of workers to create for each model that loaded at startup time. Default: available GPUs in system or number of logical processors available to the JVM.
```
vdantu, line 186: Why are we modifying the original configuration? Can't we simply change the snapshot to follow what exists in main configuration?

frontend/server/src/test/resources/snapshots/snapshot4.cfg
```diff
@@ -6,7 +6,7 @@ enable_envvars_config=true
 load_models=noop.mar
 private_key_file=src/test/resources/key.pem
 snapshot_store=FS
-NUM_WORKERS=4
+default_workers_per_init_model=4
```
vdantu, line 9: Can we just call this `default_workers_per_model`?

docs/configuration.md
harshbafna: @vdantu : The actual configuration parameter documented is `default_workers_per_model`. However, the config manager internally uses both `default_workers_per_model` and `NUM_WORKERS`. Further, the snapshots generated is a combination of config properties and state of all registered models. The config manager dumps `NUM_WORKERS` param in the snapshot instead of `default_workers_per_model`.

Thus it makes sense to use only the documented parameter.

We are also renaming the config parameter as it is only applicable for the models which are loaded/initialized at the time of starting `torchserve`. For all other models registered through API, the user needs to provide the number of workers.

vdantu: I was thinking from customers view point. If a customer was using previously released version of TorchServe and they are moving to newer version of TorchServe, would changing this option break their experience or not? Variables used internally can be modified while maintaining backwards compatibility would be my thought here. Changing the customer experience here seems odd. Can't we keep this variable as it is and change the `NUM_WORKERS` to `current_num_workers` or to some **other better name**? Also, `init` in `default_workers_per_init_model` seems redundant. I am open for other views here, but it seems like we are changing UX for something that can be fully kept internal to codebase. Correct me if I am wrong.

harshbafna: The only idea to introduce the `init` keyword here is to indicate that it applies to the startup/initialization models and not to every model.

Just `default_workers_per_model` sounds like it will apply to every model if no worker is provided.

I agree this will impact the UX where the parameter is already being set in config.properties.

chauhang:
For any config related changes, please update corresponding docs as well

shivamshriwas:
> For any config related changes, please update corresponding docs as well

@chauhang already taken care.

frontend/server/src/test/resources/snapshots/snapshot4.cfg
shivamshriwas: done

shivamshriwas:
@vdantu comments have been addressed

Issue #312, opened 2020-05-08
Install from source not working


chauhang:
The install from source is not working on the staging branch, giving ClassNotFoundException

```
(serve) ubuntu@ip-172-31-15-18:~/stage/serve$ pip install -U -e .
Obtaining file:///home/ubuntu/stage/serve
Requirement already satisfied, skipping upgrade: Pillow in /home/ubuntu/anaconda3/lib/python3.6/site-packages (from torchserve==0.0.1b20200507) (5.4.1)
Requirement already satisfied, skipping upgrade: psutil in /home/ubuntu/anaconda3/lib/python3.6/site-packages (from torchserve==0.0.1b20200507) (5.4.5)
Requirement already satisfied, skipping upgrade: future in /home/ubuntu/anaconda3/lib/python3.6/site-packages (from torchserve==0.0.1b20200507) (0.18.2)
Installing collected packages: torchserve
 Found existing installation: torchserve 0.0.1b20200507
  Uninstalling torchserve-0.0.1b20200507:
   Successfully uninstalled torchserve-0.0.1b20200507
 Running setup.py develop for torchserve
Successfully installed torchserve
(serve) ubuntu@ip-172-31-15-18:~/stage/serve$ cd ~/work
(serve) ubuntu@ip-172-31-15-18:~/work$ torchserve --ts-config config.properties
Removing orphan pid file.
(serve) ubuntu@ip-172-31-15-18:~/work$ Error: Could not find or load main class org.pytorch.serve.ModelServer
Caused by: java.lang.ClassNotFoundException: org.pytorch.serve.ModelServer
```

harshbafna:
@chauhang  : Please use `pip install .` instead of `pip install -e .`

Because of -e flag pip skips out the build process which generated the frontend jar for `TorchServe`. I will put a PR with the doc fix.

chauhang:
@harshbafna Similar issue was reported 10 days back on master as well. Can you please clarify why https://github.com/pytorch/serve/issues/269 was closed?

harshbafna:
@chauhang : This got missed out there because the OP of the issue had mixed up multiple environment and tried installed from conda repo as well as from source and created a confusion. And later the user was successfully able to install from conda repo on a fresh environment.

mycpuorg:
Both changes for Linux and MacOS are in staging. Removing Launch Blocker tag. @chauhang @dhanainme @lokeshgupta1975 Time for a merge from `staging_0_1_1` to `master` to pull this in??

[...]

PR #321, opened 2020-05-09
Install torchserve from source

mycpuorg:
This addresses issue #312 updates README
with instructions to install from local path.
It builds and installs the pip whl file

## Description

Fixes #(issue)

## Type of change

- [x] Test A
Installed with `pip install .`

- UT/IT execution results

- Logs

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,12 +106,6 @@ If you plan to develop with TorchServe and change some of the source code, you m
     sudo apt-get install openjdk-11-jdk
     ```
 
-1. Install dependencies
-
-    ```bash
-    pip install psutil future
-    ```
-
 1. Clone the repo
 
     ```bash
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -106,30 +106,24 @@ If you plan to develop with TorchServe and change some of the source code, you m
     sudo apt-get install openjdk-11-jdk
     ```
 
-1. Install dependencies
-
-    ```bash
-    pip install psutil future
```
vdantu, line 112: You might need `psutil` to build. [code reference](https://github.com/pytorch/serve/blob/master/ts/metrics/process_memory_metric.py#L10) . I think the above code reference will be invoked when running torch-serve.

README.md
```diff
@@ -106,30 +106,24 @@ If you plan to develop with TorchServe and change some of the source code, you m
...
     ```bash
     git clone https://github.com/pytorch/serve
     cd serve
     ```
 
-1. Make your changes executable
+1. Install from local path
 
     ```bash
-    pip install -e .
```
vdantu, line 125: Should we just add the manual build step here
```
python setup.py build
```
and then
```
pip install . # or pip install -U -e .
```
the `python setup.py build` should invoke [build in setup.py](https://github.com/pytorch/serve/blob/master/setup.py#L91) which should technically build the frontend and plugins and everything else :) .

mycpuorg: sounds like we should add this to scripts in PR #322 ?

README.md
mycpuorg: This is added as a dep inside setup.py. This way the conda binaries which are built out of pip won't have to install the deps separately.

mycpuorg:
I'm leaning towards abandoning this in favor of PR #322

Issue #319, opened 2020-05-08
Version number reporting for TorchServe


chauhang:
At present TorchServe reports different version from different places which defeats the purpose of having a version number. 

- Version number needs to be consistent across all exposed interfaces - `torchserve -v`, /OPT endpoint and any other place
- Same version should be logged in the logs at on startup
- Version number should be updated with each release
- There should be a way to tag the name of the branch with the version and report that in the logs

PR #333, opened 2020-05-12
embed dependencies in setup.py and fix version

aaronmarkham:
## Description

It would be nice if you could just:
`pip install torchserve`
without having to first do a bunch of other installations....
Also, I found that if we put in torchtext we get sentencepiece so why call it out separately?

Seems like this would help simplify the installation docs - at least for pip.

I also fix the version number. 
Addresses part of #319.

How to test:
I tested this on DLAMI Ubuntu w/ Python 3.6 in a conda environment from the branch root:
```
conda create -n ts_pip_req
conda activate ts_pip_req
pip install .
```
It serves densenet and infers and prints the proper version now.

```
(ts_pip_req) ubuntu@ip-172-31-44-208:~/serve-1$ python
Python 3.6.6 |Anaconda, Inc.| (default, Oct  9 2018, 12:34:16) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import ts
>>> ts.__version__
'0.1.0'
```

Fixes #(issue)
n/a
## Type of change

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -132,7 +132,7 @@ class BuildPlugins(Command):
 if __name__ == '__main__':
     version = detect_model_server_version()
 
-    requirements = ['Pillow', 'psutil', 'future']
+    requirements = ['Pillow', 'psutil', 'future', 'torch', 'torchvision', 'torchtext']
 
     setup(
         name='torchserve',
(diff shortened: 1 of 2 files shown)


lokeshgupta1975:
Closing this PR as we do not want to change the installation instructions now and the version reporting has been fixed in another PR #360

Issue #393, opened 2020-05-24
Integration tests limit the resource usage


chauhang:
Optimize the integration test suite, to do cleanup after each model test. At present, the automation test expects all models to be registered and runs the tests against them. The larger BERT based model examples have high resource requirements, and as more examples are added this will continue to increase. 

In order to limit the resource usage, optimize the test suite for proper tear down of the resources:

- [ ] Unregister model after tests for it are completed
- [ ] Delete the model artifacts to clean up the space

PR #398, opened 2020-05-25
Regression suite enhancement

### LINKED ISSUE #394: Regression test suite cannot be run multiple times consecutively in local installs
Fails on model registration tests as the mar already exists in the model store. So, we can either
1. Have instructions to restart the torchserve in a new model_store folder if running collection from PostMan client.
2. Rename the mar file downloaded with a timestamp so it doesn't clash with currently existing file when model registration API is invoked

shivamshriwas:
## Description

- fixes issue for rerun test suite consecutively.
- dumps serve logs in  /tmp/ts.log for all collections run.
- cleans test suite logs for each run
-  Cleanup after each model test  
-  added response comparison for text based and json (image classification json output) for others and binary output , status code is verified.
- JSON output scores is compared with tolerance factor given in data file. As output scores of model  will have prediction may vary.
- Enhanced the regression tests added for huggingface examples

Fixes #394 #393 

## Type of change

Run regression suite consecutively
- [x] Test A  
Regression suite runs consecutively

- [X] Test B
for model zoo inference test models are unregistered after inference test is completed of the model.
Verify in /tmp/test_exec.log

- UT/IT execution results

- Logs

## Checklist:


test/postman/inference_data.json
```diff
@@ -1,78 +1,182 @@
...
+    {
+        "url":"https://torchserve.s3.amazonaws.com/mar_files/densenet161.mar",
+        "model_name":"densenet161",
+        "worker":1,
+        "synchronous":"true",
+        "file":"../examples/image_classifier/kitten.jpg",
+        "output_type":"json",
+        "expected":[
+            {
+                "tiger_cat":0.4693359136581421
```
chauhang, line 46: @shivamshriwas The model will never return the same exact prediction scores on each attempt - there will be a slight variation. So these checks will fail. A better check will be to just search for the names in the resuling json like tiger_cat. Same comment for all the other 'expected' values that added in this PR

dhaniram-kshirsagar: @chauhang To solve this problem,  "tolerance" parameter has been added to this expected response template [see line 61 below]. However it seems there is not enough documentation around response template specification. @shivamshriwas Can you please add a description around response template parameters.

shivamshriwas: @chauhang These checks will not fail as we are comparing the scores with a tolerance factor in consideration. currently it is set to 1%  tolerance. So scores can be +/- 1% of expected value.
This tolerance value can be tuned if required, but 1% seems to be fine tested the suite with multiple consecutive runs on cpu as well as gpu based ec2

test/postman/inference_api_test_collection.json
```diff
@@ -62,11 +62,68 @@
...
+							"}",
+							"else if (type_response === \"text\") {",
+							"    pm.test(\"Test expected TEXT response\", function () {",
+							"        pm.response.to.have.body(pm.iterationData.get(\"expected\"));",
+							"    });",
+							"",
+							"}",
+							"else",
+							" if (\"json\" === \"json\") {",
+							"    pm.test(\"Test expected JSON response\", function () {",
```
dhaniram-kshirsagar, line 81: It doesn't seem to be generic as it has been written with the current expected response body i.e. for image classification. if in future new response format/json is added then this may require complete change.

test/regression_tests.sh
```diff
@@ -126,6 +129,8 @@ sudo rm -rf $ROOT_DIR && sudo mkdir $ROOT_DIR
 sudo chown -R $USER:$USER $ROOT_DIR
 cd $ROOT_DIR
 
+sudo rm -f $TEST_EXECUTION_LOG_FILE $TS_LOG_FILE
```
dhanainme, line 132: These are test report files that get uploaded to S3 after the test run. Can you please explain why these are being deleted after the test run in AWS Code Build.

**These should NOT be deleted**

test/postman/inference_api_test_collection.json
dhanainme: We seem to be creating a dependency between all the tests using parametrization & validation here. I would prefer that we dont add common validation between tests in a single testcase. 

This creates a tight coupling & maintenance would become very hard.

test/regression_tests.sh
shivamshriwas: These files are not deleted after the run instead deleted before a new run to delete logs of previous run . These files should be deleted as these files should have logs only of latest run and not historic runs in case of consecutive runs.

test/postman/inference_data.json
chauhang: Thanks @shivamshriwas Please document the details as suggested by @dhaniram-kshirsagar

shivamshriwas: @chauhang I I have added required documentation.

test/postman/inference_api_test_collection.json
dhaniram-kshirsagar: This has been addressed now. It is optional to add response content validation and code has been refactored to address maintainability.

dhaniram-kshirsagar:
> There are multiple P0 issues with this PR.
> 
> Especially,
> 
> 1. Deleting reports / logs after the tests. These tests are after all run for these imo.
> 2. Coupling validation between multiple tests. This creates a tight coupling between multiple tests & discourages / confuses new contributors to add test. The most recent example was an iteration for HuggingFace Transformers.

This has been handled now. Please refer to my responses to above comments. Also, huggingface xformer related test cases have been enhanced.

dhaniram-kshirsagar:
> Please address @dhanainme 's comments

Not sure why this PR has not been merged however note that all the concerns raised by Dhana have already been addressed.

@shivamshriwas Please resolve conflict.

shivamshriwas:
Rebased branch with master and created new PR https://github.com/pytorch/serve/pull/467

PR #418, opened 2020-06-01
[WIP]Adding an example that enables torchserve as What-If-Tool's backend

lanpa:
## Description

This PR shows an example that enables users to use torchserve as What-If-Tool's backend.

## Checklist:


maaquib:
@lanpa Thanks for this contribution. Can you also add a short README on how to use this example, something like [this](https://github.com/pytorch/serve/blob/master/examples/image_classifier/README.md)

lxning:
Could you please add README for the example?

lanpa:
@lxning Just updated.

msaroufim (approved):

PR #431, opened 2020-06-03
.circleci: Add initial circleci configuration

seemethere:
Adds initial circleci configuration for building wheels.

Wheels will show up as build artifacts in circleci.

# What does this actually do?

Runs `python setup.py bdist_wheel --release --universal` in CircelCI and then stores the artifacts for release.

Signed-off-by: Eli Uriegas <eliuriegas@fb.com>

### DIFF
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -1,0 +1,6 @@
+FROM ubuntu:18.04 as dev
+RUN apt-get update && apt-get install -y openjdk-11-jdk
+COPY --from=continuumio/miniconda3:4.8.2 /opt/conda /opt/conda
+ENV PATH="/opt/conda/bin/:${PATH}"
+ARG PYTORCH_CHANNEL=pytorch
+RUN conda install -c "${PYTORCH_CHANNEL}" -c powerai -y pytorch torchvision torchtext cpuonly psutil
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,74 @@
+version: 2.1
+
+jobs:
+    build_docker_environment:
+        parameters:
+            pytorch_channel:
+                type: string
+                default: "pytorch"
+        machine:
+            image: ubuntu-1604:201903-01
+        environment:
+            - DOCKER_BUILDKIT=1
+        steps:
+            - checkout
+            - run:
+                name: Building docker image
+                command: |
+                    cat .circleci/Dockerfile | docker build \
+                        -t pytorch/serve-circleci-dev \
+                        --progress=plain \
+                        --build-arg PYTORCH_CHANNEL="<< parameters.pytorch_channel >>" -
+            - run:
+                name: Save docker image to archive
+                command: mkdir /tmp/workspace && docker image save pytorch/serve-circleci-dev -o /tmp/workspace/dev-image.tar
+            - persist_to_workspace:
+                root: /tmp/workspace
+                paths:
+                    - dev-image.tar
+    build_wheel:
+        parameters:
+            build_directory:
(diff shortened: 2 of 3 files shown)


dhanainme:
Can you please add some info about testing that was performed against the circle ci config. Build logs would be very helpful.

seemethere:
@dhanainme Build logs can be found on circleci by clicking the status checks at the end of the PR.

I'll link them here for convenience, but they're typically just found at the end:

* https://app.circleci.com/pipelines/github/pytorch/serve/27/workflows/a4266948-3301-4b89-a0e2-f2b05af21270/jobs/38/steps
* https://app.circleci.com/pipelines/github/pytorch/serve/27/workflows/a4266948-3301-4b89-a0e2-f2b05af21270/jobs/39/steps
* https://app.circleci.com/pipelines/github/pytorch/serve/27/workflows/a4266948-3301-4b89-a0e2-f2b05af21270/jobs/40/steps

If you want a full workflow overview it's here:

* https://app.circleci.com/pipelines/github/pytorch/serve/27/workflows/a4266948-3301-4b89-a0e2-f2b05af21270

And the configuration lists how these are actually built as well, which is just a simple:

```
python setup.py bdist_wheel --release --universal
```

.circleci/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM ubuntu:18.04 as dev
+RUN apt-get update && apt-get install -y openjdk-11-jdk
+COPY --from=continuumio/miniconda3:4.8.2 /opt/conda /opt/conda
+ENV PATH="/opt/conda/bin/:${PATH}"
+ARG PYTORCH_CHANNEL=pytorch
+RUN conda install -c "${PYTORCH_CHANNEL}" -c powerai -y pytorch torchvision torchtext cpuonly psutil
```
chauhang, line 6: @seemethere Will this only run on CPU machines? How are we planning to generate binaries for GPUs?

.circleci/config.yml
```diff
@@ -0,0 +1,74 @@
+version: 2.1
+
+jobs:
+    build_docker_environment:
+        parameters:
+            pytorch_channel:
+                type: string
+                default: "pytorch"
+        machine:
+            image: ubuntu-1604:201903-01
```
mycpuorg, line 10: https://github.com/pytorch/serve/pull/431/files#r435594431
seems to use Ubuntu 18.04 docker image and build on top.
I'm not sure what the `image:` tag is for or what the naming convention used?
`ubuntu-1604:201903-01`

.circleci/config.yml
```diff
@@ -0,0 +1,74 @@
...
+                root: /tmp/workspace
+                paths:
+                    - dev-image.tar
+    build_wheel:
+        parameters:
+            build_directory:
+                type: string
+                default: "."
+        machine:
+            image: ubuntu-1604:201903-01
```
mycpuorg, line 35: same question here

.circleci/config.yml
seemethere: These are related to circleci's base image and although they look similar do not relate to docker images.

see https://circleci.com/docs/2.0/configuration-reference/#available-machine-images

.circleci/Dockerfile
seemethere: Does building with CUDA enabled binaries actually matter for using `torchserve` binaries with GPUs?

I'm unaware if it actually does.

chauhang: You are right, TorchServe binaries are not dependent on Cuda. We just need to verify that the generated binaries work for both CPU and GPU deployments

PR #447, opened 2020-06-10
added check for no objects detected in object detector

### LINKED ISSUE #435: Behavior of fastrcnn object_detector demo with a black image and no predictions
**Setup**
Running fastrcnn object_detector [demo](https://github.com/pytorch/serve/tree/master/examples/object_detector/fast-rcnn), without modifications, on a JPEG image that is completely black.  No objects are detected (as expected).

**Summary of Issue**
An exception occurs based in object_detector.py, in inference(), at the line

```pred_t = [pred_score.index(x) for x in pred_score if x > threshold][-1]```

because pred_score is an empty list.  The exception in the logs is:
```raise Exception("Please provide a custom handler in the model archive.")```

and the JSON returned is

```
{
  "code": 503,
  "type": "InternalServerException",
  "message": "Prediction failed"
}
```

**Expected Behavior**
1) Not a "Prediction failed" indication, that can't be differentiated from other scenarios.  Possibly "no predictions found".
2) Possibly a configurable option on what to do for an empty prediction list.  I've currently worked around this by returning a length-1 of `__background__`, but this may not be the best solution for all models, pre-trained or otherwise.

### PR DESCRIPTION
**Pull request by harshbafna**
## Description

Fixed object detector default handler for "no objects detected" by the model.

Fixes #435 

## Type of change

 - Start torchserve
 - register fastrcnn model from model zoo
 - run inference with a [blank image](https://user-images.githubusercontent.com/26594595/84241556-3fbd6700-ab1d-11ea-9d6b-1e874f318c54.jpg)
 - The request should succeed with response "No object detected"

## Checklist:

### CODE REVIEW

File: ts/torch_handler/object_detector.py [hunk 1]
```diff
@@ -2,13 +2,16 @@
...
 import io
+import logging
 import torch
 from PIL import Image
 from torchvision import transforms
 from torchvision import __version__ as torchvision_version
 from torch.autograd import Variable
 from .vision_handler import VisionHandler
 
+logger = logging.getLogger(__name__)
```
dhaniram-kshirsagar, line 13: Remove unused object.

ts/torch_handler/object_detector.py
```diff
@@ -91,8 +97,9 @@ def handle(data, context):
 
         data = _service.preprocess(data)
         data = _service.inference(data)
-        data = _service.postprocess(data)
-
-        return data
+        if data:
+            return _service.postprocess(data)
+        else:
+            return ["No object detected"]
```
dhanainme, line 103: Wondering if returning an empty list is more appropriate that returning an element a single element with string val.

dhanainme:
Can you also add tests that exercise this code path to confirm that it works.

harshbafna:
> Can you also add tests that exercise this code path to confirm that it works.

Added a test case in the sanity script with a blank image.

ts/torch_handler/object_detector.py
harshbafna: I am open to both ideas. Can remove the if-else block if you want .

maaquib: Agreed with @dhanainme

harshbafna: Done.

PR #462, opened 2020-06-16
Added support for model specific custom requirements.txt in mar file 

### LINKED ISSUE #370: torch-model-archiver - support inference code with external python dependencies
Inference code may depend on arbitrary packages outside of torch projects.

currently I didn't see a way to include these in the .mar file.
I had to do a hack which is to pip install --target to a temp folder, open the .mar and inject these in.

Sagemaker has an elegant solution which is to just include a requiremetns.txt in the zip and it will inject them into the environment (conda? pip install inside the model continer as entry point? I have no idea how they implement it in mms)

would be good to have a simliar support in torchserve

harshbafna:
## Description

Introduced feature to allow the user to supply a model-specific requirements.txt file with the model-archive(mar) file. These python dependencies are installed in the temp dir where the mar file is extracted before loading the model. This path is already added to the PYTHONPATH for the worker.

This approach allows us to keep the model-specific python packages independent from the actual python environment in which TorchServe is running.

Overview of implementation approach :

- Added requirements-file/-r flag in model-archiver
- added config param `install_py_dep_per_model` to control this feature
- updated documentation
- refactored environment setup code for starting python processes to a util class `EnvironmentUtils.java`
- added UT in ModelServerTest `testModelWithCustomPythonDependency` to ensure the feature works fine.
  - The new mar file includes a requirement.txt file that installs `tsv` pip package which is not available by default in python or TorchServe environment. The custom handler imports this newly installed package.
- added negative UT in ModelServerTest `testModelWithInvalidCustomPythonDependency` to ensure correct error message is returned in case python package installation from requirements file fails.
Fixes #370

## Type of change

TBD: Add torchserve-sanity logs from the ubuntu and mac environment.

## Checklist:


dhanainme:
Can you add some more info on testing that was done for this. Specifically what libs were tested / logs

frontend/server/src/main/java/org/pytorch/serve/wlm/ModelManager.java
```diff
@@ -152,6 +158,27 @@ private ModelArchive createModelArchive(
         return archive;
     }
 
+    private void setupModelDependencies(Model model) throws IOException, InterruptedException {
```
dhaniram-kshirsagar, line 161: How about supporting custom package given as tar/zip or some dist file?

harshbafna:
> Can you add some more info on testing that was done for this. Specifically what libs were tested / logs

Have added details in the main comment.

frontend/server/src/main/java/org/pytorch/serve/wlm/ModelManager.java
```diff
@@ -152,6 +158,27 @@ private ModelArchive createModelArchive(
...
+        String requirementsFile =
+                model.getModelArchive().getManifest().getModel().getRequirementsFile();
+
+        if (configManager.getInstallPyDepPerModel() && requirementsFile != null) {
+            Path requirementsFilePath =
+                    Paths.get(model.getModelDir().getAbsolutePath(), requirementsFile);
+
+            String packageInstallCommand =
+                    "pip install -U -t "
```
maaquib, line 170: Does this work if a custom python interpretor has been specified in the config? Wondering if `<python_executable> -p pip -U -t ...` would be more appropriate

harshbafna: Done.

frontend/server/src/main/java/org/pytorch/serve/wlm/ModelManager.java
harshbafna: This is already supported. Updated the documentation accordingly.

chauhang:
@harshbafna Can you please check that only the files related to requirements.txt bundling are included in the PR. I am seeing many additional files changed eg for Metrics, Snapshot that are not related to this PR and should be updated separately.

harshbafna:
> @harshbafna Can you please check that only the files related to requirements.txt bundling are included in the PR. I am seeing many additional files changed eg for Metrics, Snapshot that are not related to this PR and should be updated separately.

@chauhang: The changes in the metric collector are part of code refactoring done to avoid code duplication for getting environment string for executing python commands from frontend layer. This code has been moved to a new utility class named `EnvironmentUtils.java` and is used for
 - creating backend workers
 - probing backend metric collector to log system metrics
 - installing model-specific python packages supplied in requirements.txt with the mar file (this feature)

Also, the changes in `SnapshotManger.java` is also related as the newly added feature throws `InterruptedException` at the time of model registration, which is also called at the time snapshot restore.

frontend/server/src/main/java/org/pytorch/serve/wlm/ModelManager.java
```diff
@@ -152,6 +158,35 @@ private ModelArchive createModelArchive(
...
+                    pythonRuntime
+                            + " -m pip install -U -t "
+                            + model.getModelDir().getAbsolutePath()
+                            + " -r "
+                            + requirementsFilePath; // NOPMD
+
+            String[] envp =
+                    EnvironmentUtils.getEnvString(configManager.getModelServerHome(), null, null);
+            Process process =
+                    Runtime.getRuntime()
```
chauhang, line 181: @harshbafna How are failure cases handled? eg wrong name of package in the requirements.txt or a package that takes very long to download from internet / fails to download. How are those errors handled? Further errors need to be handled gracefully and make sure that rest of TorchServe / workers do not crash.

And for models with multiple worker threads, the download should happen only once

harshbafna: @chauhang 

> How are failure cases handled? eg wrong name of package in the requirements.txt 
Added error handling in case the `pip` install fails. Added [corresponding UT](https://github.com/pytorch/serve/pull/462/files#diff-fd2592adf5a38db5711d59c10dfe36bbR652) as well.

> a package that takes very long to download from internet / fails to download

The model registration request is synchronous and there is no defined timeout. In case the download fails, then the pip command will fail and TorchServe will return an error message to the client.

> And for models with multiple worker threads, the download should happen only once

The installation of the custom packages happens at the time of model registration in a temporary directory where the model archive is extracted. This path is then added to the PYTHONPATH and is passed on to the workers at the time of worker creation.

harshbafna:
> * Add additional automated regression tests for custom handler using a requirements.txt

@chauhang: We can add update the bert model's mar files to use this feature as they have a dependency on `transformers` package. We can take this up with #479.  Thoughts?

dhanainme:
@chauhang , can you please confirm if this can be merged.

PR #432, opened 2020-06-04
Refactor base and default handlers and restructure custom handler doc

shivamshriwas:
## Description
- Enhanced image classifier default handler
    - added support for the setter and getter method to specify topk classes to be returned in the response. [default value for topk is 5]
   - If the model itself has less than 5 classes to classify then the output will now contain scores for all the classes.
   - User can now specify the number of classes to be returned in response via newly added setter api if custom handler is extending from this image classifier handler

-  Code reusability - Mnist and Resnet-152 custom handlers have been modified to demonstrate the use of extending image_classifier default handler. A similar request was there in #414 
    - This enables users to just add their custom preprocessing, postprocessing of inference logic with ease.

- Updated required doc custome_service.md
-  Removed 'examples/text_classification/Transformers_README.md' as its not relevant.

Fixes #411 #381 

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

## Feature/Issue validation/testing

Create mar file for mnist and resnet-152 batch inferencing.
Ctart torchserve and register these models.

- [x] Test A: Run inference request for mnist model
- [x] Test B: Run inference request for resnet-152 model 
- [x] Test C: Custom handler to return top 7 classes from image classifier, create a squeezenet1_1 mar using below custom handler . Run inference  and observe probability of 7 classes returned.
[custom_handler.txt](https://github.com/pytorch/serve/files/4731943/custom_handler.txt)

- Logs

[mnist_custom_handler_log_output.txt](https://github.com/pytorch/serve/files/4731944/mnist_custom_handler_log_output.txt)
[resenet-152-custom_handler_log_output.txt](https://github.com/pytorch/serve/files/4731945/resenet-152-custom_handler_log_output.txt)
[custom_handler_log_output.txt](https://github.com/pytorch/serve/files/4731942/custom_handler_log_output.txt)

## Checklist:

- [ ] Have you added tests that prove your fix is effective or that this feature works?
- [x] New and existing unit tests pass locally with these changes?
- [x] Has code been commented, particularly in hard-to-understand areas?
- [x] Have you made corresponding changes to the documentation?

PR #489, opened 2020-07-01
NameError fixed in default image_classifier.py torch handler

TheRochVoices:
## Description

in image_classifier.py, when mapping of index to label is not defined, we should've done classes[index] instead of classes[i]. 'i' is not defined.

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)

### DIFF
--- a/ts/torch_handler/image_classifier.py
+++ b/ts/torch_handler/image_classifier.py
@@ -70,7 +70,7 @@ class ImageClassifier(VisionHandler):
 
                 results.append(tmp)
             else:
-                results.append({str(classes[i]):str(probs[i])})
+                results.append({str(classes[index]):str(probs[index])})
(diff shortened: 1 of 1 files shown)


ts/torch_handler/image_classifier.py
```diff
@@ -70,7 +70,7 @@ def inference(self, data):
 
                 results.append(tmp)
             else:
-                results.append({str(classes[i]):str(probs[i])})
+                results.append({str(classes[index]):str(probs[index])})
```
dhaniram-kshirsagar, line 73: @TheRochVoices Thanks for this however it has been already taken care in [WIP] PR #432.

PR #508, opened 2020-07-09
Update README.md

jspisak:
We are adding a disclaimer for any shared projects in the pytorch org. Let me know if you have any concerns. btw, do you have a general email address that we can add here? We have opensource@fb.com but i couldnt find the AWS open source analog.


maaquib:
> We have [opensource@fb.com](mailto:opensource@fb.com) but i couldnt find the AWS open source analog.

@jspisak Added AWS torchserve email

PR #458, opened 2020-06-15
Update Model zoo with scripted models

shivamshriwas:
## Description

Updates model zoo for scripted models.

Note: The mar file links for scripted models is not live yet and will be live once the mar files are uploaded and made public.

Fixes #426 

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

## Feature/Issue validation/testing

Please describe the tests [UT/IT] that you ran to verify your changes and relevent result summary. Provide instructions so it can be reproduced. 
Please also list any relevant details for your test configuration.

- [ ] Test A
- [ ] Test B

- UT/IT execution results

- Logs
Sanity logs :
[gpu_sanity_scripted_models.txt](https://github.com/pytorch/serve/files/4904456/gpu_sanity_scripted_models.txt)
[cpu_sanity_scripted_models.logs.txt](https://github.com/pytorch/serve/files/4904457/cpu_sanity_scripted_models.logs.txt)

## Checklist:

- [ ] Have you added tests that prove your fix is effective or that this feature works?
- [x] New and existing unit tests pass locally with these changes?
- [ ] Has code been commented, particularly in hard-to-understand areas?
- [x] Have you made corresponding changes to the documentation?

CI/CD  Failure  is due to older torch  version .  Following are logs with updated torch version.
[sanity_logs.txt](https://github.com/pytorch/serve/files/4961149/sanity_logs.txt)
[sanity_log_gpu.txt](https://github.com/pytorch/serve/files/4961490/sanity_log_gpu.txt)

PR #518, opened 2020-07-10
Fixed - torchscript models not working latest master

### LINKED ISSUE #516: Torchscript models not working with latest master
The latest commit for base_handler has introduced a bug due to which scripted models are not working.
The existing script model related test case was having model file [which is not required] hence sanity test suite had not flagged this issue.

dhaniram-kshirsagar:
## Description

Due to recent PR #444, issue #516 got introduced.
Following were the reasons it didn't get caught -
1. CI doesn't have sanity for trochscripted model
2. Tested fix majorly with external script for gpu and cpu [for both scripted and eager mode models] however end to end testing was done only with eager mode model. This is already getting addressed via in progress PR #458 
3. The UTs have mnist scripted model so got under impression that it is covered however realized that mnist_scripted.mar in UT/IT is for cusotm handler and not default handler!!! 

Fixes #516 

## Type of change

The fix has been tested on gpu and cpu machine. using recently uploaded torchscripted models. Please see attached logs for details.

- Logs
[issue_516_test.txt](https://github.com/pytorch/serve/files/4902523/issue_516_test.txt)

## Checklist:


dhaniram-kshirsagar:
> @dhaniram-kshirsagar Please plan to add some regression tests for this to catch problems in future

@chauhang New test cases for torchscript models have been added via PR #458 hence didn't add any test cases.

ashokei:
@dhaniram-kshirsagar seems file path/existence check is removed; 
you could use "get" on dict with default "" so we can still check file existence if user provides a file https://github.com/pytorch/serve/pull/540

above PR also makes "map_location" single point of setting device type/location. This helps future changes if we add another device.

ts/torch_handler/base_handler.py
```diff
@@ -41,13 +41,9 @@ def initialize(self, ctx):
...
 
-        # model def file
-        model_file = self.manifest['model']['modelFile']
-        model_def_path = os.path.join(model_dir, model_file)
-
         map_location = 'cuda' if torch.cuda.is_available() else 'cpu'
-
-        if os.path.isfile(model_def_path):
+        if 'modelFile' in self.manifest['model']:
+            model_file = self.manifest['model']['modelFile']
```
maaquib, line 46: Use get or default as suggested by @ashokei

dhanainme:
Merged as #549

PR #595, opened 2020-08-07
fixed unbound variable issue while creating binaries from script

harshbafna:
## Description

Fixes #(issue)

## Type of change

- [ ] Test A
- [ ] Test B

- UT/IT execution results

- Logs

## Checklist:


binaries/pip/build_wheels.sh
```diff
@@ -57,7 +53,13 @@ mkdir output
 
 install_java_deps
 
-install_torch_deps
+cd ../../
+
+set +u
```
seemethere, line 58: This doesn't really seem as much of a fix as much as it's just masking over the original issue.

binaries/pip/build_wheels.sh
```diff
@@ -57,7 +53,13 @@ mkdir output
...
+install_torch_deps $1
```
chauhang, line 59: Please update the readme with details of the parameter to pass for $1

chauhang (dismissed):
Please see the comments inline

binaries/pip/build_wheels.sh
harshbafna: the `install_torch_deps` function either installs the default torch packages (compatible to CUDA 10.2). For `CUDA10.1` the function expects `cuda101` to be supplied as an argument. 

Just removing the check to ignore the error in case the variable is not set.

harshbafna: In fact torch packages are not required for building the `torchserve` binaries, hence refactored the code.

binaries/pip/build_wheels.sh
harshbafna: `torch` packages are not required for building the `torchserve` binaries. Refactored the code.

HamidShojanazeri (approved):

maaquib (approved):

PR #630, opened 2020-08-20
Fixed model latency calculation logic

### LINKED ISSUE #636: Model p50, p90 and p99 latency is calculated incorrectly in benchmark-ab script
The benchmark-ab.py(https://github.com/pytorch/serve/blob/master/benchmarks/benchmark-ab.py#L273) script doesn't sort the model latency values before calculating the p50, p90, and p99 latencies.

harshbafna:
## Description

Fixed model latency calculation logic.

Fixes #636 

## Type of change

Executed benchmarking using the updated script on different models.


dhaniram-kshirsagar (approved):

benchmarks/benchmark-ab.py
```diff
@@ -270,6 +270,7 @@ def generate_csv_output():
 
     with open('/tmp/benchmark/predict.txt') as f:
         lines = f.readlines()
+        lines.sort(key=float)
```
maaquib, line 273: Since you already have Pandas imported here, it would be simpler to use the [quantile](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.quantile.html) function instead of all this calculation.

maaquib (changes requested):

maaquib:
@harshbafna Seems like there is some overlap with #666 
Is this still required?

harshbafna:
> #666

Yes, this is still required, as the data needs to be sorted before collecting the Modle's P50, P90, and P99 values.

I wasn't able to spend time replacing the overall logic with Pandas' quantile function. Can we take that up later? As the current implementation works fine.

maaquib (approved):

PR #646, opened 2020-08-25
Add support for file:// url for copying .mar files into model store

shrinath-suresh:
## Description

Current version of torchserve only supports http/https urls. Adding support for file:// in url for copying .mar files into model store locally

Fixes #(issue)

## Type of change

- [x] Test A
- [x] Test B

- UT/IT execution results

- Logs

## Checklist:

### DIFF
--- a/frontend/modelarchive/src/main/java/org/pytorch/serve/archive/ModelArchive.java
+++ b/frontend/modelarchive/src/main/java/org/pytorch/serve/archive/ModelArchive.java
@@ -60,12 +60,12 @@ public class ModelArchive {
             }
             try {
                 if (url.indexOf("file://") != -1) {
-                  String OriginalPath = url.replace("file://", "");
-                  String path = FilenameUtils.getFullPath(OriginalPath);
-                  File source = new File(path, marFileName);
-                  FileUtils.copyFile(source, modelLocation);
+                    String originalPath = url.replace("file://", "");
+                    String path = FilenameUtils.getFullPath(originalPath);
+                    File source = new File(path, marFileName);
+                    FileUtils.copyFile(source, modelLocation);
                 } else {
-                  FileUtils.copyURLToFile(new URL(url), modelLocation);
+                    FileUtils.copyURLToFile(new URL(url), modelLocation);
                 }
             } catch (IOException e) {
                 FileUtils.deleteQuietly(modelLocation);
--- a/frontend/modelarchive/src/main/java/org/pytorch/serve/archive/ModelArchive.java
+++ b/frontend/modelarchive/src/main/java/org/pytorch/serve/archive/ModelArchive.java
@@ -59,14 +59,7 @@ public class ModelArchive {
                 throw new FileAlreadyExistsException(marFileName);
             }
             try {
-                if (url.indexOf("file://") != -1) {
-                    String originalPath = url.replace("file://", "");
-                    String path = FilenameUtils.getFullPath(originalPath);
-                    File source = new File(path, marFileName);
-                    FileUtils.copyFile(source, modelLocation);
-                } else {
-                    FileUtils.copyURLToFile(new URL(url), modelLocation);
-                }
+                FileUtils.copyURLToFile(new URL(url), modelLocation);
             } catch (IOException e) {
                 FileUtils.deleteQuietly(modelLocation);
                 throw new DownloadModelException("Failed to download model from: " + url, e);
--- a/frontend/modelarchive/src/main/java/org/pytorch/serve/archive/ModelArchive.java
+++ b/frontend/modelarchive/src/main/java/org/pytorch/serve/archive/ModelArchive.java
@@ -29,7 +29,7 @@ public class ModelArchive {
     public static final Gson GSON = new GsonBuilder().setPrettyPrinting().create();
 
     private static final Pattern URL_PATTERN =
-            Pattern.compile("http(s)?://.*", Pattern.CASE_INSENSITIVE);
+            Pattern.compile("file?://.*|http(s)?://.*", Pattern.CASE_INSENSITIVE);
 
     private static final String MANIFEST_FILE = "MANIFEST.json";
 
@@ -59,7 +59,14 @@ public class ModelArchive {
                 throw new FileAlreadyExistsException(marFileName);
             }
             try {
-                FileUtils.copyURLToFile(new URL(url), modelLocation);
+                if (url.indexOf("file://") != -1) {
+                  String OriginalPath = url.replace("file://", "");
+                  String path = FilenameUtils.getFullPath(OriginalPath);
+                  File source = new File(path, marFileName);
+                  FileUtils.copyFile(source, modelLocation);
+                } else {
+                  FileUtils.copyURLToFile(new URL(url), modelLocation);
+                }
             } catch (IOException e) {
                 FileUtils.deleteQuietly(modelLocation);
                 throw new DownloadModelException("Failed to download model from: " + url, e);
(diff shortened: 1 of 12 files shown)


shrinath-suresh:
@harshbafna Can you please let us know which package should we use for formatting the java code . (like we use `black` for formatting python code) ?

harshbafna:
> @harshbafna Can you please let us know which package should we use for formatting the java code . (like we use `black` for formatting python code) ?

@shrinath-suresh: We use google java formatter. You can use the following command to format your code :

```
cd frontend
./gradlew fJ
```

shrinath-suresh:
> > @harshbafna Can you please let us know which package should we use for formatting the java code . (like we use `black` for formatting python code) ?
> 
> @shrinath-suresh: We use google java formatter. You can use the following command to format your code :
> 
> ```
> cd frontend
> ./gradlew fJ
> ```

Thank you very much

frontend/modelarchive/src/test/java/org/pytorch/serve/archive/ModelArchiveTest.java
```diff
@@ -91,4 +91,10 @@ public void testMalformURL() throws ModelException, IOException, InterruptedExce
         ModelArchive.downloadModel(
                 modelStore, "https://../model-server/models/squeezenet_v1.1/squeezenet_v1.1.mod");
     }
+
+    @Test(expectedExceptions = DownloadModelException.class)
+    public void testMalformLocalURL() throws ModelException, IOException, InterruptedException {
+        String modelStore = "src/test/resources/models";
+        ModelArchive.downloadModel(modelStore, "file://" + modelStore + "/mnist1.mar");
```
harshbafna, line 98: This test case only tests the "mar file already exists" scenario. Please add test cases for the following use cases:

 - Valid local mar file URL, validate that the mar file correctly gets copied to the model-store.
 - URL path does not exist
 -

harshbafna (changes requested):
- It will be worth to add more test cases in `ModelServerTest.java` for valid and invalid URLs supplied through the register model management API and check that proper error messages are propagated to the client.

- Also, validate that the model-store gets cleaned when the model is unregistered.

Check inline comment for adding more test cases in `ModelArchiveTest.java`

shrinath-suresh:
@harshbafna @chauhang We have added the unit tests to test the file:// change. Please review and let us know your comments.

frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java
```diff
@@ -710,6 +710,38 @@ public void testLoadModelFromURL() throws InterruptedException {
         Assert.assertTrue(new File(configManager.getModelStore(), "squeezenet1_1.mar").exists());
     }
 
+    @Test(
+            alwaysRun = true,
+            dependsOnMethods = {"testModelRegisterWithDefaultWorkers"})
```
harshbafna, line 715: Make the test cases sequential.

frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java
```diff
@@ -710,6 +710,38 @@ public void testLoadModelFromURL() throws InterruptedException {
...
+        String source = configManager.getModelStore() + "/mnist.mar";
+        String destination = parent + "/modelarchive/mnist1.mar";
+        File sourceFile = new File(source);
+        File destinationFile = new File(destination);
+        String fileUrl = "";
+        try {
+            FileUtils.copyFile(sourceFile, destinationFile);
+            fileUrl = "file://" + parent + "/modelarchive/mnist1.mar";
+        } catch (Exception e) {
+            Assert.assertEquals(e.getClass(), IOException.class);
```
harshbafna, line 730: Ideally, we should never encounter this exception. If we do, that means something went wrong. It would be better to add the exception with throws.

frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java
```diff
@@ -710,6 +710,38 @@ public void testLoadModelFromURL() throws InterruptedException {
...
+        }
+        testLoadModel(fileUrl, "mnist1", "1.0");
+        Assert.assertTrue(new File(configManager.getModelStore(), "mnist1.mar").exists());
+        FileUtils.deleteQuietly(destinationFile);
```
harshbafna, line 734: clean up the fileUrl as well.

frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java
```diff
@@ -710,6 +710,38 @@ public void testLoadModelFromURL() throws InterruptedException {
...
+    }
+
+    @Test(
+            alwaysRun = true,
+            dependsOnMethods = {"testModelWithInvalidCustomPythonDependency"})
```
harshbafna, line 739: make this test case dependent on `testLoadModelFromFileURI`.

harshbafna (dismissed):
Please see inline comments. 

The test case for `URL path does not exist` is still missing in both `ModelArchiveTest` and `ModelServerTest`

frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java
shrinath-suresh: @harshbafna In this testcase, we copied mnist.mar from resources folder as mnist1.mar. In the end we have deleted mnist1.mar. fileUrl variable is just a URL like "file://path) .

frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java
shrinath-suresh: > Please see inline comments.
> 
> The test case for `URL path does not exist` is still missing in both `ModelArchiveTest` and `ModelServerTest`

In ModelArchiveTest.java - We already have a test to check invalid URL - testMalformLocalUrl. 
In ModelServerTest.java - we have added a new test case - testModelWithInvalidFileUri

shrinath-suresh:
@harshbafna We have addressed the review comments and uploaded the changes. Please do the needful.

chauhang (approved):

frontend/modelarchive/src/main/java/org/pytorch/serve/archive/ModelArchive.java
```diff
@@ -59,7 +59,14 @@ public static ModelArchive downloadModel(String modelStore, String url)
...
             }
             try {
-                FileUtils.copyURLToFile(new URL(url), modelLocation);
+                if (url.indexOf("file://") != -1) {
+                    String originalPath = url.replace("file://", "");
+                    String path = FilenameUtils.getFullPath(originalPath);
+                    File source = new File(path, marFileName);
+                    FileUtils.copyFile(source, modelLocation);
+                } else {
+                    FileUtils.copyURLToFile(new URL(url), modelLocation);
```
maaquib, line 68: NIT: Can't we just use `file:///<path>` as the url and remove the if-else here?

shrinath-suresh: > NIT: Can't we just use `file:///<path>` as the url and remove the if-else here?

@maaquib  `FileUtils.copyURLToFile` is a two step process. It downloads the file from URL and copies it to the destination. if we pass `file://<path>` as url, step one would fail.

maaquib: [This](https://repl.it/@maaquib/ReasonablePreciousMaps) seems to be working

shrinath-suresh: @maaquib Thanks. Removed the if else block. Just the pattern change is enough for accommodating file://

Song-xx:
it is great if  you can provide a easy example

PR #686, opened 2020-09-17
TorchScript C++ based handler support

### LINKED ISSUE #648: Torchscript C++ based handler
## Is your feature request related to a problem? Please describe.
Add a Torch C++ API based handler support. 

## Describe the solution
This will add support in base handler to invoke Torch C++ based API to load and to do inference on a model.
Since the handler runs in the python process, to invoke C++ API from python, python bindings need to be added.

## Describe alternatives solution
NA

maheshambule:
## Description

- This PR adds support in base handler to invoke the Torch C++ API to load TorchScript model and to do inference on it.
- The model_archiver script is modified to accept torch-api-type (either python or cpp) value. Based on the value of Python or CPP API will be invoked by the handler.
- The registration API is modified to accept 'torch_api_type' parameter. The value provided to the registration API will override the Manifest value. If torch_api_type parameter is not provided Manifest.json value will be used. If Manifest.json does not have 'torchAPIType', by default 'python' will be used.
- The torch_api_type value will be passed to the handler context. It is up to the handler, to use this property.
- The user can extend and write their own custom binding based on the binding code provided in this PR.

TO-DO:
1.  Support for pre and post-processing. This is needed to support Non Tensor outputs such as List, List of Tensors, Dictionary of Tensors. 
2. Currently some of the default handlers for complex models can not work with current C++ binding. Need to provide support for those. For. ex. Text Classifier handler has two Torch script files, one is an architecture file and another one is a vocab file.
3. The handler may ignore the torch_api_type property. And that may confuse the user. Documentation is updated for the same. But need to think of some other solution. This is true for other available features as well such as batch_size or gpu_id.

Fixes #648

## Type of change

Test case is added in sanity suite.

- UT/IT execution results

- Logs

## Checklist:

- [ X] Have you added tests that prove your fix is effective or that this feature works?
- [ X] New and existing unit tests pass locally with these changes?
- [ X] Has code been commented, particularly in hard-to-understand areas?
- [ X] Have you made corresponding changes to the documentation?


maheshambule:
@maaquib, @dhanainme Could you please help in reviewing this PR?

docs/management_api.md
```diff
@@ -22,7 +22,17 @@ Similar to the [Inference API](inference_api.md), the Management API provides a
   * a URI using the HTTP(s) protocol. TorchServe can download .mar files from the Internet.
 * `model_name` - the name of the model; this name will be used as {model_name} in other APIs as part of the path. If this parameter is not present, `modelName` in MANIFEST.json will be used.
 * `handler` - the inference handler entry-point. This value will override `handler` in MANIFEST.json if present. **NOTE: Make sure that the given `handler` is in the `PYTHONPATH`. The format of handler is `module_name:method_name`.**
-* runtime - the runtime for the model custom service code. This value will override runtime in MANIFEST.json if present. The default value is `PYTHON`.
+* `torch_api_type` - the Torch API type to be used by the handler. This value will override `torchAPIType` in MANIFEST.json. If not provided, `torchAPIType` from MANIFEST.json will be used.
```
maheshambule, line 25: We can have a different property name instead of 'torchAPIType'.  Suggestions are welcome.

dhanainme:
Please add 
*  Example in the example folder / Model Zoo
* Regression test suite using the model Zoo

.circleci/scripts/linux_test_sanity.sh
```diff
@@ -9,29 +9,42 @@ MODEL_INPUTS=("examples/object_detector/persons.jpg,docs/images/blank_image.jpg"
  "examples/Huggingface_Transformers/Token_classification_artifacts/sample_text.txt" "examples/Huggingface_Transformers/Seq_classification_artifacts/sample_text.txt")
 HANDLERS=("object_detector" "image_segmenter" "text_classification" "image_classifier" "text_classification" "image_classifier" "image_segmenter" "custom" "custom" "custom")
 
+torch_api_types=("cpp" "python")
+#Here "both" means both cpp and python API are supported by handler
+torch_api_support=("python" "python" "python" "python" "python" "both" "python" "python" "python" "python")
```
chauhang, line 14: @maheshambule Why is "python" being repeated here so many times?

examples/cpp_handler/README.md
```diff
@@ -0,0 +1,82 @@
+#### C++ based handler for TorchScript Model using DenseNet161 image classifier
+
+## How do I invoke C++ API from Python based program?
+Since handler runs in Python worker process, in order to invoke C++ API, you need to provide the Python torch_cpp_python_bindings
```
chauhang, line 4: @maheshambule Can you please explain why we need the Python worker to invoke the C++ API? The whole purpose of introducing support for C++ backend is for performance for fast inference. Why is it not possible to call C++ api directly from the Java FE Rest APIs?

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
```diff
@@ -137,6 +149,23 @@ public void setMaxBatchDelay(int maxBatchDelay) {
...
+
+    public void setTorchAPIType(String torchAPIType) {
+        if (torchAPIType != null) {
+            this.torchAPIType = torchAPIType;
+        } else {
+            String tmpTorchAPIType = this.modelArchive.getManifest().getModel().getTorchAPIType();
+            if (tmpTorchAPIType != null) {
+                this.torchAPIType = tmpTorchAPIType;
+            } else {
+                this.torchAPIType = ConfigManager.getInstance().getTsDefaultTorchAPIType();
```
chauhang, line 164: For customer upgrading existing deployments to new release, the default should be set to Python even when there is no value specified in the config file. Will this handle for upgrade scenarios?

model-archiver/README.md
```diff
@@ -112,6 +112,12 @@ optional arguments:
   -r, -requirements-file
                         Path to requirements.txt file containing a list of model specific python
                         packages to be installed by TorchServe for seamless model serving.
+  --torch-api-type TORCH_API_TYPE
+                        Torch API type ("python" or "cpp") to be used by handler.
+                        This is an optional parameter.
```
chauhang, line 117: Clarify details for the default when no value is specified

ts/torch_handler/base_handler.py
```diff
@@ -101,7 +118,14 @@ def inference(self, data, *args, **kwargs):
         """
         marshalled_data = data.to(self.device)
         with torch.no_grad():
-            results = self.model(marshalled_data, *args, **kwargs)
+            if self.torch_api_type == 'python':
+                results = self.model(marshalled_data, *args, **kwargs)
+            else:
+                if len(kwargs):
+                    raise Exception("Keyword arguments are not supported by CPP API. Use Python API instead.")
```
chauhang, line 125: What are the kwargs used for at present for python side? Can you please clarify, what aspects will not work with C++ due to this with an example?

chauhang (changes requested):
@maheshambule Thanks for getting this work started. In future it will be better to have a design discussion before submitting PRs for a major feature like this. 

Changes are needed to add direct support for invoking C++ backend instead of going through python via pybind. Please run performance comparison benchmark between python and C++ backend for the same models for performance benefits observed with the new solution. Other comments are inline

maheshambule:
@chauhang, Thanks for the review and comments.
We have a design document for the feature here:
https://docs.google.com/document/d/1ojYTtK3XfOWj_fQ3PhgGCcvi7a7ojjzcJTcEmJk4-Hs/edit?usp=sharing

Before working on the feature, we did some experiments and we found that C++ with Python bindings (as implemented by this PR) will not and is not supposed to give any performance boost. We also found that the current Python worker overhead is very minimal and in order to utilize C++ backend to gain a performance boost, we need to consider major changes in frontend as well. All the experiments results are present in the document above.

lxning:
@maheshambule we found load model is expensive in Python. vs in C++ lib. Frontend should directly start C++ based backend as @chauhang comments if the model is C++ based.

msaroufim:
@AshkanAliabadi can I close this PR?

PR #705, opened 2020-09-30
Treat application/x-www-form-urlencoded as binary data

jonathan-conder-sm:
## Description

The frontend currently decodes an `application/x-www-form-urlencoded` parameter as UTF-8 string, then re-encodes it immediately. This results in some binary data being mangled (e.g. `b'\x80'` becomes `b'\xef\xbf\xbd'`).

The [current line](https://github.com/pytorch/serve/blob/caee0f6036d2961c3352b7d9c6a8424ce3b6fcb7/frontend/server/src/main/java/org/pytorch/serve/util/NettyUtils.java#L243) uses [`attribute.getValue()`](https://github.com/netty/netty/blob/8c5b72aaf02e7f349a9972dd9179b449b5a6067b/codec-http/src/main/java/io/netty/handler/codec/http/multipart/MemoryAttribute.java#L65), which implicitly decodes the underlying bytes as a UTF-8 string. However [InputParameter's constructor](https://github.com/pytorch/serve/blob/1f16a7e5989e046f0e5ce5a9f4c73182f63c4573/frontend/server/src/main/java/org/pytorch/serve/util/messages/InputParameter.java#L15) immediately re-encodes the String again.

To illustrate, here is some Java code which does the same thing:
```java
import java.nio.ByteBuffer;

// byte[4] { 65, -96, 0, 0 }
byte[] data = ByteBuffer.allocate(4).putFloat(20).array();

// byte[6] { 65, -17, -65, -67, 0, 0 }
data = (new String(data)).getBytes();
```

The first commit is equivalent to removing the last line, which fixes the issue, and as a bonus removes an unnecessary computation.

The most recent commit fixes a [similar issue](https://github.com/netty/netty/pull/10285) (with a different cause) by bumping Netty to the latest stable version. Feel free to revert this if you'd prefer not to update it for whatever reason. But when you do bump the version, [this workaround](https://github.com/pytorch/serve/blob/c84d3309159844b1c071136a03b9bef257a7bf2e/frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java#L685) can be removed from the tests.

The second most recent commit adds some tests. I added a new model archive to facilitate this. It has a custom handler which just returns the contents of the `data` parameter. For convenience here is the code:
```python
def handle(data, context):
    if data:
    	return [entry["data"] for entry in data]
```

The first test is just a sanity check for the handler, it should pass without my fix. The second test does require the fix though. Since the testing code is quite verbose here is a rough python equivalent:
```python
import requests
import struct

# b'\x00\x01\x02...'
data = struct.pack('256B', *range(256))

def testPredictionsEchoMultipart():
    response = requests.post("http://localhost:8080/predictions/echo", files={"data": data})
    assert response.ok
    assert response.content == data

def testPredictionsEchoNoMultipart():
    response = requests.post("http://localhost:8080/predictions/echo", data={"data": data})
    assert response.ok
    assert response.content == data
```

Fixes #(issue) - Not an existing issue

## Type of change

I ran `torchserve_sanity.sh` and also tested that the issue was fixed for my own models (by replacing `requests.post(..., files=...)` by `requests.post(..., data=...)`).


anj1:
Any update on this? The problem affects me too.

jonathan-conder-sm:
There's another somewhat related issue https://github.com/netty/netty/pull/10285 which could be fixed by bumping the netty version to 4.1.51.Final

frontend/server/src/main/java/org/pytorch/serve/util/NettyUtils.java
```diff
@@ -240,7 +240,7 @@ public static InputParameter getFormData(InterfaceHttpData data) {
             case Attribute:
                 Attribute attribute = (Attribute) data;
                 try {
-                    return new InputParameter(name, attribute.getValue());
+                    return new InputParameter(name, getBytes(attribute.getByteBuf()));
```
harshbafna, line 243: @jonathan-conder-sm: Please add a test case in `ModelServerTest.java` and also provide a detailed example in the PR comment about how this fix works.

harshbafna (changes requested):
@jonathan-conder-sm: Thanks for this PR. Please address the small review comment.

frontend/server/src/main/java/org/pytorch/serve/util/NettyUtils.java
jonathan-conder-sm: I added some tests and will edit the comment shortly

harshbafna (approved):

lxning (approved):

jonathan-conder-sm:
Thanks!

PR #709, opened 2020-10-02
Fixed minor error regarding the Docker port option

alvarobartt:
## Description

Fix minor error in [docker/README.md](https://github.com/pytorch/serve/tree/master/docker#running-torchserve-in-a-production-docker-environment) as the '_Exposing specific ports / volumes between the host & docker env._' section specifies the option `-p8080:p8080 -p8081:8081` and it is not valid as there is an extra 'p' in `-p8080:p8080`, which may be confusing.

So the only fix has been `-p8080:p8080` to `-p8080:8080`.

## Type of change

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -195,7 +195,7 @@ You may want to consider the following aspects / docker options when deploying t
 
 * Exposing specific ports / volumes between the host & docker env.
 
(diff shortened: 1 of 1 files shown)


maaquib (approved):

alvarobartt:
Hi @maaquib, merging seems to be blocked due to the `torch-serve-build-gpu` tests maybe because a resource-allocating error or so, so the merge you tried to do could not be done.

P.S.: This message is just a bump! 👍🏻

harshbafna (approved):

PR #759, opened 2020-10-28
Issues720

lxning:
## Description

Fixes #(issue)

## Type of change

Allow user to configure default batchSize and maxBatchDelay in properties when model is initiated.

- [ ] Test A
- [ ] Test B

- UT/IT execution results

- Logs

## Checklist:


harshbafna (changes requested):
@lxning 

Not sure if it is a good idea to introduce a default batch_size and max_batch_delay parameter in TorchServe, as every model may use a different batch size. 

There can be two more possible approaches for this :

 1. Allow user to provide these parameters as a part of model-archive creation

```
--batch_size 2 --max_batch_delay 50
```

 2. Allow users to specify batch_size and max_batch_delay with the `load_models` parameter. Something like: 

```
load_models=model1:2:50,model2:5:100
```

lxning:
> @lxning
> 
> Not sure if it is a good idea to introduce a default batch_size and max_batch_delay parameter in TorchServe, as every model may use a different batch size.
> 
> There can be two more possible approaches for this :
> 
> 1. Allow user to provide these parameters as a part of model-archive creation
> 
> ```
> --batch_size 2 --max_batch_delay 50
> ```
> 
> 1. Allow users to specify batch_size and max_batch_delay with the `load_models` parameter. Something like:
> 
> ```
> load_models=model1:2:50,model2:5:100
> ```

Yeah, each model may have different batchSize and maxBatchDelay. Ideally, each model's manifest defines all the parameter's default values such as #workers, batchSize and maxBatchDelay. However, the existing implementation uses property file to define default #workers for all of the models. Each model's attributes (such as #workers, batchSize and maxBatchDelay) can be changed via http request. To follow existing implementation pattern, so I choose default value for all the models, instead of only define batchSize and maxBatchDelay in manifest.

spate141:
@harshbafna any update on this?

msaroufim:
@lxning should we go ahead and get this merged? There are a couple of issues asking for configuring batch sizes

Just to clarify which manifest should users change if they want to deploy several models at once each with their own batch size?

msaroufim:
Dupe of #1122

PR #765, opened 2020-11-01
Fix socket.send does not guarantee that all data will be send.

jettify:
## Description

`socket.send` does not guarantee that all data will be send in one message and application should handle partial send [1]. It is probably not a problem for small messagse but in case when `resp = service.predict(msg)` return largish payload may lead to broken data.

[1] https://docs.python.org/3/library/socket.html#socket.socket.send

Fixes #(issue)

## Type of change

Not sure how to test this with CI.

## Checklist:


jettify:
This change should fix nasty issue, any objections to merge it? or something else needs to be done?

maaquib (approved):

dhanainme (approved):

PR #840, opened 2020-12-08
TorchServe Worker optimization

dhanainme:
## TorchServe Worker optimization

Design : https://docs.google.com/document/d/1Lu9Dy8bgop6_CxSfhjFd-t19f4NCFVjur_dyspnrdOA/edit#

### Background

Torchserve currently spawns a separate python process for every PyTorch inference worker. These separate processes when running on GPU follow a shared nothing model hence resulting in OOM after loading a very small number of models. This proposal explores different approaches to model the inference workers as threads / processes to identify the right approach to spawn concurrent workers. 

### Code changes to Frontend / Backend

Moving away from current implementation of spawning a separate python process per worker from Java frontend calls and sharing the model in the Python backend would require significant changes to the python backend.

Currently  to start a new worker and to Scale Up / Scale Down the number of workers we simply invoke a Python process from Java frontend. This python process listens in a specific port (say 9001) for Inference / Load Model commands. We also spawn Worker monitoring threads on Java Frontend to read from stdout / err. 

With the need to share model data between the workers, all the workers of a specific model need to be the child process of a single Python Parent Process. We call this WorkerServiceManager. This WorkerServiceManager would be responsible for loading the model in shared memory, spawning child worker processes and sharing the shared model reference with them. 

Upon registration of a model for the first time, we would spawn an instance of WorkerServerManager which would listen on a specific port and accept requests for Load Model, Scale Up & ScaleDown Workers. The child process spawned by this WorkerServerManager - WorkerServer would listen on its own port for inference requests. 

Also given the WorkerServer does not share parent - child relationship with the Java frontend as before, The stdout / stderr of the WorkerServer process would be redirected to a named pipe, which would be read by the Java worker monitoring threads.

### Implementation Notes

Previous Code Structure

* Every worker we spawn by sending front end scale requests has a corresponding **WorkerThread** & **WorkerLifeCycle** java object which create a python process and monitors it.

```curl -X PUT "http://localhost:8081/models/mnist_v2?min_worker=1&synchronous=true"```

* **WorkerLifeCycle** is an interface to the python worker process. It creates the process, Monitors for process coming alive, Manages state based on stdout / stderr 
* **WorkerThread** is responsible for sending messages - LOAD & INFERENCE & interfacing with other HTTP / Batch Aggregation / Netty related front end code
* **ts/model_service_worker.py** is the worker class responsible for listening for messages in a specific unix / tcp socket. It can receive messages for LOAD / INFERENCE and perform appropriate action.

Current Code Structure

* Every model we register has a **WorkerManagerThread** & **WorkerManagerLifeCycle** which is responsible for sending messages for LOAD / SCALEUP-DOWN to a python process ts/model_service_worker_manager.py
* **ts/model_service_worker_manager.py** - Is the python process responsible for listening to LOAD / SCALEUP requests from frontend and it spawns / kills an instance of ts/model_service_worker.py 
* As before WorkerThread & WorkerLifeCycle are responsible for creating & interacting the python worker process. However this time 
* **WorkerLifeCycle** does not create / destroy the python process directly. That's handled by WorkerManagerThread by sending a SCALE-UP.
* **WorkerThread** is only responsible for sending / receiving only INFERENCE messages, as LOAD / is handled by WorkerManagerThread
* ts/model_service_worker.py is the worker class responsible for handling INFERENCE. LOAD is handled ts/model_service_worker_manager.py as mentioned before.

### CODE REVIEW

File: frontend/build.gradle [hunk 1]
```diff
@@ -36,9 +36,6 @@ configure(javaProjects()) {
 
     defaultTasks 'jar'
```
dhanainme, line 38: Disabling checkstyle temporarily

bravelywangexocr:
Hi dhanainme,

Can you help share the due date of this optimization? 

I just installed this branch from source and tested loading multiple workers for a specific model on a same GPU, it appears that the child worker processes do not share the shared model reference of parent process model_service_worker_manager.py. 

So I just wonder if it's my setting's problem or the feature is still under-development. 

Thanks a lot!

dhanainme:
> Hi dhanainme,
> 
> Can you help share the due date of this optimization?

The core feature works seems to be working (We can now load 23 models vs 9 models on on p2.8). Still working on some tests. This should be a part of the next release 0.4.0

> I just installed this branch from source and tested loading multiple workers for a specific model on a same GPU, it appears that the child worker processes do not share the shared model reference of parent process model_service_worker_manager.py.
> 
> So I just wonder if it's my setting's problem or the feature is still under-development.

I just verified this again on a p2.8xl and could confirm the model is shared for BERT Eager Model in the Model Zoo. The model sharing works only for Eager models & not scripted models. 

Internally we use torch.multiprocessing for sharing the models. You may want to confirm that your model works for a very simple case like  - https://gist.github.com/dhanainme/5344dc83b1745b8fa3cda402bfc98981

Please try it out and share more info (model, gpu env, testing info) if you still dont find it working.

bravelywangexocr:
> > Hi dhanainme,
> > Can you help share the due date of this optimization?
> 
> The core feature works seems to be working (We can now load 23 models vs 9 models on on p2.8). Still working on some tests. This should be a part of the next release 0.4.0
> 
> > I just installed this branch from source and tested loading multiple workers for a specific model on a same GPU, it appears that the child worker processes do not share the shared model reference of parent process model_service_worker_manager.py.
> > So I just wonder if it's my setting's problem or the feature is still under-development.
> 
> I just verified this again on a p2.8xl and could confirm the model is shared for BERT Eager Model in the Model Zoo. The model sharing works only for Eager models & not scripted models.
> 
> Internally we use torch.multiprocessing for sharing the models. You may want to confirm that your model works for a very simple case like - https://gist.github.com/dhanainme/5344dc83b1745b8fa3cda402bfc98981
> 
> Please try it out and share more info (model, gpu env, testing info) if you still dont find it working.
> 
> Also, I ll upload specific information / logs to confirm how this was tested shortly.

Thanks for your reply!

Probably that's because I used scripted model. I'll give it a try with eager one. 
BTW, what's the cause that scripted model can not benefit from this optimization? 

Thanks!

dhanainme:
Windows build failing with insufficient memory : 

# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (malloc) failed to allocate 32784 bytes for Chunk::new

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkLoadManager.java
```diff
@@ -92,9 +94,67 @@ public int getNumRunningWorkers(ModelVersionName modelVersionName) {
             CompletableFuture<Integer> future = new CompletableFuture<>();
             int minWorker = model.getMinWorkers();
             int maxWorker = model.getMaxWorkers();
+            WorkerManagerThread workerManagerThread;
             List<WorkerThread> threads;
             if (minWorker == 0) {
+
+                logger.info("Min Worker is 0 - DHANAK");
```
lxning, line 101: "DHANAK" log clean?

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkLoadManager.java
```diff
@@ -92,9 +94,67 @@ public int getNumRunningWorkers(ModelVersionName modelVersionName) {
...
+                workerManagerThread = workerManagers.remove(model.getModelVersionName());
                 threads = workers.remove(model.getModelVersionName());
+
+                if (workerManagerThread != null) {
+
+                    logger.info("Min Worker is not null - DHANAK");
+
+                    for (WorkerThread thread : threads) {
+                        workerManagerThread.scaleDown(thread.getLifeCycle().getPort());
```
lxning, line 110: To my understanding, workerManagerThread.scaleDown is going to shut down the backend worker. It seems that the following code "workerkillprocess" does same thing.

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkLoadManager.java
```diff
@@ -92,9 +94,67 @@ public int getNumRunningWorkers(ModelVersionName modelVersionName) {
...
+                            workerkillprocess.waitFor(
+                                    configManager.getUnregisterModelTimeout(), TimeUnit.SECONDS);
+                            logger.info("Worker Terminated - DHANAK " + cmd);
+                        } catch (InterruptedException | IOException e) {
+                            logger.warn(
+                                    "WorkerManagerThread interrupted during waitFor, possible async resource cleanup.");
+                            future.complete(HttpURLConnection.HTTP_INTERNAL_ERROR);
+                            return future;
+                        }
+                    }
```
lxning, line 125: The above steps are 
1. kill backend worker process.
2. kill frontend worker thread.
what if the frontend worker thread still sends msg to the backend died worker?

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkLoadManager.java
dhanainme: Sorry may have missed it. Will fix it

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerManagerThread.java
```diff
@@ -0,0 +1,474 @@
...
+                if (reply != null) {
+                    aggregator.sendResponse(reply);
+                } else {
+                    int val = model.incrFailedInfReqs();
+                    logger.error("Number or consecutive unsuccessful inference {}", val);
+                    throw new WorkerInitializationException(
+                            "Backend worker did not respond in given time");
+                }
+                switch (req.getCommand()) {
+                    case LOAD:
```
lxning, line 157: What is WorkerManagerThread or WorkerManagerLifeCycle's action if load fails? At least, scaleUp should not be proceeded.

frontend/server/src/main/java/org/pytorch/serve/wlm/BatchAggregator.java
```diff
@@ -31,7 +34,33 @@ public BaseModelRequest getRequest(String threadName, WorkerState state)
...
+            throws InterruptedException {
+        jobs.clear();
+
+        BaseModelRequest req = null;
+
+        model.pollBatch(
+                threadName,
+                (state == WorkerManagerState.WORKER_MODEL_LOADED) ? 0 : Long.MAX_VALUE,
+                jobs,
+                true);
```
dhanainme, line 63: Explain why we need this change

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
```diff
@@ -60,6 +60,11 @@ public Model(ModelArchive modelArchive, int queueSize) {
                         this.modelArchive.getModelName(), this.modelArchive.getModelVersion());
     }
 
+    public Map<String, LinkedBlockingDeque<Job>> getJobsDB() {
+        // TODO : Do you need this ?
```
dhanainme, line 64: TODO

ts/model_server.py
```diff
@@ -56,7 +57,7 @@ def start():
         java = "java" if not java_home else "{}/bin/java".format(java_home)
 
         ts_home = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
-        cmd = [java, "-Dmodel_server_home={}".format(ts_home)]
+        cmd = [java, "-Dmodel_server_home={}".format(ts_home), "-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005"]
```
dhanainme, line 60: Remove debugging

frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java
```diff
@@ -1103,8 +1103,8 @@ public void testErrorBatch() throws InterruptedException {
 
         TestUtils.getLatch().await();
 
-        Assert.assertEquals(TestUtils.getHttpStatus(), HttpResponseStatus.INSUFFICIENT_STORAGE);
-        Assert.assertEquals(TestUtils.getResult(), "Invalid response");
+        // Assert.assertEquals(TestUtils.getHttpStatus(), HttpResponseStatus.INSUFFICIENT_STORAGE);
+        // Assert.assertEquals(TestUtils.getResult(), "Invalid response");
```
dhanainme, line 1107: Remove asserts

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
dhanainme: Fixed

ts/model_server.py
dhanainme: Fixed

frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java
dhanainme: Fixed

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkLoadManager.java
dhanainme: The moment the backend worker is killed, The frontend thread would not be able to establish a socket connection to the backend.

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerManagerThread.java
dhanainme: This may be an existing bug. It looks like we dont respond to ERROR state of the worker in the right manner. Have fixed this with an explicit condition in in the control loop.

ts/tests/unit_tests/test_model_service_worker.py
```diff
@@ -35,7 +35,7 @@ def socket_patches(mocker):
 @pytest.fixture()
 def model_service_worker(socket_patches):
     if not sys.platform.startswith("win"):
-        model_service_worker = TorchModelServiceWorker('unix', 'my-socket', None, None)
+        model_service_worker = TorchModelServiceWorker('unix', 'my-socket', None, None, fifo_path="fifo")
     else:
         model_service_worker = TorchModelServiceWorker('tcp', 'my-socket', None, port_num=9999)
```
lxning, line 40: Window build failed due to fifo_path missing at here

chauhang:
Closing this as the solution did not solve as expected

Issue #782, opened 2020-11-13
Default Netty thread in configuration.md is different from ConfigManager.java


rishabh1212:
## 📚 Documentation

I want to make a pull request for this, but I am not able to push to `pytorch/serve`
* `number_of_netty_threads`: Number frontend netty thread. This specifies the numer of threads in the child [EventLoopGroup](https://livebook.manning.com/book/netty-in-action/chapter-8) of the frontend netty server. This group provides EventLoops for processing Netty Channel events (namely inference and management requests) from accepted connections. Default: ~number of logical processors available to the JVM~.
Default is 0.

dhaniram-kshirsagar:
@rishabh1212 Commented on your PR #873

rishabh1212:
@dhaniram-kshirsagar I update PR, please review

PR #873, opened 2020-12-11
fix default netty thread in docs as 0 and minor markdown syntax issue

rishabh1212:
## Description
- In the docs [configuration.md](https://github.com/pytorch/serve/blob/master/docs/configuration.md), it is mentioned that default netty thread number is the number of available logical processors
- According to [ConfigManager.java](https://github.com/pytorch/serve/blob/bacdf0a626fc24d89f332c413a2039f7c0cf9953/frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java#L316), the Netty threads set are 0, and nowhere else they are set as ```Runtime.getRuntime().availableProcessors();```.
- Only the default number of workers are set as the number of logical processors.

Fixes https://github.com/pytorch/serve/issues/782

## Type of change

- 

## Checklist:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -198,16 +198,15 @@ By default, TorchServe uses all available GPUs for inference. Use `number_of_gpu
 
 ### Enable metrics api
 * `enable_metrics_api` : Enable or disable metric apis i.e. it can be either `true` or `false`. Default: true (Enabled)
-* `metrics_format` : Use this to specify metric report format . At present, the only supported and default value for this is `prometheus'
-		     This is used in conjunction with `enable_meterics_api` option above.
+* `metrics_format` : Use this to specify metric report format. At present, the only supported and default value for this is `prometheus`. This is used in conjunction with `enable_meterics_api` option above.
 
 ### Other properties
 
 Most of the following properties are designed for performance tuning. Adjusting these numbers will impact scalability and throughput.
 
 * `enable_envvars_config`: Enable configuring TorchServe through environment variables. When this option is set to "true", all the static configurations of TorchServe can come through environment variables as well. Default: false
-* `number_of_netty_threads`: Number frontend netty thread. This specifies the numer of threads in the child [EventLoopGroup](https://livebook.manning.com/book/netty-in-action/chapter-8) of the frontend netty server. This group provides EventLoops for processing Netty Channel events (namely inference and management requests) from accepted connections. Default: number of logical processors available to the JVM.
-* `netty_client_threads`: Number of backend netty thread. This specifies the number of threads in the WorkerThread [EventLoopGroup](https://livebook.manning.com/book/netty-in-action/chapter-8) which writes inference responses to the frontend. Default: number of logical processors available to the JVM.
+* `number_of_netty_threads`: Number of frontend netty thread. This specifies the numer of threads in the child [EventLoopGroup](https://livebook.manning.com/book/netty-in-action/chapter-8) of the frontend netty server. This group provides EventLoops for processing Netty Channel events (namely inference and management requests) from accepted connections. Default: 0.
+* `netty_client_threads`: Number of backend netty thread. This specifies the number of threads in the WorkerThread [EventLoopGroup](https://livebook.manning.com/book/netty-in-action/chapter-8) which writes inference responses to the frontend. Default: 0.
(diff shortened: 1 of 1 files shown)


rishabh1212:
@harshbafna @dhaniram-kshirsagar I made a change to documentation. Can you please review?

rishabh1212:
@harshbafna @dhaniram-kshirsagar @maaquib All checks are passed. Is there anything I should do next? I am sorry this is my first time making a PR to open source!

rishabh1212:
@harshbafna @dhaniram-kshirsagar Can you please clarify on whether https://github.com/pytorch/serve/issues/782 is actually an issue or just my misunderstanding about Netty threads?

dhaniram-kshirsagar:
> @harshbafna @dhaniram-kshirsagar Can you please clarify on whether #782 is actually an issue or just my misunderstanding about Netty threads?

@rishabh1212 The modification you have made are incorrect. However, it seems like our doc is indeed incorrect a bit. Here is netty code snippet which does trick of using available number of processor to jvm when we specify number of threads as 0 from torchserve configmanager. 
```java
 DEFAULT_EVENT_LOOP_THREADS = Math.max(1, SystemPropertyUtil.getInt(
                "io.netty.eventLoopThreads", NettyRuntime.availableProcessors() * 2));

Please correct those doc lines according to above snippet.

**Comment by rishabh1212** [comment]
> @rishabh1212 The modification you have made are incorrect. However, it seems like our doc is indeed incorrect a bit. Here is netty code snippet which does trick of using available number of processor to jvm when we specify number of threads as 0 from torchserve configmanager.
> 
> ```java
>  DEFAULT_EVENT_LOOP_THREADS = Math.max(1, SystemPropertyUtil.getInt(
>                 "io.netty.eventLoopThreads", NettyRuntime.availableProcessors() * 2));
> 
> Please correct those doc lines according to above snippet.
> ```

@dhaniram-kshirsagar thanks for your comment. I fixed the docs in PR and also updated description of PR. Please review

**Review by harshbafna — CHANGES_REQUESTED** [review]
@rishabh1212: This looks good. Could you also update the default values in the `dumpConfiguration` method of ConfigManager? As it prints the default value as 0 at TorchServe startup and might create confusion.

**Comment by rishabh1212** [comment]
> @rishabh1212: This looks good. Could you also update the default values in the `dumpConfiguration` method of ConfigManager? As it prints the default value as 0 at TorchServe startup and might create confusion.

@harshbafna I made the changes, please review

File: frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java [hunk 1]
```diff
@@ -572,9 +578,9 @@ public String dumpConfigurations() {
                 + "\nMetrics dir: "
                 + getCanonicalPath(System.getProperty("METRICS_LOCATION"))
                 + "\nNetty threads: "
-                + getNettyThreads()
+                + (getNettyThreads() == 0 ? DEFAULT_EVENT_LOOP_THREADS : getNettyThreads())
```
harshbafna, line 581: Can't we simply return the default value set in Netty's MultithreadEventLoopGroup class?

harshbafna (changes requested):

frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
rishabh1212: @harshbafna I did not want to change the way Netty calculate the value. I could have returned `2*runtime.availableProcessors()`.  For example, in future if they change the value returned the function `NettyRuntime.availableProcessors`, then the value printed by us will not be consistent with they function they use to calculate default.

harshbafna:
@rishabh1212: The build is failing because of a formatting issue. You can use the following commands:

```
cd frontend
./gradlew fJ
```

rishabh1212:
> @rishabh1212: The build is failing because of a formatting issue. You can use the following commands:
> 
> ```
> cd frontend
> ./gradlew fJ
> ```

Thanks, I was trying to fix it.

rishabh1212:
@harshbafna, thanks! I was able to fix it. Please check https://github.com/pytorch/serve/pull/873#discussion_r544047443 and review

rishabh1212:
@msaroufim I have made requested fixes with some [explanation](https://github.com/pytorch/serve/pull/873#issuecomment-746422289). I think its ready for review

Issue #898, opened 2020-12-14
BERT model fails to run on k8s with docker image built from source


jeremiahschung:
## Context

* torchserve version: source
* torch-model-archiver version: source
* torch version: 1.7.1 
* torchvision version [if any]: 0.8.2
* torchtext version [if any]:0.8.1
* torchaudio version [if any]:
* java version:openjdk11 
* Operating System and version: ubuntu18.04

## Your Environment

* Installed using source? [yes/no]:
* Are you planning to deploy it using docker container? [yes/no]:yes
* Is it a CPU or GPU environment?: GPU
* Using a default/custom handler? [If possible upload/share custom handler/model]: Transformer_handler_generalized.py
* What kind of model is it e.g. vision, text, audio?: BERT Seq classification 

## Expected Behavior

Prediction succeeds.

## Current Behavior

Prediction fails with error "ModuleNotFoundError: No module named 'captum'"

## Possible Solution

In /serve/docker/Dockerfile, add "pip install captum"

## Steps to Reproduce

1. Build the a torchscripted default BERT model .mar file using the guide [here](https://github.com/pytorch/serve/tree/master/examples/Huggingface_Transformers)
2. Build a custom docker image from source with the guide [here](https://github.com/pytorch/serve/tree/master/docker), push it into the docker hub for remote access.
3. Create an EKS cluster per instructions [here](https://github.com/pytorch/serve/tree/master/kubernetes/EKS)
4. Before running `helm install ts .`, modify the config.properties file to use your torchserve docker image built in step #2, change the model setting to include the BERT model definition & file and copy the config file and the .mar file into the persistent volume k8s pod. 
5. Install torchserve with `helm install ts .` in /serve/kubernetes/Helm and make an inference to the BERT model
6. Run `kubectl exec <torchserve_pod_name> -- cat logs/ts_log.log and notice the error "ModuleNotFoundError: No module named 'captum'"

## Failure Logs [if any]

[1214_master_bert_k8s_failed.txt](https://github.com/pytorch/serve/files/5687494/1214_master_bert_k8s_failed.txt)

This was fixed after I rebuilt the Torchserve docker image with "pip install captum" in the Dockerfile and used that for the Helm chart install.

jeremiahschung:
Initially reported the issue with Kubernetes, but this is due to the Transformer_handler_generalized.py being updated to have captum as a dependency. Newly archived BERT models with the updated handler will face the same issue. 

[...]

PR #901, opened 2020-12-14
Issue 898

jeremiahschung:
## Description

Fixes the torchserve compiled package missing the python dependencies (captum and transformers) which are required for the custom handler for the HuggingFace models. 

Also moves the transformers from /serve/requirements/developer.txt to /serve/requirements/common.txt as it's required to run BERT models. Without it, users would need to build their own TorchServe docker image in order to run the BERT models as opposed to downloading it from our official images or pip.

Fixes https://github.com/pytorch/serve/issues/898

## Type of change

- [x] Fix production Dockerfile and use it for the TorchServe deployment to make inference to a BERT Seq classification model.

- UT/IT execution results
[1214_fix_bert_k8s_success.txt](https://github.com/pytorch/serve/files/5691992/1214_fix_bert_k8s_success.txt)

- Logs

## Checklist:

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -64,7 +64,6 @@ RUN TORCH_VER=$(curl --silent --location https://pypi.org/pypi/torch/json | pyth
         pip install --no-cache-dir torch==$TORCH_VER+cpu torchvision==$TORCH_VISION_VER+cpu -f https://download.pytorch.org/whl/torch_stable.html; \
     fi
 RUN pip install --no-cache-dir torchtext torchserve torch-model-archiver
-RUN pip install --no-cache-dir transformers captum
 
 # Final image for production
 FROM ${BASE_IMAGE} AS runtime-image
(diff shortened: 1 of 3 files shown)


requirements/common.txt
```diff
@@ -6,3 +6,4 @@ sentencepiece
 pillow
 captum
 packaging
+transformers==2.10.0
```
maaquib, line 9: We should keep this in `developers.txt`. `transformers` is not a runtime dependency for TS. It is only required as a test dependency.

docker/Dockerfile
```diff
@@ -64,6 +64,7 @@ RUN TORCH_VER=$(curl --silent --location https://pypi.org/pypi/torch/json | pyth
         pip install --no-cache-dir torch==$TORCH_VER+cpu torchvision==$TORCH_VISION_VER+cpu -f https://download.pytorch.org/whl/torch_stable.html; \
     fi
 RUN pip install --no-cache-dir torchtext torchserve torch-model-archiver
+RUN pip install --no-cache-dir transformers captum
```
maaquib, line 67: This should only be added to the Dev dockerfile

maaquib (changes requested):

maaquib:
@jeremiahschung I am going to close this pr since the changes made here are for model specific dependencies which should not be bundled with the runtime ones

PR #978, opened 2021-02-23
update doc for scale workers 

lxning:
## Description

Currently number_gpu is only applied in GRPC implementation. In fact, gpu assignment is based on "number_of_gpu" defined in config.properties. And also "If number_gpu exceeds the number of available GPUs, the rest of workers will run on CPU." is not true. So remove this statement.

Fixes #(issue)
950

## Type of change
document update

- [ ] Test A
- [ ] Test B

- UT/IT execution results

- Logs

## Checklist:


docs/management_api.md
```diff
@@ -83,7 +83,6 @@ This API follows the [ManagementAPIsService.ScaleWorker](../frontend/server/src/
 
 * `min_worker` - (optional) the minimum number of worker processes. TorchServe will try to maintain this minimum for specified model. The default value is `1`.
 * `max_worker` - (optional) the maximum number of worker processes. TorchServe will make no more that this number of workers for the specified model. The default is the same as the setting for `min_worker`.
-* `number_gpu` - (optional) the number of GPU worker processes to create. The default value is `0`. If number_gpu exceeds the number of available GPUs, the rest of workers will run on CPU.
```
maaquib, line 86: I see this being referred in the Open API [desc](https://github.com/pytorch/serve/search?q=%22number_gpu%22) as well. Given that this parameter will still be available for gRPC, we should change the documentation instead of removing it completely

maaquib (changes requested):

docs/management_api.md
lxning: "number_gpu" is defined as a parameter in management API. However, it is never used in the [GPU assignment code](https://github.com/pytorch/serve/blob/master/frontend/server/src/main/java/org/pytorch/serve/wlm/WorkLoadManager.java#L163). That's why I remove this parameter in this API page to avoid confusion.

maaquib (approved):

dhanainme (approved):

PR #944, opened 2021-01-23
Add missing repo

vivanov879:
Add the missing repo name in docker command

## Description

Add the missing repo name in docker command

## Type of change
Docs update

PR #1029, opened 2021-04-08
Update docker README.md

msaroufim:
* In step 1, user sets up a name for the container so just use it in step 2
* Step 3-4 aren't super clear that these commands need to be run locally. Can't run wget or curl on the docker instance without root access or changing the dockerfile
* pytorch/ needs to be added to container name otherwise it won't be found

This is purely a documentation change so below is N/A

## Description

Fixes #(issue)

## Type of change

- [ x] This change requires a documentation update


msaroufim:
@dhaniram-kshirsagar  can you please check this out - this PR is a superset of this one https://github.com/pytorch/serve/pull/944/files so at least one other person was confused by Docker instructions

Issue #734, opened 2020-10-17
Java index out of bounds exception when running many requests through server


adumit:
## Context
Trying to loadtest a torch serve model to gauge performance on a custom handler. 

* torchserve version: 0.2.0
* torch version: 1.6.0
* java version: openjdk 11.0.8
* Operating System and version: Debian via the python 3.7-buster image.

## Your Environment

* Are you planning to deploy it using docker container? [yes/no]: yes
* Is it a CPU or GPU environment?: CPU
* Using a default/custom handler? custom
* What kind of model is it e.g. vision, text, audio?: feed forward for custom input.
* Are you planning to use local models from model-store or public url being used e.g. from S3 bucket etc.? from model store
* Provide config.properties, logs [ts.log] and parameters used for model registration/update APIs: number of netty threads=32

## Expected Behavior
Expected torch serve not to throw this error or understand what properties of the environment I could change to address it. It only seems to happen on medium load.

## Current Behavior
With a load of ~5rps and varying batch size and CPU memory and count allocations the server will throw an errors in ~4%+ of requests.

## Failure Logs [if any]
`2020-10-17 00:16:41,887 [INFO ] epollEventLoopGroup-5-3 org.pytorch.serve.wlm.WorkerThread - 9002 Worker disconnected. WORKER_MODEL_LOADED
2020-10-17 00:16:41,887 [ERROR] epollEventLoopGroup-5-3 org.pytorch.serve.wlm.WorkerThread - Unknown exception
io.netty.handler.codec.DecoderException: java.lang.IndexOutOfBoundsException: readerIndex(1021) + length(4) exceeds writerIndex(1024): PooledUnsafeDirectByteBuf(ridx: 1021, widx: 1024, cap: 1024)
        at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:471)
        at io.netty.handler.codec.ByteToMessageDecoder.channelInputClosed(ByteToMessageDecoder.java:404)
        at io.netty.handler.codec.ByteToMessageDecoder.channelInputClosed(ByteToMessageDecoder.java:371)
        at io.netty.handler.codec.ByteToMessageDecoder.channelInactive(ByteToMessageDecoder.java:354)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:262)
[...]

PR #1035, opened 2021-04-15
check number of available bytes before read

fefe982:
## Description

The frontend communicates with the backend using unix socket or tcp. During the transmission of TCP, the data might be split at any byte in between, and must be reconstructed at the receiver end to form a full bulk of data. And int of 4 bytes with the value of 0x12345678 sent by the sender might be split into two packets, with one ending with 0x5678, the next starting with 0x1234 (depending on the byte order used to transmit the int). So there must be a check for completeness of the data, before interpreting it as an int. 

This check exists for every read in `ModelResponseDecoder.decode`, except for the `httpStausCode`. This is possible cause for #734 . This fix incorprate the fix mentioned by @punshriv at in [this comment](https://github.com/pytorch/serve/issues/734#issuecomment-745279709)

## Type of change

- [ ] Test A
- [ ] Test B

- UT/IT execution results

- Logs

## Checklist:


msaroufim:
Hi @fefe982 interesting catch. Can you confirm that issue #734 indeed goes away once this is done. Also looks like the CI is failing.

Thank you for the investigative work!

fefe982:
#734 is kind of random and kind of highly rely on network stack of the OS. It is hard to reproduce so hard to confirm that it is "gone".

I cannot personally confirm the effect of the fix on #734. One of my colleagues, who has already left our company, met this problem. This problem can be confirmed with several of my other colleagues.  He left this fixed code, which he claims that it fixed the problem. As a matter of fact, with this fix, our service online is not suffering from this problem anymore.

This should fix #734 theoretically.

For the failing CI, I would need some help. I don't know what happened based on the log.

fefe982:
I'm have the problem described in #734 in one of my projects now. And I can confirm that this fixes the problem for me.

punshriv:
@fefe982 i am one of the original reporters of #734 and i can confirm that this issue gets resolved by the suggested fix so it will be great if we have this merged.

lxning (approved):

maaquib (approved):

Issue #1037, opened 2021-04-21
Duplicated process getting launched  on second gpu in multi-gpu setting during serving CV models 


HamidShojanazeri:
Please have a look at [FAQ's](../../docs/FAQs.md) and [Troubleshooting guide](../../docs/Troubleshooting.md), your query may be already addressed.

Your issue may already be reported!
Please search on the [issue tracker](https://github.com/pytorch/serve/issues) before creating one.

## Context
During serving of CV examples (e.g image classification/image segementation) on a multi-gpu machine and monitoring the GPUs, for each worker two processes get launched on two different gpus with same PID. The expected behavior is having one process per worker. ---update this happens in all examples where handler is inheriting/ using initialize method of the base_handler, as there is redundant load of model to device in this method. 

* torchserve version: 0.3.0 
* torch-model-archiver version:0.3.0
* torch version:1.7 & 1.8
* torchvision version [if any]:0.8.0 and 0.9.0
* Operating System and version: ubuntu 18.04

## Your Environment

* Installed using source? [yes/no]:yes
* Are you planning to deploy it using docker container? [yes/no]:no
* Is it a CPU or GPU environment?:GPU
* Using a default/custom handler? [If possible upload/share custom handler/model]:default vision handler
* What kind of model is it e.g. vision, text, audio?: CV
* Provide config.properties, logs [ts.log] and parameters used for model registration/update APIs:No

## Expected Behavior
The expected behavior is having one process per worker. 

## Steps to Reproduce
serving each of CV examples in the repo ( image_classification resnet 18/resnet152_batch) with one worker on a multi-gpu machine and monitor the GPUs using Nvidia-smi.

## Failure Logs [if any]

Here is screenshot of serving image segmentation example with one worker. 

<img width="1792" alt="Screen Shot 2021-04-21 at 1 20 40 PM" src="https://user-images.githubusercontent.com/9162336/115615938-694e2f80-a2a4-11eb-9d37-d2977fe12534.png">

PR #1038, opened 2021-04-22
Issue 1037

HamidShojanazeri:
## Description

Fixes the issue #1037 

Fixes #(issue)

## Type of change

- [ ] Test A
- [ ] Test B

- UT/IT execution results

- Logs
Sanity test logs are added.
Torchscripted model traced on Cuda:3 and ran on Cuda:0,1
Screen Shot 2021-04-27 at 6 34 42 PM
Eager model
Screen Shot 2021-04-27 at 6 32 55 PM

[sanity.txt](https://github.com/pytorch/serve/files/6355540/sanity.txt)

## Checklist:

### DIFF
--- a/ts/torch_handler/base_handler.py
+++ b/ts/torch_handler/base_handler.py
@@ -91,7 +91,7 @@ class BaseHandler(abc.ABC):
         Returns:
             (NN Model Object) : Loads the model object.
         """
-        return torch.jit.load(model_pt_path, map_location=self.map_location)
+        return torch.jit.load(model_pt_path, map_location=self.device)
 
     def _load_pickled_model(self, model_dir, model_file, model_pt_path):
         """
@@ -126,7 +126,7 @@ class BaseHandler(abc.ABC):
         model_class = model_class_definitions[0]
         model = model_class()
         if model_pt_path:
-            state_dict = torch.load(model_pt_path, map_location=self.map_location)
+            state_dict = torch.load(model_pt_path, map_location=self.device)
             model.load_state_dict(state_dict)
         return model
 
--- a/ts/torch_handler/base_handler.py
+++ b/ts/torch_handler/base_handler.py
@@ -44,12 +44,11 @@ class BaseHandler(abc.ABC):
 
         """
         properties = context.system_properties
-        self.map_location = "cuda" if torch.cuda.is_available() else "cpu"
-        self.device = torch.device(
-            self.map_location + ":" + str(properties.get("gpu_id"))
-            if torch.cuda.is_available()
-            else self.map_location
-        )
+        self.device = "cuda" if torch.cuda.is_available() else "cpu"
+        if torch.cuda.is_available():
+            torch.cuda.set_device(properties.get("gpu_id"))
+
+
         self.manifest = context.manifest
         model_dir = properties.get("model_dir")
         model_pt_path = None
@@ -125,7 +124,7 @@ class BaseHandler(abc.ABC):
         model_class = model_class_definitions[0]
         model = model_class()
         if model_pt_path:
-            state_dict = torch.load(model_pt_path)
+            state_dict = torch.load(model_pt_path, map_location=self.device)
             model.load_state_dict(state_dict)
         return model
 
--- a/ts/torch_handler/base_handler.py
+++ b/ts/torch_handler/base_handler.py
@@ -58,12 +58,10 @@ class BaseHandler(abc.ABC):
 
         # model def file
         model_file = self.manifest["model"].get("modelFile", "")
-
         if model_file:
             logger.debug("Loading eager model")
             self.model = self._load_pickled_model(model_dir, model_file, model_pt_path)
             self.model.to(self.device)
-            
         else:
             logger.debug("Loading torchscript model")
             if not os.path.isfile(model_pt_path):
--- a/ts/torch_handler/base_handler.py
+++ b/ts/torch_handler/base_handler.py
@@ -63,6 +63,8 @@ class BaseHandler(abc.ABC):
         if model_file:
             logger.debug("Loading eager model")
             self.model = self._load_pickled_model(model_dir, model_file, model_pt_path)
+            self.model.to(self.device)
+            
         else:
             logger.debug("Loading torchscript model")
             if not os.path.isfile(model_pt_path):
@@ -70,7 +72,6 @@ class BaseHandler(abc.ABC):
 
             self.model = self._load_torchscript_model(model_pt_path)
 
-        self.model.to(self.device)
         self.model.eval()
         logger.debug('Model file %s loaded successfully', model_pt_path)
 
@@ -89,7 +90,7 @@ class BaseHandler(abc.ABC):
         Returns:
             (NN Model Object) : Loads the model object.
         """
-        return torch.jit.load(model_pt_path)
+        return torch.jit.load(model_pt_path, map_location=self.device)
 
     def _load_pickled_model(self, model_dir, model_file, model_pt_path):
         """
--- a/ts/torch_handler/base_handler.py
+++ b/ts/torch_handler/base_handler.py
@@ -51,7 +51,6 @@ class BaseHandler(abc.ABC):
             else self.map_location
         )
         self.manifest = context.manifest
-    
         model_dir = properties.get("model_dir")
         model_pt_path = None
         if "serializedFile" in self.manifest["model"]:
--- a/ts/torch_handler/base_handler.py
+++ b/ts/torch_handler/base_handler.py
@@ -51,7 +51,7 @@ class BaseHandler(abc.ABC):
             else self.map_location
         )
         self.manifest = context.manifest
-
+    
         model_dir = properties.get("model_dir")
         model_pt_path = None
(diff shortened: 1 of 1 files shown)


ts/torch_handler/base_handler.py
```diff
@@ -91,7 +89,7 @@ def _load_torchscript_model(self, model_pt_path):
         Returns:
             (NN Model Object) : Loads the model object.
         """
-        return torch.jit.load(model_pt_path, map_location=self.map_location)
```
chauhang, line 94: @HamidShojanazeri Are things working fine for CPU/GPU cases and multi-GPU scenarios after thus fix?

chauhang (changes requested):
Please see comments inline

ts/torch_handler/base_handler.py
```diff
@@ -64,16 +62,16 @@ def initialize(self, context):
...
+            self.model.to(self.device)
+            
         else:
             logger.debug("Loading torchscript model")
             if not os.path.isfile(model_pt_path):
                 raise RuntimeError("Missing the model.pt file")
 
             self.model = self._load_torchscript_model(model_pt_path)
 
-        self.model.to(self.device)
```
vdantu, line 74: why are we removing this?

nskool:
Could we fix the failing PR builds?

ts/torch_handler/base_handler.py
```diff
@@ -44,14 +44,12 @@ def initialize(self, context):
...
         """
         properties = context.system_properties
-        self.map_location = "cuda" if torch.cuda.is_available() else "cpu"
-        self.device = torch.device(
-            self.map_location + ":" + str(properties.get("gpu_id"))
-            if torch.cuda.is_available()
-            else self.map_location
-        )
-        self.manifest = context.manifest
+        self.device = "cuda" if torch.cuda.is_available() else "cpu"
```
ngimel, line 47: setting self.device to just "cuda" without id is suspect

ts/torch_handler/base_handler.py
```diff
@@ -64,16 +62,16 @@ def initialize(self, context):
         if model_file:
             logger.debug("Loading eager model")
             self.model = self._load_pickled_model(model_dir, model_file, model_pt_path)
+            self.model.to(self.device)
```
ngimel, line 65: map_location (if done correctly) should have taken care of sending model to device

ts/torch_handler/base_handler.py
```diff
@@ -91,7 +89,7 @@ def _load_torchscript_model(self, model_pt_path):
         Returns:
             (NN Model Object) : Loads the model object.
         """
-        return torch.jit.load(model_pt_path, map_location=self.map_location)
+        return torch.jit.load(model_pt_path, map_location=self.device)
```
ngimel, line 92: here you should have device with device_id, instead of just "cuda"

ts/torch_handler/base_handler.py
ngimel: Since this map_location was just "cuda", it would initialize context on gpu 0, this was an error

ts/torch_handler/base_handler.py
HamidShojanazeri: @vdantu we are moving it to the IF block for the eager mode in line#65 as it is not required for torchscript, in fact having the leaving it here will make the duplicated process for torchscript mode.

ts/torch_handler/base_handler.py
HamidShojanazeri: @ngimel Thanks for your comments, in fact the device was cuda+id,  
```
self.device = torch.device(
            self.map_location + ":" + str(properties.get("gpu_id"))
            if torch.cuda.is_available()
            else self.map_location
        )
```
however, this way, setting the torch.jit.load(model_pt_path, map_location=self.device) was failing. In current fix, the device id is getting set using [torch.cuda.set_device(properties.get("gpu_id"))](https://github.com/pytorch/serve/blob/issue_1037/ts/torch_handler/base_handler.py#L49).

ngimel: In what way is it failing?

ngimel: Looks like `torch.load` ignores map location, but `torch.jit.load` correctly loads the model to the specified location:
```
In [1]: import torch

In [2]: m=torch.jit.load("alexnet.pt", map_location="cuda:1")

In [3]: p=m.parameters().__next__()

In [4]: p.device
Out[4]: device(type='cuda', index=1)

In [5]: torch.cuda.memory.list_gpu_processes(0)
Out[5]: 'GPU:0\nno processes are running'

In [6]: torch.cuda.memory.list_gpu_processes(1)
Out[6]: 'GPU:1\nprocess    2011967 uses     1113.000 MB GPU memory'

In [7]: m=torch.load("alexnet.pt", map_location="cuda:1")
/data/users/ngimel/pytorch/torch/serialization.py:601: UserWarning: 'torch.load' received a zip file that looks like a TorchScript archive dispatching to 'torch.jit.load' (call 'torch.jit.load' directly to silence this warning)
  warnings.warn("'torch.load' received a zip file that looks like a TorchScript archive"

In [8]: p=m.parameters().__next__()

In [9]: p.device
Out[9]: device(type='cpu')

```
`torch.load` actually warns and suggests using `torch.jit.load`

HamidShojanazeri: @ngimel, exactly, torch.load ignores the map_location, but jit.load works fine. The current workaround, using torch.cuda.set_device(gpu_id) seems working. So this might be a torchscripted checkpoint?

If we load the pre-trained model save the checkpoints then map_location works fine. 

```
from torchvision
import models
import torch
import time
from torchvision.models.alexnet import AlexNet
class ImageClassifier(AlexNet):
    def __init__(self):
        super(ImageClassifier, self).__init__()
              
device1='cuda:0'
device2='cuda:2'
model = models.alexnet(pretrained=True)
model2 = ImageClassifier()
model.eval()
model = model.to(device1)
torch.save(model.state_dict(),"alexnet_saved_cuda.pt")
state_dict = torch.load("alexnet_saved_cuda.pt", map_location=device2)
model2.load_state_dict(state_dict)
model2 = model2.to(device1)

```

ts/torch_handler/base_handler.py
HamidShojanazeri: @ngimel, I wonder if map_location moves the state_dict then model should be moved separately as well, as recommended [here](https://pytorch.org/tutorials/beginner/saving_loading_models.html#save-on-cpu-load-on-gpu)

ts/torch_handler/base_handler.py
```diff
@@ -44,14 +44,12 @@ def initialize(self, context):
...
-        self.map_location = "cuda" if torch.cuda.is_available() else "cpu"
-        self.device = torch.device(
-            self.map_location + ":" + str(properties.get("gpu_id"))
-            if torch.cuda.is_available()
-            else self.map_location
-        )
-        self.manifest = context.manifest
+        self.device = "cuda" if torch.cuda.is_available() else "cpu"
+        if torch.cuda.is_available():
+            torch.cuda.set_device(properties.get("gpu_id"))
```
cceyda, line 49: just an FYI:
If I'm running torchserve gpu version(in which case torch.cuda.is_available returns True) and choose to use `number_of_gpu=0` (temporarily for whatever reason)
`gpu_id` is set to None and it errors out because of "cuda:None"
before assigning device we should also check `gpu_id is not None`
I was gonna open a separate issue last week but maybe we can just solve it here?~
As @ngimel points out; "cuda" without id is suspect

HamidShojanazeri: Thanks, @cceyda we will check for this issue as well and good point, adding a check for gpu_id is needed as well.

fmassa:
FYI this is probably related to https://github.com/pytorch/vision/issues/2068 -- some of the oldest model checkpoints in torchvision have been serialized with an old version of the serialization script, which could be the culprit here.

Note that we have re-uploaded the checkpoints and updated the model links recently. This fix is present in the torchvision nigthlies and will be available in the next release

ts/torch_handler/base_handler.py
fmassa: Cross linking https://github.com/pytorch/serve/pull/1038#issuecomment-831823980 -- I think this might be fixed with latest torchvision and was due to the fact that those checkpoints were saved with a very old version of PyTorch serializer

fmassa:
Another update on this: there are still 3 models that we need to re-upload (VGG and inception), see https://github.com/pytorch/vision/issues/2473#issuecomment-831863505. We will be re-uploading them shortly and then we will let you know, it should fix those issues.

cc @NicolasHug

ts/torch_handler/base_handler.py
HamidShojanazeri: Thanks @fmassa for the comments and updates, I have replaced the updated weights form torchvision master and adding the changes in this [PR](https://github.com/pytorch/serve/pull/1053).

HamidShojanazeri:
Changes are merged into master through this [PR](https://github.com/pytorch/serve/pull/1053).

PR #1042, opened 2021-04-23
updated java dependencies to java 16

msaroufim:
JDK 16 is the highest version available on ubuntu and mac os x so updating installing dependencies

* Mac: https://github.com/AdoptOpenJDK/homebrew-openjdk
* Ubuntu: https://packages.ubuntu.com/hirsute/openjdk-16-jdk-headless

OS X version working fine locally, haven't tested the Ubuntu version yet

### DIFF
--- a/ts_scripts/install_dependencies.py
+++ b/ts_scripts/install_dependencies.py
@@ -60,7 +60,7 @@ class Linux(Common):
         os.system(f"{self.sudo_cmd}apt-get update")
 
     def install_java(self):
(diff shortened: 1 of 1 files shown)


msaroufim:
I just noticed that there is also a web UI so I'm guessing open JDK headless won't work for Ubuntu meaning to support Java 16 users have to go through some extra hoops to get started. I think from a usability standpoint this PR isn't worth it so closing for now.

JonathanSum:
So JDK 16 won't work? I am using Windows 10 here.
JDK 15 has been stopped for downloading
image

PR #1211, opened 2021-08-24
Updated README.md

AshwinRaikar88:
## Description

In step 2 of  Create torch-model-archiver from container
We will need to enter the bash as the root user else we won't be able to follow along with the given steps

Fixes #(issue)

## Type of change

Sucessfully served custom models with this docker image

## Checklist:
- [ ] Made corresponding changes to the documentation

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -212,7 +212,7 @@ docker ps
 2. Bind and get the bash prompt of running container
 
 ```bash
-docker exec -it <container_name> /bin/bash
+docker exec -it -u="root" <container_name> /bin/bash
 ```
 
 You will be landing at /home/model-server/.


docker/README.md
```diff
@@ -212,7 +212,7 @@ docker ps
 2. Bind and get the bash prompt of running container
 
 ```bash
-docker exec -it <container_name> /bin/bash
+docker exec -it -u="root" <container_name> /bin/bash
```
chauhang, line 215: @AshwinRaikar88 Can you please explain why the root user needs to be set here?

jagadeeshi2i: @AshwinRaikar88 both volumes are mounted with model-server as owner. So `-u root` is not needed.

Issue #1142, opened 2021-06-25
Arguments validation in the prediction model


andrey-chernykh:
## Is your feature request related to a problem? Please describe.

I want to validate input arguments of prediction, with the ability to say to the sender, what is incorrect.
The current behavior is to send HTTP 503 Error without any details.

## Describe the solution

1. Create a new type of Exceptions, for example, `InvalidArgumentError`. Using ValueError exceptions directly may cause backward incompatibility (maybe your prediction model in the deep is raising this exception?).
2. Add handler of this exception in the `ts/service.py` at the `predict` method
Something like:
```py
        except InvalidArgumentError as e:
            logger.warning("Invoking custom service failed, invalid arguments.", exc_info=True)
            return create_predict_response(None, req_id_map, e.message, 400)
```

## Describe alternatives solution

Reading the specification of HTTP Statuses - https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422 it may be better to use HTTP 422 error, because its syntax is correct, but values are incorrect.

Another way is to add something like `validate_entry_point` and call it before `entry_point` of the prediction model, but it's less understandable behavior and it's not backward compatible.

Another type of errors is bad syntax, for example I'm expecting `{"a": 1}`, but received `{"a":"1"}. In this situation it's better to send 400 Bad Request, but it requires `InvalidArgumentTypeError`. 

Let me know, what do you think about this feature. I can help with its implimentation

msaroufim:
@saurabheights-ecr what do you think?

PR #1225, opened 2021-09-01
Add InvalidArgumentException

valboldakov:
## Description

Closes https://github.com/pytorch/serve/issues/1142

This  PR intends to implement the enhancement from https://github.com/pytorch/serve/issues/1142

Use-case in a client code:

```
from ts.service import InvalidArgumentException
...
    val: int = int(data[0].get( "val"))
    if val < min_val or val > max_val:
        raise InvalidArgumentException(f"Requested val {val} is too big or too small.")
```


valboldakov:
@msaroufim @saurabheights-ecr

Would you please review it and tell me what do I need to enhance or fix?

msaroufim:
Looks good to me, the main thing I'd caution though is that for batch inference this may not be the desired behavior. You'd rather gracefully ignore an example rather than throw an exception but in this case since we're not forcing anyone to throw an exception unless they want to in a handler then this looks good

msaroufim:
Looks good to me, the main thing I'd caution though is that for batch inference this may not be the desired behavior. You'd rather gracefully ignore an example rather than throw an exception but in this case since we're not forcing anyone to throw an exception unless they want to in a handler then this looks good

lxning:
handler can throw different exception depending on different scenarios. This PR is not a generic solution, not able to cover different use cases. I filed a [ticket](https://github.com/pytorch/serve/issues/1309) to cover all use cases. So I'm going to close this PR.

valboldakov:
@lxning 

Thank you for your review!

Maybe I could change `InvalidArgumpythonentException` to `PredictionException` and provide error code to the interface?

So user's use-case will be:

```python
from ts.service import PredictionException

val: int = int(data[0].get( "val"))
if val < min_val or val > max_val:
    raise PredictionException(f"Requested val {val} is too big or too small.", 400)
```

And in the service code we'll:

```python
except PredictionException as argument_exception:
            logger.warning("Invoking custom service failed, invalid arguments.", exc_info=True)
            return create_predict_response(None, req_id_map, str(argument_exception), argument_exception.error_code))
```

What do you think?

cc @msaroufim @HamidShojanazeri

lxning:
@valboldakov Thank you for the reply.

The exception should be thrown from model's handler. Here is an [example](https://github.com/dhanainme/multi-model-server/blob/cd5a693032b1bec4c46b0f7a9844df496a62c1a8/frontend/modelarchive/src/test/resources/models/custom-return-code/service.py#L17). In other words, PredictionException message should be able to allow customization. I'll land the [code](https://github.com/awslabs/multi-model-server/pull/965/files) into TS.

Thanks again!

PR #1230, opened 2021-09-01
[WIP] Batch Processing Error Handling

msaroufim:
# Description

If there's a single problematic example in a batch, the Java frontend right now assumes that all the examples are problematic.

This code is a WIP that makes changes to `ModelWorkerResponse`, `BatchAggregator` and `WorkerThread` to fix this.

I still need an end to end test case after I figure out how to debug Java/Python code at once


msaroufim:
Closing this for now and we can revisit potentially later

PR #1240, opened 2021-09-08
feat: Add kubernetes HPA for torchserve

### LINKED ISSUE #714: Autoscaling TorchServe Questions and feature request
I am trying to deploy TorchServe in a kubernetes cluster and it seems hard to figure out how to autoscale this server to handle a variety of traffic load while not deploying maximum possible number of replicas to save money as well.

It seems like there is a Prometheus metrics framework with the v0.2 release and it has metrics like `ts_inference_latency_microseconds` , `ts_inference_requests_total` and `ts_queue_latency_microseconds`. These could be useful for autoscaling but the documentation does not explain how these metrics are measured, making it hard to use them. For example these do not say are these instantaneous measurements, measurements of the last request served, averaged measurements and if so then at what averaging interval. I would appreciate better documentation for the metrics.

For workloads that don't have a constant inference time (language-generation models), it is again hard to understand which one of these metrics would help the most with autoscaling. In our experience deploying models with TF-serve, we learnt that the best metric to autoscale is the queue depth, that is how many requests are waiting to be processed and are waiting in a queue. It seems like there is a similar metric `ts_queue_latency_microseconds` but that seems ill defined and seems to measure time rather than depth. Could you clarify what queue is this measuring the latency for? Is it the time waiting before handled by the Handler? If this is not the queue-wait-time, then could you please implement the queue wait time and queue depth as metrics so that we can use them for reliable autoscaling?

jagadeeshi2i:
## Description

Adds kubernetes HPA to torchserve tuned with torchserve's metrics.

Fixes #714 

## Type of change

- [ ] Test A
- [ ] Test B

- UT/IT execution results

- Logs

## Checklist:


chauhang (approved):

maaquib (approved):

kubernetes/hpa.yaml
```diff
@@ -0,0 +1,17 @@
...
+    kind: Deployment
+    name: torchserve
+  # autoscale between 1 and 10 replicas
+  minReplicas: 1
+  maxReplicas: 5
+  metrics:
+  - type: External
+    external:
+      metricName: ts_queue_latency_microseconds
+      targetValue: "7000000m"
```
dagsha, line 17: @chauhang I'm from Walmart. I have some questions on the metrics used for autoscaling.

Is there a way to enable the autoscale based on multiple set of (**metrics**, **metrics query** and **targetValue**) combinations?
Say, if the metrics used are _ts_queue_latency_microseconds_, _ts_inference_requests_total_, _CPUUtilization_ each with it's own targetValue and metrics query. Can we set the autoscale work if **any**of these metrics conditions are met?

jagadeeshi2i: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-multiple-metrics

PR #1249, opened 2021-09-21
Batch inferencing tests

msaroufim:
## Description

Fixes #(issue)

## Type of change

- [ ] Test A
- [ ] Test B

- UT/IT execution results

- Logs

## Checklist:

### DIFF
--- a/test/pytest/test_handler.py
+++ b/test/pytest/test_handler.py
@@ -92,7 +92,7 @@ def run_inference_using_url_with_data(purl=None, pfiles=None, ptimeout=120):
     else:
         return response
 
-def run_inference_using_url_with_data_json(purl=None, json_input=None, ptimeout=120):
+def run_inference_using_url_with_data_json(purl=None, pfiles=None, json_input=None, ptimeout=120):
     if purl is None and pfiles is None:
         return
     try:
@@ -223,6 +223,27 @@ def test_kfserving_mnist_model_register_and_inference_on_valid_model_explain():
     response = run_inference_using_url_with_data_json(KF_INFERENCE_API + '/v1/models/mnist:explain', data)
 
     assert np.array(json.loads(response.content)['explanations']).shape == (1, 1, 28, 28)
+    test_utils.unregister_model("mnist")
+
+def test_mnist_batch_inference():
+    test_utils.start_torchserve()
+
+    test_utils.register_model('mnist', 'mnist.mar')
+    files = {
+        'data': (data_file_mnist,
+                 open(data_file_mnist, 'rb')),
+    }
+
+    batch_size = 8
+    for _ in range(batch_size):
+        response = run_inference_using_url_with_data(TF_INFERENCE_API + '/predictions/mnist', files)
+
+    response = response.content.decode("utf-8")
+    response = ast.literal_eval(response)
+    response = [n.strip() for n in response]
+    
+    # We could instead assert len(response) == batch_size but this test will be flaky unless batch delay is about 10s
+    assert len(response) > 1
     test_utils.unregister_model("mnist")
     
     
--- a/test/pytest/test_utils.py
+++ b/test/pytest/test_utils.py
@@ -58,12 +58,15 @@ def torchserve_cleanup():
     delete_all_snapshots()
 
 
-def register_model(model_name, url):
+def register_model(model_name, url, batch_size=1, initial_workers=1):
     params = (
         ('model_name', model_name),
         ('url', url),
-        ('initial_workers', '1'),
+        ('initial_workers', str(initial_workers)),
         ('synchronous', 'true'),
+        ('batch_size', str(batch_size)),
+        ('batch_delay', '1')
+        
     )
     return register_model_with_params(params)
 


frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java
```diff
@@ -1068,6 +1068,55 @@ public void testPredictionMemoryError() throws InterruptedException {
...
+            HttpUtil.setContentLength(req, req.content().readableBytes());
+            req.headers()
+                .set(
+                        HttpHeaderNames.CONTENT_TYPE,
+                        HttpHeaderValues.APPLICATION_X_WWW_FORM_URLENCODED);
+        channel.writeAndFlush(req);
+
+        TestUtils.getLatch().await();
+        Assert.assertEquals(TestUtils.getHttpStatus(), HttpResponseStatus.ACCEPTED);
+        }
```
lxning, line 1112: Why these 4 inference requests can confirm batching is working?

test/pytest/test_handler.py
```diff
@@ -92,7 +92,7 @@ def run_inference_using_url_with_data(purl=None, pfiles=None, ptimeout=120):
     else:
         return response
 
-def run_inference_using_url_with_data_json(purl=None, json_input=None, ptimeout=120):
+def run_inference_using_url_with_data_json(purl=None, pfiles=None, json_input=None, ptimeout=120):
```
lxning, line 95: Why "pfiles" is needed? Will this change break existing tests?

frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java
```diff
@@ -1068,6 +1068,55 @@ public void testPredictionMemoryError() throws InterruptedException {
...
+        channel.close().sync();
+    }
+
+
+    @Test(
+            alwaysRun = true,
+            dependsOnMethods = {"testSuccessBatch"})
     public void testErrorBatch() throws InterruptedException {
```
lxning, line 1120: Why "testPredictionMemoryError" is replaced with "testSuccessBatch"?

frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java
```diff
@@ -1068,6 +1068,55 @@ public void testPredictionMemoryError() throws InterruptedException {
     @Test(
             alwaysRun = true,
             dependsOnMethods = {"testPredictionMemoryError"})
```
lxning, line 1070: Why this test case depends on "testPredictionMemoryError"?

test/pytest/test_handler.py
```diff
@@ -224,6 +224,27 @@ def test_kfserving_mnist_model_register_and_inference_on_valid_model_explain():
 
     assert np.array(json.loads(response.content)['explanations']).shape == (1, 1, 28, 28)
     test_utils.unregister_model("mnist")
+
+def test_mnist_batch_inference():
```
lxning, line 228: where is this test case called? And also what's the difference b/s test_mnist_batch_inference vs testSuccessBatch?

test/pytest/test_utils.py
```diff
@@ -58,12 +58,15 @@ def torchserve_cleanup():
...
+def register_model(model_name, url, batch_size=1, initial_workers=1):
     params = (
         ('model_name', model_name),
         ('url', url),
-        ('initial_workers', '1'),
+        ('initial_workers', str(initial_workers)),
         ('synchronous', 'true'),
+        ('batch_size', str(batch_size)),
+        ('batch_delay', '10000')
+
```
lxning, line 69: This part should not be changed if the old registerModel  (see following) is kept. 
public static void registerModel(
            Channel channel,
            String url,
            String modelName,
            boolean withInitialWorkers,
            boolean syncChannel)

frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java
msaroufim: Yeah thinking about this some more it doesn't confirm anything really. In the case of the Python backend I can create a a batch size counter in the preprocess handler and pass it all the way to the post process and then validate there that my batch size was correct. Not sure how to do the same for frontend yet

frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java
msaroufim: So the dependencies here just determine the order in which tests are ran, it's not a real dependency. All tests in this file are dependent on each other without needing it. We could perhaps have a a separate PR where we remove unnecessary dependencies but I believe this may be a known limitation because can't run more than one torchserve instance at a time so can't run more than one torchserve test at a time

test/pytest/test_handler.py
msaroufim: Purely a frontend vs backend test - test is called if you run pytest here

test/pytest/test_utils.py
msaroufim: my bad - was under the impression that you can add arguments to the end of a function with optional values to make sure no backwards compatbility is broken

msaroufim:
Closing for now

frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java
msaroufim: They're not dependent, I believe `dependsOnMethods` is used as a way to set the order in which tests are called

PR #1252, opened 2021-09-27
HF example update

HamidShojanazeri:
## Description
Updating HF example to support the latest Transformers. Currently example support Tracing a model and serving it on multi-gpus.

## Type of change

- [X ] Update an existing example and Transformers version in dev requirements. 

- Logs
https://gist.github.com/HamidShojanazeri/90e1135da5669af6ef2b05c92fa5e999


requirements/developer.txt
```diff
@@ -9,5 +9,5 @@ nvgpu==0.8.0; sys_platform == 'win32'
 grpcio
 protobuf
 grpcio-tools
-transformers==4.6.0
+transformers==4.11.0
```
lxning, line 12: Does this new version still support the old api (ie. outputs = self.model(input_batch))?

lxning: right

chauhang: @HamidShojanazeri If this is a breaking change, please include some details in the Readme. It will be good to also check with HF team if there can be a workaround so that people don't need to change code for things to work

Issue #1195, opened 2021-08-11
TorchServe should not overwrite /usr/bin/python and /usr/bin/pip


sharvil:
https://github.com/pytorch/serve/blob/cdca64c821d4106b752a5e9421e205b685afbfda/ts_scripts/install_dependencies.py#L71-L72

Under no circumstances should an install script silently and without user consent forcibly relink system binaries.

AdityaSoni19031997:
This has messed up my env, thanks for that.

sharvil:
Is it possible to increase the priority on this bug? Silently breaking the systemwide Python on a user's machine is a very severe outcome.

AdityaSoni19031997:
It broke my env, thanks 🙏. Took me a week to figure out what has happened and what caused it.

PR #1254, opened 2021-09-28
Remove forceful links and silent upgrades in `install_dependencies.py`

msaroufim:
Fixes https://github.com/pytorch/serve/issues/1195

Goal of this PR is to remove all silent upgrades on local machines or forceful links since for non fresh machines this could easily break code - mostly by either deleting code or gating behind `os.system("package --version") != 0 or args.force`

TODO
* What's up with the `brew link/unlink`: seems like this is standard practice when working with multiple versions of a package https://stackoverflow.com/a/54175781
* Is curling the nodejs installation script necessary if we have apt-get?
* Relinking python seems strange (was this a CI convenience?) because we do expect people to install dependencies within some conda/pip environment before calling `pip install torchserve`


ts_scripts/install_dependencies.py
```diff
@@ -68,8 +68,6 @@ def install_nodejs(self):
         python_path = Path(sys.executable).resolve()
         os.system(f"{self.sudo_cmd}curl -sL https://deb.nodesource.com/setup_14.x | {self.sudo_cmd}bash -")
         os.system(f"{self.sudo_cmd}apt-get install -y nodejs")
```
sharvil, line 70: The remainder of the code in here is extremely dangerous as well. First, running a shell script off the internet as root is a major security vulnerability. Second, force-upgrading `node` on a system can silently break other applications / services in the same way that force-relinking `python` can.

A little further down, `wget` is also force-upgraded, which is again potentially system-breaking.

Since we're already in here, can we remove everything that silently / forcibly overwrites system binaries?

msaroufim: Hi @sharvil would you be interested in making a PR? I can make sure it works but your perspective would be invaluable

msaroufim:
@nskool @maaquib could you also review? This has been a frustrating issue for users doing development on their local machines

ts_scripts/install_dependencies.py
```diff
@@ -72,22 +72,20 @@ class Linux(Common):
 
     def __init__(self):
         super().__init__()
-        os.system(f"{self.sudo_cmd}apt-get update")
+        # os.system(f"{self.sudo_cmd}apt-get update")
 
     def install_java(self):
         os.system(f"{self.sudo_cmd}apt-get install -y openjdk-11-jdk")
```
sharvil, line 78: I think you missed a check for this one

sharvil:
Thanks for making this change, @msaroufim. Speaking on what I'd want as a Linux user: I recommend the script notify the user of all dependencies that are going to be installed and getting consent to run them before executing `sudo apt-get ...`.

Here's what I'm imagining:
```
Warning: this script will install the following packages to your system using `apt`
if they are not already installed and in your $PATH:
  - wget
  - nodejs
  - openjdk-11-jdk

Do you want to continue (y/N): 
```

Waiting for user consent is important because if they had recently run a `sudo` command, they won't even get a password prompt.

msaroufim:
I think that's fair 

> Thanks for making this change, @msaroufim. Speaking on what I'd want as a Linux user: I recommend the script notify the user of all dependencies that are going to be installed and getting consent to run them before executing `sudo apt-get ...`.
> 
> Here's what I'm imagining:
> 
> ```
> Warning: this script will install the following packages to your system using `apt`
> if they are not already installed and in your $PATH:
>   - wget
>   - nodejs
>   - openjdk-11-jdk
> 
> Do you want to continue (y/N): 
> ```
> 
> Waiting for user consent is important because if they had recently run a `sudo` command, they won't even get a password prompt.

I think is great feedback. I'm thinking `install_dependencies.py` needs some `-y` flag that would be enabled for CI and disabled by default

ts_scripts/install_dependencies.py
```diff
@@ -175,20 +177,11 @@ def get_brew_version():
...
-        choices=['prod', 'dev'],
-        help=
-        "environment(production or developer) on which dependencies will be installed"
-    )
-
+    parser = argparse.ArgumentParser(description="Install various build and test dependencies of TorchServe")
+    parser.add_argument('--cuda', default=None, choices=['cu92', 'cu101', 'cu102', 'cu111'], help="CUDA version for torch")
+    parser.add_argument('--environment', default='prod', choices=['prod', 'dev'],
+                        help="environment(production or developer) on which dependencies will be installed")
+    parser.add_argument("--reinstall_dependencies", default='no', choices=['yes', 'no'], help="force reinstall dependencies")
```
nskool, line 184: Since this is a boolean value, use action="store_true", instead of specifying 'yes', 'no'.

ts_scripts/install_dependencies.py
```diff
@@ -72,22 +72,22 @@ class Linux(Common):
 
     def __init__(self):
         super().__init__()
-        os.system(f"{self.sudo_cmd}apt-get update")
+        if args.reinstall_dependencies == "yes":
+            os.system(f"{self.sudo_cmd}apt-get update")
```
nskool, line 76: Is there any advantage doing this change? It seems like this the update command won't run if reinstall_dependencies is 'no'?

msaroufim: correct, I think running an `apt-get update` on a user machine is a tad aggressive so this is deliberate

PR #1265, opened 2021-10-04
remove mention of s3 in server.md

### LINKED ISSUE #1228: Better document s3 support ("unknown protocol: s3")
## 📚 Documentation

It's unclear from the documentation that (to my knowledge) only s3 via http is supported.  In fact, the documentation on https://pytorch.org/serve/server.html seems to indicate the s3 protocol is supported:

> Model path can be mar file name in model store or URI (s3 link, or http link). s3 link: s3://S3_endpoint[:port]/… 

but if you try to use an s3:// uri you get:
```
Caused by: java.net.MalformedURLException: unknown protocol: s3
```

msaroufim:
Fixes #1228


lxning (approved):

chauhang (dismissed):
Please see inline

PR #1244, opened 2021-09-15
[WIP] Update BatchAggregator.java

msaroufim:

Issue #1234, opened 2021-09-03
Load model failed: mymodel, error: Worker died


ovlia:
## 📚 Documentation

2021-09-03 14:26:44,745 [WARN ] W-9000-mymodel1_1.0-stderr MODEL_LOG - model_service_worker.py: error: unrecognized arguments: --sock-type unix --sock-name /tmp/.ts.sock.9000
2021-09-03 14:26:44,745 [INFO ] epollEventLoopGroup-5-7 org.pytorch.serve.wlm.WorkerThread - 9000 Worker disconnected. WORKER_STARTED
2021-09-03 14:26:44,745 [DEBUG] W-9000-mymodel1_1.0 org.pytorch.serve.wlm.WorkerThread - System state is : WORKER_STARTED
2021-09-03 14:26:44,745 [DEBUG] W-9000-mymodel1_1.0 org.pytorch.serve.wlm.WorkerThread - Backend worker monitoring thread interrupted or backend worker process died.
java.lang.InterruptedException
        at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.awaitNanos(AbstractQueuedSynchronizer.java:1668)
        at java.base/java.util.concurrent.ArrayBlockingQueue.poll(ArrayBlockingQueue.java:435)
        at org.pytorch.serve.wlm.WorkerThread.run(WorkerThread.java:188)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
        at java.base/java.lang.Thread.run(Thread.java:831)
2021-09-03 14:26:44,746 [WARN ] W-9000-mymodel1_1.0 org.pytorch.serve.wlm.BatchAggregator - Load model failed: mymodel1, error: Worker died.

toretak:
Hi @ovlia, please provide more information about the model. Is it default or custom handler? Are you running TorchServe in docker, or locally? Is it latest version of torchserve? Is the version of torchserve and model-archiver the same?

ovlia:
running TorchServe in docker, latest version of torchserve 0.4.2

HamidShojanazeri:
@ovlia Can you please share more details on the model and handler and full model_log, this would help us better understand the context and suggest solutions.

msaroufim:
[...]

PR #1272, opened 2021-10-06
Bert base batch test

msaroufim:
## Goal

@nskool @HamidShojanazeri @lxning this is ready for review

This is a subset PR of the scope of https://github.com/pytorch/serve/pull/1249 which I will most likely close

This is the first test for batch inferencing in our repo

## Repro instructions
To test run  `pytest test/pytest/test_handler.py -k test_huggingface_bert_batch_inference 2>&1 | tee regression.txt`

Test passes

```
(conc) ubuntu@ip-172-31-0-52:~/serve$ pytest test/pytest/test_handler.py -k test_huggingface_bert_batch_inference  2>&1 | tee regression.txt
============================= test session starts ==============================
platform linux -- Python 3.9.7, pytest-6.2.5, py-1.10.0, pluggy-0.13.1
rootdir: /home/ubuntu/serve
plugins: mock-3.6.1, cov-3.0.0
collected 9 items / 8 deselected / 1 selected

test/pytest/test_handler.py .                                            [100%]

================= 1 passed, 8 deselected in 177.50s (0:02:57) ==================
```

Can also inspect `model_log.log` to see that a batched inference is indeed happening

Handler is available here `https://bert-mar-file.s3.us-west-2.amazonaws.com/BERTSeqClassification.mar'`

```
2021-10-07 00:29:49,365 [INFO ] W-9000-BERTSeqClassification_1.0-stdout MODEL_LOG - INFERENCE LIST
2021-10-07 00:29:49,365 [INFO ] W-9000-BERTSeqClassification_1.0-stdout MODEL_LOG - ['Not Accepted', 'Not Accepted']
2021-10-07 00:29:49,365 [INFO ] W-9000-BERTSeqClassification_1.0-stdout MODEL_LOG - LENGTH OF INFERENCE
2021-10-07 00:29:49,365 [INFO ] W-9000-BERTSeqClassification_1.0-stdout MODEL_LOG - 2
```

If you just run the new test without updating any java code then you will get this error

```
java.util.ConcurrentModificationException
        at java.base/java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:719)
        at java.base/java.util.LinkedHashMap$LinkedKeyIterator.next(LinkedHashMap.java:741)
        at org.pytorch.serve.wlm.BatchAggregator.sendResponse(BatchAggregator.java:81)
        at org.pytorch.serve.wlm.WorkerThread.run(WorkerThread.java:194)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at java.base/java.lang.Thread.run(Thread.java:829)
```

This error was resolved here https://github.com/pytorch/serve/pull/1244 so we are merging both PRs

## Test on multiple clients
https://docs.google.com/document/d/1fLcdAFJyDMHSq6QkJSrbVsNZDGb9NnCA/edit?usp=sharing&ouid=106222972308395582904&rtpof=true&sd=true

Branch was tested and worked independently for someone else as well https://github.com/pytorch/serve/issues/1283#issuecomment-945098138

## Other solutions that didn't work out

We could potentially revisit this as we create more batched inference tests but for now I'd suggest we go with this so the concurrency bug is fixed and so we have some test for batch inferencing.

### Async python
Using async python which likely has some bug preventing us still from making multiple concurrent inferences

```
    async def run_async_inference():
        await run_inference_using_url_with_data(TF_INFERENCE_API + '/v1/models/BERTSeqClassification:predict', pfiles=files, ptimeout=1000)


    loop = asyncio.get_event_loop()
    tasks = []
    for i in range(batch_size - 1):
        task = loop.create_task(run_async_inference())
        tasks.append(task)
    
    loop.run_until_complete(asyncio.gather(*tasks))
```

### Multi processing
Or using Python multiprocessing was returning an empty response back

```
   def invoke():
        res = run_inference_using_url_with_data(TF_INFERENCE_API + '/v1/models/BERTSeqClassification:predict', pfiles=files, ptimeout=1000)                                                                        
        return res

    pool = multiprocessing.Pool(batch_size)
    results = pool.map(invoke, batch_size*[])
    pool.close()
    pool.join()
    print(results)
```

### DIFF
--- a/frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java
+++ b/frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java
@@ -1068,32 +1068,6 @@ public class ModelServerTest {
     @Test(
             alwaysRun = true,
             dependsOnMethods = {"testPredictionMemoryError"})
-    public void testSuccessBatch() throws InterruptedException {
-        int batchSize = 4;
-        int maxBatchDelay = 10000;
-        Channel channel = TestUtils.connect(ConnectorType.MANAGEMENT_CONNECTOR, configManager);
-        Assert.assertNotNull(channel);
-
-        TestUtils.setHttpStatus(null);
-        TestUtils.setResult(null);
-        TestUtils.setLatch(new CountDownLatch(1));
-
-        TestUtils.registerModel(
-                channel, "noop.mar", "noop", true, false, batchSize, maxBatchDelay);
-        TestUtils.getLatch().await();
-
-        StatusResponse status =
-                JsonUtils.GSON.fromJson(TestUtils.getResult(), StatusResponse.class);
-        Assert.assertEquals(
-                status.getStatus(),
-                "Model \"noop\" Version: 1.0 registered with 1 initial workers");
-
-        channel.close().sync();
-    }
-
-    @Test(
-            alwaysRun = true,
-            dependsOnMethods = {"testSuccessBatch"})
     public void testErrorBatch() throws InterruptedException {
         Channel channel = TestUtils.connect(ConnectorType.MANAGEMENT_CONNECTOR, configManager);
         Assert.assertNotNull(channel);


frontend/server/src/main/java/org/pytorch/serve/wlm/BatchAggregator.java
```diff
@@ -103,20 +105,20 @@ public void sendError(BaseModelRequest message, String error, int status) {
...
                 jobs.clear();
-                logger.error("Not all jobs get response.");
+                logger.error("Not all jobs got an error response.");
             }
         } else {
             // Send the error message to all the jobs
             for (Map.Entry<String, Job> j : jobs.entrySet()) {
                 String jobsId = j.getValue().getJobId();
-                Job job = jobs.remove(jobsId);
+                Job job = jobs.get(jobsId);
```
HamidShojanazeri, line 121: @msaroufim do we have any test to repro the batch aggregator issue?

test/pytest/test_handler.py
```diff
@@ -224,7 +224,28 @@ def test_kfserving_mnist_model_register_and_inference_on_valid_model_explain():
 
     assert np.array(json.loads(response.content)['explanations']).shape == (1, 1, 28, 28)
     test_utils.unregister_model("mnist")
+
+def test_huggingface_bert_batch_inference():
+    batch_size = 2
+    batch_delay = 10000 # 10 seconds
+    params = (
+        ('model_name', 'BERTSeqClassification'),
+        ('url', 'https://bert-mar-file.s3.us-west-2.amazonaws.com/BERTSeqClassification.mar'),
```
HamidShojanazeri, line 233: @msaroufim We need to change the url to mar file to the one in Torchserve s3 bucket

msaroufim: Fixed

msaroufim:
@lxning @HamidShojanazeri 2 different tickets are open with this issue https://github.com/pytorch/serve/issues/1234 and https://github.com/pytorch/serve/issues/1283 can we merge it in?

frontend/server/src/main/java/org/pytorch/serve/wlm/BatchAggregator.java
lxning: @msaroufim [ModelServerTest.java#L1071](https://github.com/pytorch/serve/blame/master/frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java#L1071) is a frontend integration test case which is called during frontend build.

Test test_huggingface_bert_batch_inference is an end 2 end integration which is called by regression test.

So they are different level test. Usually, a java test case should be added into dir frontend/server/src/test if there are frontend source code changes. Currently frontend misses a batch positive test case.

It is fine if you prefer adding batch end2end test cases in regression test. Then, I suggest you add both positive and negative test cases at here. Otherwise, the problem at regression for batch use case is as same as current Java frontend test (ie. miss a test case).

HamidShojanazeri (approved):

PR #1297, opened 2021-10-27
PyTorch profiler integration

shrinath-suresh:
Signed-off-by: Shrinath Suresh <shrinath@ideas2it.com>

## Description

This pull request contains changes in base handler to integrate PyTorch profiler - https://pytorch.org/blog/introducing-pytorch-profiler-the-new-and-improved-performance-tool/  with torchserve

When user opts for the profiler, the preprocess/inference/postprocess calls are profiled and the traces are saved as a json file . User can load the trace json into pytorch profiler and view the detailed output.

This feature will help users to debug the issues in preprocess  or inference. 

the basis structure of the implementation is in place. More detailed requirements and unit tests to be captured. once it is done, WIP tag will be removed.

Initial Requirements:

1. The profiler integration is done only for the custom handler. User can opt for the profiler flow by setting the class variable.
2. User can override torch.profiler.profiler arguments - This helps user to add conditions based on the environment. For example: by default, the profiler is set to trace only CPU. In case, user wants to profiler CUDA activity as well, profiler arguments can be overridden using class variable.
3. Saving the profiler output and ability to load the profiler output using pytorch profiler plugin - https://github.com/pytorch/kineto  

Fixes #(issue)

## Type of change

- [ ] Test A
- [ ] Test B

- UT/IT execution results

Yet to be added

- Logs

## Checklist:


yqhu (approved):
Looks good.

test/pytest/resnet_profiler_override.py
```diff
@@ -0,0 +1,12 @@
+from ts.torch_handler.image_classifier import ImageClassifier
+from torch.profiler import ProfilerActivity, tensorboard_trace_handler
+
+class ResnetHandler(ImageClassifier):
+    def __init__(self):
+        super().__init__()
+        print("<<<<<<<<<<<<<<<<<< Custom handler invoked \n\n\n\n\n\n\n")
```
shrinath-suresh, line 7: will remove the print

test/pytest/test_model_archiver.py
```diff
@@ -19,37 +19,8 @@ def teardown_module(module):
     test_utils.torchserve_cleanup()
 
 
-def model_archiver_command_builder(model_name=None, version=None, model_file=None, serialized_file=None, handler=None, extra_files=None, force=False):
```
shrinath-suresh, line 22: moved this method to `test_utils.py` to reuse

ts/torch_handler/base_handler.py
```diff
@@ -194,18 +199,50 @@ def handle(self, data, context):
...
 
+    def infer_with_profiler(self, data):
+        # Setting the default profiler arguments to profile cpu usage and record shapes
+        # User can override this argument based on the requirement . Ex: Profiling Gpu
+        if not self.profiler_args:
+            self.profiler_args["activities"] = [ProfilerActivity.CPU]
+            self.profiler_args["record_shapes"] = True
+
+        if "on_trace_ready" not in self.profiler_args:
+            result_path = "/tmp/pytorch_profiler"
```
HamidShojanazeri, line 226: @shrinath-suresh I wonder if we give this option to user to choose the result_path and if not provided we fall back to default path.

HamidShojanazeri (commented):
Thanks @shrinath-suresh, it looks great.  I think adding some notes to docs probably in configuration would be helpful.

ts/torch_handler/base_handler.py
HamidShojanazeri: Also adding/extending one of the examples, like Resnet or BERT on using this profilers would be very helpful to show the usage.

ts/torch_handler/base_handler.py
```diff
@@ -194,18 +199,50 @@ def handle(self, data, context):
         self.context = context
         metrics = self.context.metrics
 
-        data_preprocess = self.preprocess(data)
-
-        if not self._is_explain():
-            output = self.inference(data_preprocess)
-            output = self.postprocess(output)
+        is_profiler_enabled = os.environ.get("ENABLE_TORCH_PROFILER", None)
```
HamidShojanazeri, line 202: @shrinath-suresh I wonder if this could be added to the config.properties, this would help to make sure all the configs are gathered in one place. An example of adding to the config can be found [here](https://github.com/pytorch/serve/pull/1319/files).

shrinath-suresh:
> Thanks @shrinath-suresh, it looks great. I think adding some notes to docs probably in configuration would be helpful.

Thanks for your review comments. I will work on addressing them

ts/torch_handler/base_handler.py
shrinath-suresh: @HamidShojanazeri Sure Hamid.. I will pick any one example (mnist or resnet) to showcase profiler integration and update the readme accordingly. Thanks for the suggestion

ts/torch_handler/base_handler.py
shrinath-suresh: @chauhang your thoughts on this ?

ts/torch_handler/base_handler.py
shrinath-suresh: @HamidShojanazeri Regarding the profiler path.. By default the outputs get stored into "/tmp/pytorch_profiler". User can override this behaviour like [this](https://github.com/shrinath-suresh/serve/blob/c5568731c1dcac9859e05375938e811eaab3cb56/test/pytest/profiler_utils/resnet_profiler_override.py#L11). 

@HamidShojanazeri Do you have any suggestions for default path ? At `base_handler.py` i couldn't find any default path (model store?) to store profiler trace files. `/tmp/pytorch_profiler` looks hardcoded.. Any suggestion ?

ts/torch_handler/base_handler.py
msaroufim: I agree with Hamid, it's a simple change that'll make sure people don't need to think about both OS environment variables and config.properties. It's how we manage all configs. Let us know if you have any questions about how this would work

ts/torch_handler/base_handler.py
msaroufim: For the profiler path I'd suggest using the same path the pytorch profiler uses by default. If 2 succesive runs of a profile just create new directories in the `/tmp/pytorch_profiler` then there is no need to overwrite but if they do then it's best to allow users to configure what the output path is

msaroufim (changes requested):
Some minor feedback is left, we can merge this whenever all questions are addressed. Thank for your contribution - this is good stuff!

shrinath-suresh:
> Some minor feedback is left, we can merge this whenever all questions are addressed. Thank for your contribution - this is good stuff!

I will work on addressing the review comments.

ts/torch_handler/base_handler.py
```diff
@@ -194,18 +199,50 @@ def handle(self, data, context):
...
 
         stop_time = time.time()
         metrics.add_time('HandlerTime', round((stop_time - start_time) * 1000, 2), None, 'ms')
         return output
 
+    def infer_with_profiler(self, data):
+        # Setting the default profiler arguments to profile cpu usage and record shapes
+        # User can override this argument based on the requirement . Ex: Profiling Gpu
+        if not self.profiler_args:
+            self.profiler_args["activities"] = [ProfilerActivity.CPU]
```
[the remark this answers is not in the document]
shrinath-suresh: makes sense.. Updated the same

shrinath-suresh:
For the profiler path I'd suggest using the same path the pytorch profiler uses by default. If 2 succesive runs of a profile just create new directories in the /tmp/pytorch_profiler then there is no need to overwrite but if they do then it's best to allow users to configure what the output path is

@msaroufim I have updated the code to generate output based on the model name.

For example:
`/tmp/pytorch_profiler/mnist`
`/tmp/pytorch_profiler/resnet`

It would help to identify the trace files when multiple models are served at once.
users can always override this path in the custom handler - I have added a unit test for the same. Example here - https://github.com/pytorch/serve/pull/1297/files#diff-3881efebc0822a8b104d109f9123b95006816aa0f666fd0ca78bf8c9989d30c0 

Initially i tried logging into the present working directory using (`os.getcwd()). Unfortunately, TorchServe isolates all the dependent files into a temp directory and serves the model. `os.getcwd points to the tmp location which will of be no use to the user and difficult to locate as well.

At the Base Handler level (base_handler.py) i couldnt find any variable (model store path or something )to store the output, which forced me to set the path as `/tmp/pytorch_profiler`. If you find any better path, please feel free to suggest. 

if we are going to add pytorch profiler into config.properties, should we make the output path as a mandatory parameter ?

ts/torch_handler/base_handler.py
HamidShojanazeri: > I will pick any one example (mnist or resnet) to showcase profiler integration and update the readme accordingly. Thanks for the suggestion

It would be great. would you be able to include it in this PR or it would be a separate one.

HamidShojanazeri (approved):
Thanks @shrinath-suresh, discussed with @chauhang offline, we will move forward with current setting for env variable. Just it would be great to have example in separate PR as well.

msaroufim:
Hi @shrinath-suresh thank you for your contribution - the PR looks almost ready to go. The main thing that needs to be fixed next is using `config.properties` instead of environment variables.

Otherwise I'd also suggest you create an example in `examples/` with a list of instructions that would explain how users can configure the profiler and where in the code they should look at and what kind of output they should get @chauhang

shrinath-suresh:
@HamidShojanazeri  and @msaroufim Thanks for the review. I was just waiting for the decision on env variable to avoid rework. I am working on updating the examples and docs . Will update the PR soon.

examples/image_classifier/mnist/README.md
```diff
@@ -34,6 +34,42 @@ Run the commands given in following steps from the parent directory of the root
...
 
+# Profiling inference output
+
+To enable pytorch profiler, set the following environment variable.
+
+```
+export ENABLE_TORCH_PROFILER=true
+```
+
+Note: Ensure`enable_envvars_config` is set to `true` in torchserve config
```
HamidShojanazeri, line 45: @shrinath-suresh is this gonna be added to the config.properties?

examples/image_classifier/mnist/README.md
```diff
@@ -34,6 +34,42 @@ Run the commands given in following steps from the parent directory of the root
...
+By default the pytorch profiler trace files are generated under "/tmp/pytorch_profiler/<model_name>" directory.
+
+The path can be overridden by setting `on_trace_ready` parameter in `profiler_args` - [Example here](../../../test/pytest/profiler_utils/resnet_profiler_override.py)
+
+And the trace files can be loaded in tensorboard using torch-tb-profiler. Check the following link for more information - https://github.com/pytorch/kineto/tree/main/tb_plugin 
+
+Install torch-tb-profiler and run the following command to view the results in UI
+
+```
+tensorboard --logdir /tmp/pytorch_profiler/mnist/
```
HamidShojanazeri, line 66: @shrinath-suresh in the handler below I think, "on_trace_ready" to override the path is missing.

shrinath-suresh:
@HamidShojanazeri and @msaroufim I have updated the MNIST example with Profiler information. Updated the README with screenshot as well. Can you please review and let me know if anything else needs to be added ?

examples/image_classifier/mnist/README.md
shrinath-suresh: on_trace_ready is not a mandatory argument. If user provides the argument, the trace files will get generated into the location. if the user doesn't provide the argument, the trace files get generated inside "/tmp/pytorch_profiler/<model_name>" folder. Its handled [here](https://github.com/shrinath-suresh/serve/blob/19a521189e574f07cd465cde8019b59ee73d3c13/ts/torch_handler/base_handler.py#L245)

For demonstration, we can add on_trace_ready in handler file. Again, we cannot set the `on_trace_ready` to `os.getcwd()` as it will be a temp directory in run time. May be we can try use home directory or something .. Your thoughts ?

examples/image_classifier/mnist/README.md
shrinath-suresh: I added a note here, since the env based setup doesnt work unless user sets `enable_envvars_config` to `true` in config.properties. 

I will add `config.properties` file in this example and update the instructions accordingly.

HamidShojanazeri: @shrinath-suresh as mentioned in previous comment, you can add a setting to the config following [this example](https://github.com/pytorch/serve/pull/1319/files). This basically would not work for now.

HamidShojanazeri (commented):
@shrinath-suresh please trying adding the config using the example shared in the comments.

PR #1314, opened 2021-11-10
Fix #1304: use simplejson instead of json in JSONEnvelope

### LINKED ISSUE #1304: non-JSON output of out-of-range float values (nan, inf, -inf)
## Context

The Python handler returns invalid response body (non-JSON) in case the inference result dict contains `nan`, `inf` or `-inf` values. This is due to Python's `json` encoder actually **not** returning valid JSON in these cases, unless the `allow_nan=False` argument is passed when calling `json.dumps`.

* torchserve version: 0.4.2

## Expected Behavior
When the handler returns a Python dict such as this `{"number": nan}`, the response should be this JSON: `{"number": null}`.

## Current Behavior
The response is  `{"number": Nan}`, which is not JSON.

## Possible Solution
Add `allow_nan=False` argument to the `json.dumps` call at https://github.com/pytorch/serve/blob/master/ts/torch_handler/request_envelope/json.py#L73.

## Steps to Reproduce
1. Override the BaseHandler to return a dict containing out-of-range float values.
2. Run a request
3. Observe non-JSON responses

vvekic:
## Description

Fixes #1304 by serializing with `simplejson` and its `ignore_nan` argument, which properly converts Python `nan`, `inf` and `-inf` values to JSON `null`. `simplejson` is also added to the main requirements list.

## Type of change

Failing test case for this issue wasn't present. It is added in this PR in `ts/torch_handler/unit_tests/test_envelopes.py`.
Confirmed passing by running `/workspaces/serve/ts/torch_handler/unit_tests/run_unit_tests.sh`.


msaroufim:
Very cool! PR looks fine overall, let me try it out and keep you posted

requirements/common.txt
```diff
@@ -2,6 +2,7 @@ future
 psutil
 wheel
 requests
+simplejson
```
msaroufim, line 5: let's just double check that other requirements.txt also need this or not and whether json is used anywhere else in the codebase

ts/torch_handler/unit_tests/test_envelopes.py
```diff
@@ -46,6 +56,17 @@ def test_json_batch(handle_fn, model_context):
     results = envelope.handle(test_data, model_context)
     assert(results == expected_result)
 
+def test_json_nan(identity_handle_fn, identity_model_context):
+    test_data = [{'body':{
+        'instances': [[2.0, float("NaN")], [5.0, 4.0]]
```
msaroufim, line 61: mind also adding a test for `inf`? you can put it here

msaroufim (changes requested):

requirements/common.txt
vvekic: i wasn't sure where to put this, so i chose common.txt. it's unlikely simpljson is used elsewhere. if that's the case - where do you propose to move this?

msaroufim (approved):

msaroufim:
@lxning what do you think, should we merge this? LGTM

maaquib:
@vvekic Thanks for this
We should just make sure there isn't any significant perf difference with using `simplejson` with larger payloads

msaroufim:
While I do appreciate the benefits that come with better data imputation, my concern with using non standard JSON packages is performance and long term maintenance. So for now will close this PR and we can revisit if you feel strongly about it

Issue #1322, opened 2021-11-15
Prediction request timeout including time spent in queue


nateagr:
## Is your feature request related to a problem? Please describe.
The feature request is not related to a bug. But at my company, we have this following problem: we would like to be able to dismiss prediction requests after a certain amount of time. Why ? Because they are not relevant anymore after this amount of time. As of today, it is possible to define a timeout, responseTimeout (https://pytorch.org/serve/configuration.html), but it only applies to model inference (handler). It does not include the time spend in queue.

## Describe the solution
To avoid any regression, I would suggest to add a new model parameter: predictionRequestTimeout. This parameter defines the timeout after which a prediction request is dismissed. It includes the time spend in queue and inference time. From an implementation point of view, I would suggest to modify:
- pollBatch method of Model (https://github.com/pytorch/serve/blob/master/frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java#L174) to dismiss prediction requests that are still in queue and too old wrt to predictionRequestTimeout.
- run method of WorkerThread (https://github.com/pytorch/serve/blob/master/frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerThread.java#L189) to timeout after min(responseTimeout, remaining time of predictionRequestTimeout) instead of after responseTimeout.

If you agree with this feature requests, I'm willing to push a PR.

nateagr:
Hello there! Should I suggest a PR ? Or do you need time to think about it ?

HamidShojanazeri:
@nateagr please go a head with a PR, we would appreciate it and happy review and help to merge.

PR #1341, opened 2021-11-22
Introduce queue timeout for prediction requests

nateagr:
## Description

This PR follows this feature request https://github.com/pytorch/serve/issues/1322

The feature request is not related to a bug. But at my company, we have this following problem: we would like to be able to dismiss prediction requests after a certain amount of time. Why ? Because they are not relevant anymore after this amount of time. As of today, it is possible to define a timeout, responseTimeout (https://pytorch.org/serve/configuration.html), but it only applies to model inference (handler). It does not include the time spent in queue. This PR adds the possibility to define a maximum amount of time spent in queue. Combined with responseTimeout (time spent in Python for inference), the total duration correspond to the maximum amount of time before a prediction request is stale. Instead of creating one parameter that would include the time spent in queue and the inference time, I've decided to add a maximum duration in queue for the following reasons:
1. The new parameter does not interfere with the existing parameter responseTimeout, that I don't want to modify to not add regressions for users already using this parameter.
2. The latency of a prediction request includes the time spent in queue and inference time. For inference, predictions requests are grouped in batches. It is not clear what duration to use to timeout inference given a batch of prediction requests that have potentially spent different amounts of time in queue, and thus having a different remaining time budget.
3. In my opinion, splitting the maximum latency of a prediction request in two (maximum duration in queue and maximum duration for inference) gives more control to users and is more flexible. Users normally now inference time (at least we have a rough idea). Let's say I've put a model in production that takes ~300ms for inference and prediction requests are stale after 500ms. I don't want torchserve to send prediction requests that have spent more than 200ms in queue to Python for inference (because predictions will be stale when available). So instead, I prefer telling torchserve to dismiss any prediction requests that were in queue for more than 200ms.

## Type of change

I've added a simple unit test. I did not find tests that test this kind of small feature so I'm opened to any suggestion/advice.

## Checklist:

### DIFF
--- a/.github/workflows/ci_cpu.yml
+++ b/.github/workflows/ci_cpu.yml
@@ -1,6 +1,13 @@
 name: CI CPU
 
-on: push
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
 concurrency:
   group: ci-cpu-${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.run_number || github.ref }}
   cancel-in-progress: true
--- a/.github/workflows/ci_gpu.yml
+++ b/.github/workflows/ci_gpu.yml
@@ -1,6 +1,13 @@
 name: CI GPU
 
-on: push
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
 concurrency:
   group: ci-gpu-${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.run_number || github.ref }}
   cancel-in-progress: true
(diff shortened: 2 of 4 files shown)


nateagr:
Do I need an AWS account to fix build issues ? I don't have one unfortunately

HamidShojanazeri:
Thanks @nateagr we will review the PR and get back to you.

nateagr:
> Thanks @nateagr, you don't need AWS account to fix the build issues, are you able to see the Build logs above?

Yes I do. Thanks!

maaquib:
@nateagr Thanks for adding this feature. Can you also add a negative test case to the `ModelServerTest` suite

PR #1397, opened 2022-01-12
Regression test fix - PyTorch Profiler tests

shrinath-suresh:
## Description

Regression tests failing with the below mentioned stack trace

```

set_custom_handler = None, handler_name = '/home/ubuntu/serve/test/pytest/../../test/pytest/profiler_utils/resnet_profiler_override.py'

    @pytest.mark.parametrize(
        "handler_name",
        [os.path.join(profiler_utils, "resnet_profiler_override.py")],
    )
    def test_batch_input(set_custom_handler, handler_name):
        """
        Tests pytorch profiler integration with batch inference
        """
        CUSTOM_PATH = "/tmp/output/resnet-152-batch"
        if os.path.exists(CUSTOM_PATH):
            shutil.rmtree(CUSTOM_PATH)
        assert os.path.exists(data_file_resnet)
    
        cmd = ["bash", os.path.join(profiler_utils, "resnet_batch.sh")]
    
>       proc = subprocess.run(cmd, stdout=subprocess.PIPE, check=True, timeout=1000)

test_pytorch_profiler.py:147: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

input = None, capture_output = False, timeout = 1000, check = True
popenargs = (['bash', '/home/ubuntu/serve/test/pytest/../../test/pytest/profiler_utils/resnet_batch.sh'],), kwargs = {'stdout': -1}
process = <subprocess.Popen object at 0x7f84cf08f8e0>, stdout = b'', stderr = None, retcode = 26

    def run(*popenargs,
            input=None, capture_output=False, timeout=None, check=False, **kwargs):
        """Run command with arguments and return a CompletedProcess instance.
    
        The returned instance will have attributes args, returncode, stdout and
        stderr. By default, stdout and stderr are not captured, and those attributes
        will be None. Pass stdout=PIPE and/or stderr=PIPE in order to capture them.
    
        If check is True and the exit code was non-zero, it raises a
        CalledProcessError. The CalledProcessError object will have the return code
        in the returncode attribute, and output & stderr attributes if those streams
        were captured.
    
        If timeout is given, and the process takes too long, a TimeoutExpired
        exception will be raised.
    
        There is an optional argument "input", allowing you to
        pass bytes or a string to the subprocess's stdin.  If you use this argument
        you may not also use the Popen constructor's "stdin" argument, as
        it will be used internally.
    
        By default, all communication is in bytes, and therefore any "input" should
        be bytes, and the stdout and stderr will be bytes. If in text mode, any
        "input" should be a string, and stdout and stderr will be strings decoded
        according to locale encoding, or by "encoding" if set. Text mode is
        triggered by setting any of text, encoding, errors or universal_newlines.
    
        The other arguments are the same as for the Popen constructor.
        """
        if input is not None:
            if kwargs.get('stdin') is not None:
                raise ValueError('stdin and input arguments may not both be used.')
            kwargs['stdin'] = PIPE
    
        if capture_output:
            if kwargs.get('stdout') is not None or kwargs.get('stderr') is not None:
                raise ValueError('stdout and stderr arguments may not be used '
                                 'with capture_output.')
            kwargs['stdout'] = PIPE
            kwargs['stderr'] = PIPE
    
        with Popen(*popenargs, **kwargs) as process:
            try:
                stdout, stderr = process.communicate(input, timeout=timeout)
            except TimeoutExpired as exc:
                process.kill()
                if _mswindows:
                    # Windows accumulates the output in a single blocking
                    # read() call run on child threads, with the timeout
                    # being done in a join() on those threads.  communicate()
                    # _after_ kill() is required to collect that and add it
                    # to the exception.
                    exc.stdout, exc.stderr = process.communicate()
                else:
                    # POSIX _communicate already populated the output so
                    # far into the TimeoutExpired exception.
                    process.wait()
                raise
            except:  # Including KeyboardInterrupt, communicate handled that.
                process.kill()
                # We don't call process.wait() as .__exit__ does that for us.
                raise
            retcode = process.poll()
            if check and retcode:
>               raise CalledProcessError(retcode, process.args,
                                         output=stdout, stderr=stderr)
E               subprocess.CalledProcessError: Command '['bash', '/home/ubuntu/serve/test/pytest/../../test/pytest/profiler_utils/resnet_batch.sh']' returned non-zero exit status 26.

/home/ubuntu/anaconda3/lib/python3.8/subprocess.py:512: CalledProcessError
```

Objective of the test case is to run batched inputs and verify if only one profiler trace file is generated. https://github.com/pytorch/serve/blob/b11ba0fa97ceea7003bfa7c4fd4203f36fee3d34/test/pytest/test_pytorch_profiler.py#L145 - uses shell script to invoke two parallel request using curl.

Instead, simplifying the logic to invoke parallel requests using python threads.

Fixed PyTorch profiler test - `test_pytorch_profiler.py` - `test_batch_input`

Fixes #(issue)

## Type of change

- [ ] Test A
- [ ] Test B

- UT/IT execution results

- Logs
[result.txt](https://github.com/pytorch/serve/files/7856049/result.txt)

## Checklist:


test/pytest/test_pytorch_profiler.py
```diff
@@ -137,17 +138,22 @@ def test_batch_input(set_custom_handler, handler_name):
...
+        response = requests.post("{}/predictions/resnet152".format(TF_INFERENCE_API), data)
+        print(response.text)
 
-    proc = subprocess.run(cmd, stdout=subprocess.PIPE, check=True, timeout=1000)
 
-    assert "tiger_cat" in proc.stdout.decode("utf-8")
-    assert "Labrador_retriever" in proc.stdout.decode("utf-8")
+    with futures.ThreadPoolExecutor(2) as executor:
+        for _ in range(2):
+            executor.submit(invoke_batch_input)
```
lxning, line 156: 1. The asserts were deleted. Where is the verification of the results? 
2. I don't fully understand why threadpooling can fix regression test? Could you pls elaborate the original issue?

shrinath-suresh: The objective of the test case (test_batch_input) : When user provides inputs as a batch, only 1 PyTorch trace file should be generated. 

Initially i created the test case with a shell script - the shell script invokes 2 curl commands. The regression test is failing at this line - while forming the path for the shell script.
https://github.com/pytorch/serve/blob/b11ba0fa97ceea7003bfa7c4fd4203f36fee3d34/test/pytest/test_pytorch_profiler.py#L145 .

Instead of fixing the path construction, i thought it would be more simpler to use python threads. [This line](https://github.com/pytorch/serve/blob/b11ba0fa97ceea7003bfa7c4fd4203f36fee3d34/test/pytest/test_pytorch_profiler.py#L152) verifies the test case.

test/pytest/test_pytorch_profiler.py
```diff
@@ -137,17 +138,22 @@ def test_batch_input(set_custom_handler, handler_name):
...
+
     if os.path.exists(CUSTOM_PATH):
         shutil.rmtree(CUSTOM_PATH)
     assert os.path.exists(data_file_resnet)
 
-    cmd = ["bash", os.path.join(profiler_utils, "resnet_batch.sh")]
+    def invoke_batch_input():
+        data = open(data_file_resnet, "rb")
+        response = requests.post("{}/predictions/resnet152".format(TF_INFERENCE_API), data)
+        print(response.text)
```
lxning, line 151: Could you add response validation instead of printing?

shrinath-suresh: Done.

lxning (approved):

HamidShojanazeri (approved):
Tests are passing, [logs](https://github.com/pytorch/serve/pull/1397)

PR #1415, opened 2022-02-07
black linting

msaroufim:
Included a new script
* `ts_scripts/lint.sh` that takes care of only linting new files with `black` @nskool is there any way we can integrate this in CI? Feel free to push directly to my branch
* Added `black, flake8` in dev requirements
* Ran `black .` and added initial linting
* TODO: `flake8` errors will be addressed in a separate PR


nskool:
In the CI/PR builds - we can have the linter run and recommend contributors to use it. For example, if the lint.sh script can return a non-zero code if the code is not linted, then the PR test can be marked as 'failed'. Contributors can then be requested to run 'lint.sh', and update the PR.

ts_scripts/lint.sh
```diff
@@ -0,0 +1,79 @@
...
+if [ "$CHANGED_FILES" != "" ]
+then
+    # Processing files one by one since passing directly $CHANGED_FILES will
+    # treat the whole variable as a single file.
+    echo "Running linters ..."
+    for file in $CHANGED_FILES
+    do
+        echo "Checking $file"
+        usort format "$file"
+        black "$file" -q
```
nskool, line 55: black, by default, lints at column 80. It might be helpful to increase this to 100, or preferably 120 by supplying `-l 120`, which is more in line with recent times.

msaroufim: Good call, fixed

PR #1424, opened 2022-02-09
Refactoring `README.md`

msaroufim:
Refactoring `README.md` to be shorter, better describe the torchserve value proposition and break up apart interesting information into linked but separate docs with a focus on internals, optimization and profiling and the handler UX


README.md
```diff
@@ -1,273 +1,88 @@
...
-
-To test the model server, send a request to the server's `predictions` API. TorchServe supports all [inference](docs/inference_api.md) and [management](docs/management_api.md) api's through both [gRPC](docs/grpc_api.md) and [HTTP/REST](docs/rest_api.md).
-
-#### Using GRPC APIs through python client
-
- - Install grpc python dependencies :
- 
 ```bash
-pip install -U grpcio protobuf grpcio-tools
+curl http://127.0.0.1:8080/predictions/bert -T input.txt
```
HamidShojanazeri, line 6: @msaroufim are we keeping this here as for showing the ease of use?

README.md
```diff
@@ -1,273 +1,88 @@
...
-
-## Quick Start with Docker
 Refer to [torchserve docker](docker/README.md) for details.
 
-## Highlighted Examples
-* [HuggingFace Transformers](examples/Huggingface_Transformers)
+## ⚡ Why TorchServe
+* [Model Management API](docs/management_api.md): multi model management with optimized worker to model allocation
+* [Inference API](docs/inference_api.md): REST and gRPC support for batched inference
+* [TorchServe Workflows](examples/Workflows/README.md): deploy complex DAGs with multiple interdependent models
```
HamidShojanazeri, line 30: @msaroufim how about adding the ensemble(term) somewhere in the doc to make sure it relates to user terminology.

HamidShojanazeri (approved):
Thanks @msaroufim, directionally LGTM.

README.md
msaroufim: Yeah do you think it does a good job at showcasing ease of use? I also thought we could showcase the requests library but I figured everyone uses curl

README.md
msaroufim: I'm not sure how I feel about this because ensemble to me means aggregating the predictions of `n` models into a single one whereas a workflow is about setting up a DAG which is more general and what we support

HamidShojanazeri: @msaroufim right, as it supports parallel workflow so it may fit into that defintion as well?

PR #1426, opened 2022-02-14
Github Action Nightly Build

msaroufim:
Using this PR as a draft for Github Action nightly builds - does this work? I think the only missing thing is making sure if `ts_scripts/install_dependencies.py` is needed for the build @nskool 

Otherwise I can create a new project called `torchserve-nightly` on pypi and after I upload my personal access token we should be good to go? @seemethere 

repurposed the workflow here from @d4l3k  https://github.com/pytorch/torchx/blob/main/.github/workflows/nightly.yaml

TODO - can potentially do this in another PR
* Enable `DRY_RUN` by default
* Add last git hash to PKG_INFO
* Use github to publish artifacts

# Local test
1. Run commands from github actions script one by one - the last one should print
```
(serve) ubuntu@ip-172-31-25-104:~/serve$ python setup.py --override-name torchserve-nightly bdist_wheel
-- torchserve-nightly building version: 2022.2.14
```
2. If you have access to the `PYPI_TOKEN` you can try uploading a release otherwise you can go to the egg info folder and verify the information in `PKG_INFO` which will look like

```
Metadata-Version: 2.1
Name: torchserve-nightly
Version: 2022.2.14
```

And if you go `dist/` then you'll see a file named `torchserve_nightly-2022.2.14-py3-none-any.whl`

Output available here: https://pypi.org/project/torchserve-nightly/2022.2.14/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -142,12 +142,15 @@ class BuildPlugins(Command):
 
 
 if __name__ == '__main__':
-    from argparse import ArgumentParser
-    parser = ArgumentParser()
-    parser.add_argument("--override-name", default='torchserve')
-    args = parser.parse_args()
-    name = args.override_name
     # Get nightly version if nightly in name
+    name = 'torchserve'
+
+    # Clever code to figure out if setup.py was trigger by ts_scripts/push_nightly.sh
+    NAME_ARG = "--override-name"
+    if NAME_ARG in sys.argv:
+        idx = sys.argv.index(NAME_ARG)
+        name = sys.argv.pop(idx + 1)
+        sys.argv.pop(idx)
     is_nightly = "nightly" in name
 
     version = get_nightly_version() if is_nightly else detect_model_server_version()
--- a/setup.py
+++ b/setup.py
@@ -51,6 +51,9 @@ def pypi_description():
     with open('PyPiDescription.rst') as df:
         return df.read()
 
+def get_nightly_version():
+    today = date.today()
+    return f"{today.year}.{today.month}.{today.day}"
 
 def detect_model_server_version():
     sys.path.append(os.path.abspath("ts"))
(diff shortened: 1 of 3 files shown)


nskool:
Binary is generated after running install_dependencies.py, that's how the nightly binaries are built. https://github.com/pytorch/serve/blob/master/ci/torchserve_nightly_build.yml#L12.

setup.py
```diff
@@ -51,6 +51,9 @@ def pypi_description():
     with open('PyPiDescription.rst') as df:
         return df.read()
 
+def get_nightly_version():
+    today = date.today()
+    return f"{today.year}.{today.month}.{today.day}"
```
malfet, line 56: nit (see https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior )
```suggestion
    return today.strftime("%Y.%m.%d")
```

setup.py
```diff
@@ -139,12 +142,25 @@ def run(self):
...
-    version = detect_model_server_version()
+    # Get nightly version if nightly in name
+    name = 'torchserve'
+
+    # Clever code to figure out if setup.py was trigger by ts_scripts/push_nightly.sh
+    NAME_ARG = "--override-name"
+    if NAME_ARG in sys.argv:
+        idx = sys.argv.index(NAME_ARG)
+        name = sys.argv.pop(idx + 1)
+        sys.argv.pop(idx)
```
malfet, line 153: Please use [ArgumentParser](https://docs.python.org/3/library/argparse.html)
```suggestion
    from argparse import ArgumentParser
    parser = ArgumentParser()
    parser.add_argument("--override-name", default='torchserve')
    args = parser.parse_args()
    name = args.override_name
    # Get nightly version if nightly in name
```

malfet (approved):

setup.py
nskool: +1

ts_scripts/push_nightly.sh
```diff
@@ -0,0 +1,16 @@
...
+python setup.py --override-name torchserve-nightly bdist_wheel
+
+if [ -z "$PYPI_TOKEN" ]; then
+    echo "must specify PYPI_TOKEN"
+    exit 1
+fi
+
+python3 -m twine upload \
+    --username __token__ \
+    --password "$PYPI_TOKEN" \
```
nskool, line 15: Can you ensure that the "$PYPI_TOKEN" is not printed in the logs?

msaroufim: Just inspect visually and it's not

nskool (approved):

d4l3k (approved):
LGTM! Glad to see our setup is helping other projects out :)

maaquib (approved):

ts_scripts/push_nightly.sh
malfet: GitHub (and CircleCI for that matter) logging facilities ensure that secrets are always filtered out before being added to the log

PR #1428, opened 2022-02-14
delete circleci

msaroufim:
Our current CI story is
* `.circleci` old, noone uses this so deleting in this PR
* `.github` used for docs and soon nightly builds and over time more things
* `binaries` scripts to build code
* `ci` AWS CI which also has a codebuild script which runs scripts from `binaries` when we need to make releases https://github.com/pytorch/serve/blob/master/ci/torchserve_nightly_build.yml


malfet (approved):
Nice!

nskool (approved):

PR #1436, opened 2022-02-16
Update permissions on nightly-build.yml

msaroufim:
Nightly build build fails right now when apt-get update is run because of a permission denied issue

So either we remove these two lines or add sudo - what do you think is best? @nskool 

Also I wonder how necessary this flag is for `install_dependencies.py` `--cuda=cu102` not sure how github actions plays with GPU runners

https://github.com/pytorch/serve/actions/runs/1852465948

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -20,6 +20,8 @@ jobs:
         run: |
           set -eux
           pip install twine
+          sudo apt-get update
+          sudo apt-get install sudo -y
           python ts_scripts/install_dependencies.py --cuda=cu102 --environment=dev
           pip install -e .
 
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-build.yml
```diff
@@ -20,8 +20,8 @@ jobs:
         run: |
           set -eux
           pip install twine
-          apt-get update
-          apt-get install sudo -y
+          sudo apt-get update
+          sudo apt-get install sudo -y
```
nskool, line 24: You wouldn't need the line 24, sudo is already shipped with the github-actions ubuntu18.04 image.

Just running `sudo apt-get update` should suffice.

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#administrative-privileges-of-github-hosted-runners

nskool (approved):

seemethere (approved):

PR #1460, opened 2022-02-24
decorator based serving

msaroufim:
New UX to serve models. Looking for feedback.

TODO
* Fix a bug with model packaging and make sure mar file is generated
* Run a curl request against the deployed endpoint as an end to end integration test


ts/utils/serve_decorator.py
```diff
@@ -0,0 +1,215 @@
...
+    Create a new handler that inherits from the base handler
+    Update the inference() function with the function captured by the decorator
+    Write the handler to disk
+
+    TODO: Improve story for adding custom dependencies, all the hidden imports or library code
+    TODO: Improve workflow story, coming in future PR
+
+    """
+    
+    # If user provides a function then override inference() function in base handler
```
HamidShojanazeri, line 74: @msaroufim is this general to any function? including inference, initialize, pre and post process?

HamidShojanazeri: what is the reason that we only reference the inference func here?

ts/tests/unit_tests/test_decorator.py
```diff
@@ -0,0 +1,37 @@
...
+    create_torchserve_config()
+    assert os.path.exists("config.properties")
+
+def test_start():
+    error = torchserve_start(create_handler(ToyHandler))
+    assert error == 0
+    error = torchserve_stop()
+    assert error == 0
+
+def test_integration():
```
HamidShojanazeri, line 31: @msaroufim it will be helpful to have a test/ example that showcase the decorator part.

ts/utils/serve_decorator.py
```diff
@@ -0,0 +1,215 @@
...
+    Use this instead on top of another handler
+    """
+    @wraps(cls)
+    def archive_model(*args, **kwargs):
+        class_source = inspect.getsource(cls)
+        with open(filename, "w") as f:
+            f.write(class_source)
+
+
+def torchserve_start(handler : str, model_store : str ="model_store", ts_config : str ="config.properties", model_mar : str ="model.mar"):
```
HamidShojanazeri, line 102: I wonder if using existing utils in [ts_script](https://github.com/pytorch/serve/blob/master/ts_scripts/tsutils.py#L27) would be helpful/ avoid redundancy? or alternatively expand it to cover workflows( in case we re deprecating yet).

ts/tests/unit_tests/test_decorator.py
msaroufim: The test for the decorator is `serve(ToyHandler)`. A decorator is nothing but a function that takes in as argument the function or class below it

ts/utils/serve_decorator.py
msaroufim: Sure lemme me how to refactor

For workflows specifically this is not coming soon so will not address in this PR

ts/utils/serve_decorator.py
msaroufim: It's purely for simplicity, technically you only need the `inference()` function because you can do pre and post processing as well in the same function

ts/utils/serve_decorator.py
msaroufim: `ts_script` is not part of the torchserve library, it only includes what's in the `ts` code. I can refactor this in a future PR

PR #1473, opened 2022-02-25
added Text Generation example

Tegzes:
## Description

Fixes #(1386)

## Type of change

- [ ] Test A
- [ ] Test B

- Logs

## Checklist:


pull_request_template.md
```diff
@@ -9,9 +9,9 @@ Fixes #(issue)
 Please delete options that are not relevant.
 
 - [ ] Bug fix (non-breaking change which fixes an issue)
-- [ ] New feature (non-breaking change which adds functionality)
+- [x] New feature (non-breaking change which adds functionality)
```
msaroufim, line 12: Please revert the changes here, the template is automatically used when you open a new PR - no need to update it here

examples/Huggingface_Transformers/Transformer_handler_generalized.py
```diff
@@ -126,7 +129,7 @@ def preprocess(self, requests):
             max_length = self.setup_config["max_length"]
             logger.info("Received text: '%s'", input_text)
             # preprocessing text for sequence_classification and token_classification.
-            if self.setup_config["mode"] == "sequence_classification" or self.setup_config["mode"] == "token_classification":
+            if self.setup_config["mode"] == "sequence_classification" or self.setup_config["mode"] == "token_classification" or self.setup_config["mode"] == "text_generation":
```
msaroufim, line 132: nit: let's break this into multiple lines

Tegzes: maybe something like "if self.setup_config["mode"] in {"sequence_classification", "token_classification", "text_generation"}:" would be more elegant?

pull_request_template.md
Tegzes: Will do!

examples/Huggingface_Transformers/Transformer_handler_generalized.py
```diff
@@ -219,6 +222,16 @@ def inference(self, input_batch):
...
+            
+        # Handling inference for text_generation.
+        if self.setup_config["mode"] == "text_generation":
+            
+            for input in input_ids_batch:
+                prompt_length = len(self.tokenizer.decode(input[0]))
+                outputs = self.model.generate(input, max_length=150, do_sample=True, top_p=0.95, top_k=60)
+                generated = self.tokenizer.decode(input) + self.tokenizer.decode(outputs[0])[prompt_length + 1 :]
+            
+                inferences.append(generated)
```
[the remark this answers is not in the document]
Tegzes: I tried to keep the same approach/style as the other transformer examples, so for each text example (prompt) from the batch, there is one instance of generated text (which is put together with the prompt, as in the official example). Max sequence is per every text example. The output will have the length of the max sequence and will be decoded in text. The parameters chosen in this function are the ones used in the official example from Huggingface, because I figured that maybe a lot of people would like see the same example served. Of course this might be improved in the future, by adding more flexibility in parameters selection.

The model size is comparable of the other transformers (~500 mb), but I'm not sure about the speed.

The output is not deterministic, because we can't know in advance what the generated text will be. From the local tests that I've done, each time I got a different output. I think the user can be sure the example is executed correctly when some text is generated, regardless of the content.

Considering this, I'm not sure how I should make assertion tests with this example.

msaroufim (approved):
Excellent, thank you for your contribution!

msaroufim:
Flagging @lxning @maaquib to merge after 0.5.3 release

PR #1493, opened 2022-03-08
Fix checking lbl_classes

### LINKED ISSUE #1020: Possible bug in ts.utils.util.map_class_to_label
This doesn't seem quite right:
https://github.com/pytorch/serve/blob/f0da8d2714628cb5de0c6eaa3ac0448a6b226640/ts/utils/util.py#L63-L64

It's checking the same thing twice and talking about the classes being wrong. Looks like it should be checking `lbl_classes` and `prob` are both lists (or None in the case of `lbl_classes`).

avinashsai:
Fixes #1020


msaroufim (approved):

avinashsai:
@msaroufim Not sure why the checks are failing.

maaquib:
Failing test

```
## Register marfile:https://torchserve.s3.amazonaws.com/mar_files/my_text_classifier_v4.mar

Model my_text_classifier_v4 registered successfully
  File "ts_scripts/torchserve_grpc_client.py", line 82, in <module>
    infer(get_inference_stub(), args[1], args[2])
  File "ts_scripts/torchserve_grpc_client.py", line 26, in infer
    response = stub.Predictions(
  File "C:\Python38\lib\site-packages\grpc\_channel.py", line 946, in __call__
    return _end_unary_response_blocking(state, call, False, None)
  File "C:\Python38\lib\site-packages\grpc\_channel.py", line 849, in _end_unary_response_blocking
    raise _InactiveRpcError(state)
grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
	status = StatusCode.UNAVAILABLE
	details = "Prediction failed
org.pytorch.serve.http.InternalServerException"
	debug_error_string = "{"created":"@1646851005.293000000","description":"Error received from peer ipv6:[::1]:7070","file":"src/core/lib/surface/call.cc","file_line":904,"grpc_message":"Prediction failed\norg.pytorch.serve.http.InternalServerException","grpc_status":14}"
```

ts/utils/util.py
```diff
@@ -60,7 +60,7 @@ def map_class_to_label(probs, mapping=None, lbl_classes=None):
     Given a list of classes & probabilities, return a dictionary of
     { friendly class name -> probability }
     """
-    if not (isinstance(probs, list) and isinstance(probs, list)):
+    if not (isinstance(probs, list) and isinstance(lbl_classes, list)):
```
msaroufim, line 63: I don't think this is correct behavior, 

`lbl_classes` could be `None` and that's OK so the fix should be instead

```suggestion
    if not (isinstance(probs, list):
```

PR #1474, opened 2022-02-26
added nvgpu to common dependencies

msaroufim:
Fixes

for EC2 ubuntu base gpu torchserve fails to start due to missing nvgpu package
https://github.com/pytorch/serve/blob/ff4d90bfe18556739bee8465b83acb99f99be4a7/ts/metrics/system_metrics.py#L61

PR #1453, opened 2022-02-20
add gpu and memory usage in metrics

lxning:
## Description

Please include a summary of the feature or issue being fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes #(issue)
#1452 

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Feature/Issue validation/testing

Please describe the tests [UT/IT] that you ran to verify your changes and relevant result summary. Provide instructions so it can be reproduced. 
Please also list any relevant details for your test configuration.

- [ ] Test A
- [ ] Test B

- Logs
model_metrics.log:
2022-02-20T04:51:12,858 [INFO ] pool-3-thread-1 TS_METRICS - GPUUtilization.Percent:0|#Level:Host,device_id:0|#hostname:ip-172-31-37-20,timestamp:1645332672
2022-02-20T04:51:12,858 [INFO ] pool-3-thread-1 TS_METRICS - GPUMemoryUtilization.Percent:0|#Level:Host,device_id:0|#hostname:ip-172-31-37-20,timestamp:1645332672
2022-02-20T04:51:12,858 [INFO ] pool-3-thread-1 TS_METRICS - GPUMemoryUsed.Megabytes:0|#Level:Host,device_id:0|#hostname:ip-172-31-37-20,timestamp:1645332672

cat ab_report.csv:
Benchmark,Batch size,Batch delay,Workers,Model,Concurrency,Requests,TS failed requests,TS throughput,TS latency P50,TS latency P90,TS latency P99,TS latency mean,TS error rate,Model_p50,Model_p90,Model_p99,predict_mean,handler_time_mean,waiting_time_mean,worker_thread_mean,memory_percentage_mean,gpu_percentage_mean,gpu_memory_percentage_mean,gpu_memory_used_mean
AB,1,100,4,https://torchserve.pytorch.org/mar_files/vgg16.mar,100,10000,0,279.86,354,368,414,357.327,0.0,13.31,13.9,14.21,13.57,13.53,341.0,0.34,5.54,11.25,5.88,920.0

## Checklist:

- [ ] Have you added tests that prove your fix is effective or that this feature works?
- [ ] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the documentation?

PR #1509, opened 2022-03-15
Add common/requirements.txt to GPU docker image

aaronzs:
## Description

`nvgpu` is missing from `torchserve` GPU container, and it causes import error when `ts.metrics.system_metrics.gpu_utilization()` is called.

It may be related to PR #1453 and #1474.


msaroufim:
I think we should consider leveraging requirements.txt in Dockerfile so this doesnt happen again https://github.com/pytorch/serve/blob/master/docker/Dockerfile

msaroufim (approved):

msaroufim:
Just extended this to install the existing `requirements/common.txt` - the dev dependencies already take care of this

HamidShojanazeri (approved):

PR #1530, opened 2022-03-24
Bugfix/Support Workflow with Kserve V2 Envelope

### LINKED ISSUE #1529: Workflow API doesn't work with Kserve V2 envelope
## Context

* torchserve version: `0.5.3`
* torch-model-archiver version:
* torch version: `1.10.0`
* torchvision version [if any]:
* torchtext version [if any]:
* torchaudio version [if any]:
* java version:
* Operating System and version: Ubuntu 18.04

## Your Environment

I use the image `pytorch/torchserve:0.5.3-cpu` as the environment.

* Installed using source? [yes/no]: No
* Are you planning to deploy it using docker container? [yes/no]: yes
* Is it a CPU or GPU environment?: CPU
* Using a default/custom handler? [If possible upload/share custom handler/model]: custom, see below.
* What kind of model is it e.g. vision, text, audio?: vision

**I test with two workflows:**

- dog_breed_classification in the examples
- default mnist model, and make a testing workflow with the following files:

**mnist_workflow_handlers.py**

```python
import json
import base64

def print_data(data, context):
    print("+" * 50)
    print("node: print_data")
    print("data:", data)
    print("context:", context)
    if data is None:
        return data
    output = []
    for row in data:
        d = row.get("data") or row.get("body")
        # if isinstance(d, (bytes, bytearray)):
        #     d = d.decode('utf-8')
        output.append(d)
    return output

def test_data(data, context):
    print("+" * 50)
    print("node: test_data")
    print("data:", data)
    print("context:", context)
    if data is None:
        return data
    output = []
    for row in data:
        d = row.get("data") or row.get("body")
        # if isinstance(d, (bytes, bytearray)):
        #     d = d.decode('utf-8')
        output.append(d)
    return output

def combine_data(data, context):
    if data is None:
        return data
    print("+" * 50)
    print("node: combine_data")
    print("data:", data)
    print("context:", context)
    output = []
    for row in data:
        data_dict = {}
        for k in row:
            d = row.get(k)
            # if isinstance(d, (bytes, bytearray)):
            #     d = d.decode('utf-8')
            data_dict[k] = d
        output.append(data_dict)
    print([d.get("print_data") for d in output])
    return [d.get("print_data") for d in output]
```

**mnist_workflow.yaml**

```yaml
models:
    min-workers: 1
    max-workers: 1
    batch-size: 4
    max-batch-delay: 100
    retry-attempts: 5
    timeout-ms: 300000

    mnist:
      url: mnist.mar

dag:
  print_data: [test_data, combine_data]
  test_data: [combine_data]
  combine_data: [mnist]
```

* Are you planning to use local models from model-store or public url being used e.g. from S3 bucket etc.?
  [If public url then provide link.]: 
* Provide config.properties, logs [ts.log] and parameters used for model registration/update APIs:
```
inference_address=http://0.0.0.0:8000
management_address=http://0.0.0.0:8001
metrics_address=http://0.0.0.0:8002
number_of_netty_threads=32
job_queue_size=1000
install_py_dep_per_model=true
service_envelope=kservev2
load_models=all
```
* Link to your project [if any]:

## Expected Behavior
The workflow api should work out of box with kserve v2 envelope.

## Current Behavior
The workflow api doesn't work with kserve v2 envelope.

### 1.The envelope of the workflow function is not set correctly.

> https://github.com/pytorch/serve/blob/6a191f822f2e759c2aad07b8f9eacbedcb039827/frontend/server/src/main/java/org/pytorch/serve/wlm/ModelManager.java#L122

While the models' envelope are set to the default envelope specified in config.properties, the workflow funcitons' envelope is null.

### 2. The request envelope `kservev2.py` cannot internal workflow requests and nested requests properly.

> https://github.com/pytorch/serve/blob/master/ts/torch_handler/request_envelope/kservev2.py

If a node has multiple parent nodes, for example node `A` with result `["a"]` and node `B` with result `["b"]`, the output of these node will be combined by the `frontend` into:

```json
[
    {
        "A": {"outputs": ["data": ["a"]]},
        "B": {"outputs": ["data": ["b"]]},
    }
]
```

However, because there is no key `data` or `body` (only `A` and `B`)in this json, the envelope class cannot process the data.

Even if the data is found, the kserve v2 envelope can only handle inputs key of `inputs` other than `outputs`.

Also in order to pass it to the model properly as it does with null envelope, the request data needs to be converted into the form:

```json
[
    "data": {
        "A": ["a"],
        "B": ["b"]
    }
]
```

## Possible Solution

It is possible to set all the workflow model/functions' envelope to null, whatever the service_envelope specifed in the config.properties. However, this creates an inconsistence bewteen the model api standard and the workflow api standard. And all the models in the workflow use null envelope while other models use another envelope. Therefore, it is not a desired solution.

In order to solve the issue and maintain the consistency bewteen APIs of workflow and standalone models:

1. Set the envelope of workflow functions to the default envelope as regular models (those have `.mar` files)
2. Identify `internal workflow requests` between nodes, and let kserve v2 envelope be able to parse `outputs` data. For example, pass a special header from frontend, telling the backend this is an `internal workflow request`.
3. Identify internal workflow requests type, whether it is a `flat` or `nested` data. For example, pass a special header from frontend, telling backend the request is `flat` or `nested`.
4. Kserve v2 envelope handler will handle the data according to the header.

**I've prepared a pull request and will submit it asap for you review**

I divide the changes into seperate commits targeting different problems with description in the commit message. Once all of the commits fulfill the requirements, I will add more unittests, rebase and squash the commits.

If possible, please help to give suggestions whether I'm on the right track of this problem.

## Steps to Reproduce
1. Pull the image `pytorch/torchserve:0.5.3-cpu`
2. Don't set the envelope in `config.properties`. Run the container with the workflow mnist I provided above, or the example workflow dog_breed_classification. The prediction will succeed.
3. Set the `service_envelope=kservev2`, run the container with the workflow in `2.`. Now the prediction will fail

## Failure Logs [if any]

I add some print in `ts/model_service_worker.py`:

```python
class TorchModelServiceWorker(object):

    @staticmethod
    def load_model(load_model_request):
        try:
            model_dir = load_model_request["modelPath"].decode("utf-8")
            model_name = load_model_request["modelName"].decode("utf-8")
            handler = load_model_request["handler"].decode("utf-8") if load_model_request["handler"] else None
            envelope = load_model_request["envelope"].decode("utf-8") if "envelope" in load_model_request else None
            envelope = envelope if envelope is not None and len(envelope) > 0 else None
            print("=" * 50)
            print("model name:", model_name)
            print("envelope: ", envelope)
```

Below are the logs when loading the `dog_breed_classification`:

```diff 
+ 2022-03-22T06:39:04,375 [INFO ] W-9024-dog_breed_wf_new__pre_processing_1.0-stdout MODEL_LOG - ==================================================
+ 2022-03-22T06:39:04,375 [INFO ] W-9024-dog_breed_wf_new__pre_processing_1.0-stdout MODEL_LOG - model_name: dog_breed_wf_new__pre_processing
+ 2022-03-22T06:39:04,375 [INFO ] W-9024-dog_breed_wf_new__pre_processing_1.0-stdout MODEL_LOG - envelope: None
2022-03-22T06:39:04,375 [INFO ] W-9024-dog_breed_wf_new__pre_processing_1.0-stdout MODEL_LOG - model_name: dog_breed_wf_new__pre_processing, batchSize: 1
2022-03-22T06:39:04,376 [INFO ] W-9024-dog_breed_wf_new__pre_processing_1.0 org.pytorch.serve.wlm.WorkerThread - Backend response time: 1
2022-03-22T06:39:04,376 [DEBUG] W-9024-dog_breed_wf_new__pre_processing_1.0 org.pytorch.serve.wlm.WorkerThread - W-9024-dog_breed_wf_new__pre_processing_1.0 State change WORKER_STARTED -> WORKER_MODEL_LOADED
2022-03-22T06:39:04,376 [INFO ] W-9024-dog_breed_wf_new__pre_processing_1.0 TS_METRICS - W-9024-dog_breed_wf_new__pre_processing_1.0.ms:959|#Level:Host|#hostname:8b026932f222,timestamp:1647931144
2022-03-22T06:39:04,376 [INFO ] W-9024-dog_breed_wf_new__pre_processing_1.0 TS_METRICS - WorkerThreadTime.ms:1|#Level:Host|#hostname:8b026932f222,timestamp:null
2022-03-22T06:39:04,875 [DEBUG] pool-6-thread-2 org.pytorch.serve.wlm.ModelVersionedRefs - Adding new version 1.0 for model dog_breed_wf_new__dog_breed_classification
2022-03-22T06:39:04,875 [DEBUG] pool-6-thread-2 org.pytorch.serve.wlm.ModelVersionedRefs - Setting default version to 1.0 for model dog_breed_wf_new__dog_breed_classification
2022-03-22T06:39:04,875 [INFO ] pool-6-thread-2 org.pytorch.serve.wlm.ModelManager - Model dog_breed_wf_new__dog_breed_classification loaded.
2022-03-22T06:39:04,875 [DEBUG] pool-6-thread-2 org.pytorch.serve.wlm.ModelManager - updateModel: dog_breed_wf_new__dog_breed_classification, count: 1
2022-03-22T06:39:04,876 [DEBUG] W-9026-dog_breed_wf_new__dog_breed_classification_1.0 org.pytorch.serve.wlm.WorkerLifeCycle - Worker cmdline: [/home/venv/bin/python, /home/venv/lib/python3.8/site-packages/ts/model_service_worker.py, --sock-type, unix, --sock-name, /home/model-server/tmp/.ts.sock.9026]
2022-03-22T06:39:05,040 [INFO ] W-9025-dog_breed_wf_new__cat_dog_classification_1.0-stdout MODEL_LOG - Listening on port: /home/model-server/tmp/.ts.sock.9025
2022-03-22T06:39:05,041 [INFO ] W-9025-dog_breed_wf_new__cat_dog_classification_1.0-stdout MODEL_LOG - [PID]14321
2022-03-22T06:39:05,041 [INFO ] W-9025-dog_breed_wf_new__cat_dog_classification_1.0-stdout MODEL_LOG - Torch worker started.
2022-03-22T06:39:05,041 [INFO ] W-9025-dog_breed_wf_new__cat_dog_classification_1.0-stdout MODEL_LOG - Python runtime: 3.8.0
2022-03-22T06:39:05,041 [DEBUG] W-9025-dog_breed_wf_new__cat_dog_classification_1.0 org.pytorch.serve.wlm.WorkerThread - W-9025-dog_breed_wf_new__cat_dog_classification_1.0 State change null -> WORKER_STARTED
2022-03-22T06:39:05,041 [INFO ] W-9025-dog_breed_wf_new__cat_dog_classification_1.0 org.pytorch.serve.wlm.WorkerThread - Connecting to: /home/model-server/tmp/.ts.sock.9025
2022-03-22T06:39:05,042 [INFO ] W-9025-dog_breed_wf_new__cat_dog_classification_1.0-stdout MODEL_LOG - Connection accepted: /home/model-server/tmp/.ts.sock.9025.
2022-03-22T06:39:05,042 [INFO ] W-9025-dog_breed_wf_new__cat_dog_classification_1.0 org.pytorch.serve.wlm.WorkerThread - Flushing req. to backend at: 1647931145042
+ 2022-03-22T06:39:05,042 [INFO ] W-9025-dog_breed_wf_new__cat_dog_classification_1.0-stdout MODEL_LOG - ==================================================
+ 2022-03-22T06:39:05,043 [INFO ] W-9025-dog_breed_wf_new__cat_dog_classification_1.0-stdout MODEL_LOG - model_name: dog_breed_wf_new__cat_dog_classification
+ 2022-03-22T06:39:05,043 [INFO ] W-9025-dog_breed_wf_new__cat_dog_classification_1.0-stdout MODEL_LOG - envelope: kservev2
2022-03-22T06:39:05,043 [INFO ] W-9025-dog_breed_wf_new__cat_dog_classification_1.0-stdout MODEL_LOG - model_name: dog_breed_wf_new__cat_dog_classification, batchSize: 4
2022-03-22T06:39:05,781 [INFO ] W-9026-dog_breed_wf_new__dog_breed_classification_1.0-stdout MODEL_LOG - Listening on port: /home/model-server/tmp/.ts.sock.9026
2022-03-22T06:39:05,782 [INFO ] W-9026-dog_breed_wf_new__dog_breed_classification_1.0-stdout MODEL_LOG - [PID]14334
2022-03-22T06:39:05,782 [INFO ] W-9026-dog_breed_wf_new__dog_breed_classification_1.0-stdout MODEL_LOG - Torch worker started.
2022-03-22T06:39:05,782 [INFO ] W-9026-dog_breed_wf_new__dog_breed_classification_1.0-stdout MODEL_LOG - Python runtime: 3.8.0
2022-03-22T06:39:05,782 [DEBUG] W-9026-dog_breed_wf_new__dog_breed_classification_1.0 org.pytorch.serve.wlm.WorkerThread - W-9026-dog_breed_wf_new__dog_breed_classification_1.0 State change null -> WORKER_STARTED
2022-03-22T06:39:05,782 [INFO ] W-9026-dog_breed_wf_new__dog_breed_classification_1.0 org.pytorch.serve.wlm.WorkerThread - Connecting to: /home/model-server/tmp/.ts.sock.9026
2022-03-22T06:39:05,783 [INFO ] W-9026-dog_breed_wf_new__dog_breed_classification_1.0-stdout MODEL_LOG - Connection accepted: /home/model-server/tmp/.ts.sock.9026.
2022-03-22T06:39:05,783 [INFO ] W-9026-dog_breed_wf_new__dog_breed_classification_1.0 org.pytorch.serve.wlm.WorkerThread - Flushing req. to backend at: 1647931145783
+ 2022-03-22T06:39:05,783 [INFO ] W-9026-dog_breed_wf_new__dog_breed_classification_1.0-stdout MODEL_LOG - ==================================================
+ 2022-03-22T06:39:05,783 [INFO ] W-9026-dog_breed_wf_new__dog_breed_classification_1.0-stdout MODEL_LOG - model_name: dog_breed_wf_new__dog_breed_classification
+ 2022-03-22T06:39:05,783 [INFO ] W-9026-dog_breed_wf_new__dog_breed_classification_1.0-stdout MODEL_LOG - envelope: kservev2
2022-03-22T06:39:05,783 [INFO ] W-9026-dog_breed_wf_new__dog_breed_classification_1.0-stdout MODEL_LOG - model_name: dog_breed_wf_new__dog_breed_classification, batchSize: 4
2022-03-22T06:39:06,366 [INFO ] W-9025-dog_breed_wf_new__cat_dog_classification_1.0-stdout MODEL_LOG - Missing the index_to_name.json file. Inference output will not include class name.
2022-03-22T06:39:06,366 [INFO ] W-9025-dog_breed_wf_new__cat_dog_classification_1.0 org.pytorch.serve.wlm.WorkerThread - Backend response time: 1324
2022-03-22T06:39:06,366 [DEBUG] W-9025-dog_breed_wf_new__cat_dog_classification_1.0 org.pytorch.serve.wlm.WorkerThread - W-9025-dog_breed_wf_new__cat_dog_classification_1.0 State change WORKER_STARTED -> WORKER_MODEL_LOADED
2022-03-22T06:39:06,366 [INFO ] W-9025-dog_breed_wf_new__cat_dog_classification_1.0 TS_METRICS - W-9025-dog_breed_wf_new__cat_dog_classification_1.0.ms:2231|#Level:Host|#hostname:8b026932f222,tim
```

wjiuhe:
## Description

Support kserve v2 envelope with workflows.

Fixes #1529

### Problems to Fix

Currently:

### 1. The envelope of the workflow function is not set correctly.

> https://github.com/pytorch/serve/blob/6a191f822f2e759c2aad07b8f9eacbedcb039827/frontend/server/src/main/java/org/pytorch/serve/wlm/ModelManager.java#L122

While the models' envelope are set to the default envelope specified in config.properties, the workflow funcitons' envelope is null.

### 2. The request envelope `kservev2.py` cannot handle internal workflow requests and nested requests properly.

> https://github.com/pytorch/serve/blob/master/ts/torch_handler/request_envelope/kservev2.py

If a node has multiple parent nodes, for example node `A` with result `["a"]` and node `B` with result `["b"]`, the output of these node will be combined by the `frontend` into:

```json
[
    {
        "A": {"outputs": ["data": ["a"]]},
        "B": {"outputs": ["data": ["b"]]},
    }
]
```

However, because there is no key `data` or `body` (only `A` and `B`)in this json, the envelope class cannot process the data.

Even if the data is found, the kserve v2 envelope can only handle inputs key of `inputs` other than `outputs`.

Also in order to pass it to the model properly as it does with null envelope, the request data needs to be converted into the form:

```json
[
    "data": {
        "A": ["a"],
        "B": ["b"]
    }
]
```
Details are describe in the issue: https://github.com/pytorch/serve/issues/1529

### About the Commits and Changes

I divide the changes into seperate commits targeting different problems with description in the commit message. Therefore you can view the code changes with the detailed reason in the commit message.

Once approved, I will rebase and squash the commits.

Similar issues might exists in other envelopes. I'd like to contribute to fixing the rest of the envelopes as it will benefit my work as well. 

### How I fix it

Belows are what I've done. I'm sure it is far from acceptable, and I'd like your advice on how to fix it in the correct way.

1. Set the envelope of workflow functions to the default envelope as regular models (those have `.mar` files)
2. Pass a header from the frontend for all the internal workflow requests:
    - Header
        - Workflow-Request-Type
    - Possible Values
        - Flat
        - Nested
4. Update Kserve v2 envelope handler will handle the data according to the header.

## Type of change

~~ New feature (non-breaking change which adds functionality)~~
~~Breaking change (fix or feature that would cause existing functionality to not work as expected)~~
~~This change requires a documentation update~~

- [x] Test A
    1. On master branch.
    2. Don't set the envelope in `config.properties`. Run the container with the workflow mnist I provided above, or the example workflow dog_breed_classification. The prediction will succeed.
    3. Set the `service_envelope=kservev2`, run the container with the workflow in `2.`. Now the prediction will fail.
    4. Switch to this pull request branch.
    5. build the frontend and cp the jar into ts/frontend/.
    6. repeat `Step 3.`, the prediction will succeed.

## Checklist:

    - I will add the unittests asap, once the approach in this pull request seems acceptable to the maintainers.


shrinath-suresh:
@chauhang @jagadeeshi2i

jagadeeshi2i:
@wjiuhe Are you using the inputs from here for your tests https://github.com/pytorch/serve/blob/master/kubernetes/kserve/kf_request_json/v2/mnist/mnist_v2_bytes.json. Can you share your curl request.

wjiuhe:
> @wjiuhe Are you using the inputs from here for your tests https://github.com/pytorch/serve/blob/master/kubernetes/kserve/kf_request_json/v2/mnist/mnist_v2_bytes.json. Can you share your curl request.

Please find the curl request to test the workflow below, I believe it's equivalent to the inputs in your link.

```bash
curl --location --request POST 'http://localhost:8080/v2/workflows/mnist_workflow/infer' \
--header 'Content-Type: application/json' \
--data-raw '{
    "inputs": [
        {
            "data": "iVBORw0KGgoAAAANSUhEUgAAABwAAAAcCAAAAABXZoBIAAAAw0lEQVR4nGNgGFggVVj4/y8Q2GOR83n+58/fP0DwcSqmpNN7oOTJw6f+/H2pjUU2JCSEk0EWqN0cl828e/FIxvz9/9cCh1zS5z9/G9mwyzl/+PNnKQ45nyNAr9ThMHQ/UG4tDofuB4bQIhz6fIBenMWJQ+7Vn7+zeLCbKXv6z59NOPQVgsIcW4QA9YFi6wNQLrKwsBebW/68DJ388Nun5XFocrqvIFH59+XhBAxThTfeB0r+vP/QHbuDCgr2JmOXoSsAAKK7bU3vISS4AAAAAElFTkSuQmCC"
        }
    ]
}'
```

jagadeeshi2i:
@wjiuhe KServe service envelope is to support KServe data format in TS. KServe does not have endpoint for workflows. Are you planning to use this service envelope outside KServe. If you want json data format, you could use json service envelope. Pls let know you use case.

wjiuhe:
> @wjiuhe KServe service envelope is for using torchserve server in KServe. KServe does not have endpoint for workflows. Are you planning to use this service envelope outside KServe. If you want json data format, you could use json service envelope. Pls let know you use case.

## My Use Case

Hi, @jagadeeshi2i , I'm using torchserve to serve models and workflows, and I want to use the `kserve v2` api interface.

Please refer to the issue https://github.com/pytorch/serve/issues/1529.

## Problem

The problem is when configuring the `config.properties` to use `kservev2` service envelope, it breaks the workflow api. The workflow works using null envelope now won't work any more.

## Cause of the Problem

Actually I believe it to be a bug that affects not just the `kservev2` envelope but also other envelopes.

The main reason is: the functions of the workflow are registered with `null` envelope while the models of the workflow are registered with the specified envelope (in this case, it is the `kserve v2` envelope.

And also, workflows has compound inputs (like from both node A and node B to node C) and currently `kservev2` envelope cannot handle such case.

## What I Expected

The envelope is torchserve's interface definition. Switching envelopes should not lead to malfunction of the workflow.

## What I'm Try to Achieve

Please refer to the issue, or the commit message in the pull request  for more detailed info.

What I'm trying to achieve is: when the envelope is set to `kservev2`, the torchserve workflow api can still work without any change of the models/workflows.

jagadeeshi2i:
Hi @wjiuhe the KServe and KServev2 are for Torchserve, KServe integration and the changes you make to KServev2 envelope is not usable with KServe.

wjiuhe:
> Hi @wjiuhe the KServe and KServev2 are for Torchserve, KServe integration and the changes you make to KServev2 envelope is not usable with KServe. Note: KServe uses Kubeflow Pipelines for workflow building.

Hi, @jagadeeshi2i @shrinath-suresh @chauhang Perhaps I haven't been clear enough.

I'm not using Kserve. I'm using torchserve, triton and our runtimes with the kserve v2 api interface.

It may not be usable to kserve, however it is highly relevant to the use of torchserve(that is why I create the issue and pull request in torchserve project). Because when using the kserve v2 envelope, the workflow api will always return errors.

Regardless how crappy the implementation I might have done in the pull request, as a torchserve user, here are my questions:

### 1. There is an envelope registration problem in current torchserve frontend codes, should we fix it?(plz check out the issue or the commit message)

I think it should be fixed.

### 2. If a torchserve envelope (no matter it is a kserve or json envelope) causes the workflow api returning error, is it acceptable? Does that mean all the envelope should only work for the model API, not for the workflow API? 

Because from my point of view, the envelope is torchserve's interface of data processing, it is an internal thing, like the handlers, it shouldn't break any torchserve api, no matter it is for kserve or google cloud. Using an envelope will lead to the malfunction of workflow is problematic. 

### 3. Is it hard to modify the envelopes to support workflow? 

Not at all. So, why leave such a problematic or a buggy thing unfixed?

If this is a problem should be fixed, I'd like it to be fixed, by anyone.

wjiuhe:
> Hi @wjiuhe the KServe and KServev2 are for Torchserve, KServe integration and the changes you make to KServev2 envelope is not usable with KServe. Note: KServe uses Kubeflow Pipelines for workflow building.

By the way, I use k8s related during my daily work. I have also used kserve/kfserving, I do believe that is built on knative, and using torchserve in kserve doesn't mean I cannot use the workflow api. Actually I can access the workflow api, but just it won't work because the bug I mentioned here.

wjiuhe:
@maaquib @lxning Hi, I think this pull request may need some time to get merged due to the conversion above. I have separate the changes into two issues & pull requests. If you have time, plz help review this https://github.com/pytorch/serve/pull/1550 first. It's more urgent. And the kserve v2 envelope I can simple rewrite the codes in my own fork for now.

- I'm not sure adding extra headers from frontend is acceptable to you.It seems a bit more straight forward to me, so I don't need to guess whether the data is nested one or not. However, I'd like to modify the codes if there is a more suitable solution.
- I'm current adding postman tests for kserve api to test the workflows. I'm not very familiar with the postman format, but I will push the codes asap.

maaquib:
@wjiuhe Do you still want this to be merged or is this addressed by #1550?

wjiuhe:
> @wjiuhe Do you still want this to be merged or is this addressed by #1550?

Yes, I still want this to be merged.

Because this PR seems quite controversial, I put the commit which is not directly related to kserve in #1550.

#1550 is the prerequisite of this PR.

If possible, please help merge that first and I'll supply the test accordingly (without #1550, the test won't be able to pass).

Thanks you.

PR #1545, opened 2022-04-01
Optimized image preprocessing for image classification 

min-jean-cho:
We have been observing that TorchServe preprocessing time for image classification is a bottleneck - preprocessing time takes a very long time (longer than the actual inference time itself). You can verify this by modifying the [HandlerTime in the base_handler](https://github.com/pytorch/serve/blob/master/ts/torch_handler/base_handler.py#L228). 

The preprocessing overhead is mostly caused from [image_processing steps in the image_classifier](https://github.com/pytorch/serve/blob/master/ts/torch_handler/image_classifier.py#L21). 

The preprocessing time can be reduced by (1) removing unnecessary aten::copy operations; (2) replacing costly [torch.stack call in the vision_handler](https://github.com/pytorch/serve/blob/master/ts/torch_handler/vision_handler.py#L56); (3) using accimage package which uses the Intel IPP library. 

Below is the profiling result of this optimized preprocessing tested _without torchserve_ (just copied preprocessing code to another python file) tested on batch size = 128, and input image is the default torchserve [kitten.jpg](https://github.com/pytorch/serve/blob/master/examples/image_classifier/kitten.jpg) for image classification. The CPU time speedup is 3.03x. 

optimized_preprocessing_without_torchserve

Below is some sample results on applying this optimized preprocessing to TorchServe. The end-to-end throughput speedup is almost 2x. 

shared config:
[default torchserve ab params](https://github.com/pytorch/serve/blob/master/benchmarks/benchmark-ab.py#L18)
batch_size = 128
cpu_launcher_enabled=true
cpu_launcher_args=--node_id 0

**pytorch_fp32_original_preprocessing**
**ts e2e throughput**: 86.78
**preprocessing_time_mean**: 1212.27 

**pytorch_fp32_new_preprocessing**
**ts e2e throughput**: 163.17
**preprocessing_time_mean**: 457.78
**throughput speedup: 1.8802719520626872 
preprocessing time speedup: 2.6481497662632707**

Below are profiling results that show similar optimization as before without TorchServe. 
(1) ipex with original preprocessing 
ipex_original_preprocessing_ts

(2) ipex with optimized preprocessing 
ipex_optimized_preprocessing_ts

Currently, I've made the code change so that the _optimized_preprocessing is only applied when ipex_enable=true_. But this optimization is applicable to without ipex as well (as shown by the first profiling results outside of torchserve). 

Let me know of your thoughts, and am curious to know if this preprocessing overhead has already been an aware issue with current use-cases. Thanks.

### DIFF
--- a/ts/torch_handler/image_classifier.py
+++ b/ts/torch_handler/image_classifier.py
@@ -19,7 +19,7 @@ if os.environ.get("TS_IPEX_ENABLE", "false") == "true":
         import accimage
         ipex_enabled = True
     except ImportError as error:
-        logger.warning("Please instlal accimage for optimized preprocessing")
+        logger.warning("Please install accimage for optimized preprocessing")
  
 
 
--- a/ts/torch_handler/image_classifier.py
+++ b/ts/torch_handler/image_classifier.py
@@ -4,9 +4,23 @@ Module for image classification default handler
 import torch
 import torch.nn.functional as F
 from torchvision import transforms
+import logging
+import os
 
 from .vision_handler import VisionHandler
 from ..utils.util  import map_class_to_label
+
+logger = logging.getLogger(__name__)
+
+ipex_enabled = False
+if os.environ.get("TS_IPEX_ENABLE", "false") == "true":
+    ipex_enabled = True
+    try:
+        import accimage
+        ipex_enabled = True
+    except ImportError as error:
+        logger.warning("Please instlal accimage for optimized preprocessing")
+ 
 
 
 class ImageClassifier(VisionHandler):
@@ -18,13 +32,28 @@ class ImageClassifier(VisionHandler):
     topk = 5
     # These are the standard Imagenet dimensions
     # and statistics
-    image_processing = transforms.Compose([
-        transforms.Resize(256),
-        transforms.CenterCrop(224),
-        transforms.ToTensor(),
-        transforms.Normalize(mean=[0.485, 0.456, 0.406],
-                             std=[0.229, 0.224, 0.225])
-    ])
+    
+    if not ipex_enabled:
+        image_processing = transforms.Compose([
+            transforms.Resize(256),
+            transforms.CenterCrop(224),
+            transforms.ToTensor(),
+            transforms.Normalize(mean=[0.485, 0.456, 0.406],
+                                 std=[0.229, 0.224, 0.225])
+        ])
+    
+    else:
+        mean_ = [[[0.4850]],[[0.4560]],[[0.4060]]]
+        mean_ = torch.as_tensor(mean_)
+        
+        std_ = [[[0.229]],[[0.224]],[[0.225]]]
+        std_ = torch.as_tensor(std_)
+        
+        
+        image_processing = transforms.Compose([
+            transforms.ToTensor(),
+            transforms.Normalize(mean=mean_,std=std_, inplace=True)
+        ])
 
     def set_max_result_classes(self, topk):
         self.topk = topk
--- a/ts/torch_handler/vision_handler.py
+++ b/ts/torch_handler/vision_handler.py
@@ -8,8 +8,21 @@ import io
 import base64
 import torch
 from PIL import Image
+import logging
+import os
 from captum.attr import IntegratedGradients
 from .base_handler import BaseHandler
+
+logger = logging.getLogger(__name__)
+
+ipex_enabled = False
+if os.environ.get("TS_IPEX_ENABLE", "false") == "true":
+    ipex_enabled = True
+    try:
+        import accimage
+        ipex_enabled = True
+    except ImportError as error:
+        logger.warning("Please instlal accimage for optimized preprocessing")
 
 
 class VisionHandler(BaseHandler, ABC):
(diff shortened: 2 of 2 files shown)


min-jean-cho:
cc. @msaroufim @ashokei

ts/torch_handler/image_classifier.py
```diff
@@ -18,13 +32,28 @@ class ImageClassifier(VisionHandler):
...
     # and statistics
-    image_processing = transforms.Compose([
-        transforms.Resize(256),
-        transforms.CenterCrop(224),
-        transforms.ToTensor(),
-        transforms.Normalize(mean=[0.485, 0.456, 0.406],
-                             std=[0.229, 0.224, 0.225])
-    ])
+    
+    if not ipex_enabled:
```
msaroufim, line 36: why are there 2 code paths for image preprocessing?

msaroufim (changes requested):

ts/torch_handler/image_classifier.py
```diff
@@ -4,10 +4,24 @@
...
+logger = logging.getLogger(__name__)
+
+ipex_enabled = False
+if os.environ.get("TS_IPEX_ENABLE", "false") == "true":
+    ipex_enabled = True
+    try:
+        import accimage
+        ipex_enabled = True
+    except ImportError as error:
+        logger.warning("Please instlal accimage for optimized preprocessing")
```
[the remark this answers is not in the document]
min-jean-cho: OK, it is indeed generally useful.

ts/torch_handler/image_classifier.py
min-jean-cho: I've just made the code change so that the _optimized_preprocessing is only applied when ipex_enable=true. But this optimization is generalizable to without ipex as well. We do not need 2 code paths.

msaroufim:
> (1) removing unnecessary aten::copy operations; (2) replacing costly [torch.stack call in the vision_handler](https://github.com/pytorch/serve/blob/master/ts/torch_handler/vision_handler.py#L56); (3) using accimage package which uses the Intel IPP library.


1.  Where were these removed, can't see
2. Can we remove torch.stack for all code paths? Amazing find, thank you!
3. Seems great I'm just asking internally what's the state of all the torchvision backends before we merge

ts/torch_handler/image_classifier.py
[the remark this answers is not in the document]
HamidShojanazeri: can you pls elaborate a bit how it help with preprocessing optimizations. 

is it necessary to import accimage here? it seems been only used in vision handler.

HamidShojanazeri, line 22:
```suggestion
        logger.warning("Please install accimage for optimized preprocessing")
```

HamidShojanazeri:
> Let me know of your thoughts, and am curious to know if this preprocessing overhead has already been an aware issue with current use-cases.

@min-jean-cho  This has been historically a bottleneck specially with large payload. It sounds promising.

min-jean-cho:
> 



> > (1) removing unnecessary aten::copy operations; (2) replacing costly [torch.stack call in the vision_handler](https://github.com/pytorch/serve/blob/master/ts/torch_handler/vision_handler.py#L56); (3) using accimage package which uses the Intel IPP library.
> 
> 1. Where were these removed, can't see
> 2. Can we remove torch.stack for all code paths? Amazing find, thank you!
> 3. Seems great I'm just asking internally what's the state of all the torchvision backends before we merge

There are total 5 aten::copy ops per batch in the original preporcessing, 128 * 5=640 (128 is the batch size).  
After the optimiazation, there are total 2 aten::copy ops remaining, 128 * 2 = 256. Please see the profiling images above.

 The 3 aten::copy ops were removed here:

1. 2 of them are removed from [transforms.ToTensor()](https://github.com/pytorch/serve/blob/master/ts/torch_handler/image_classifier.py#L24) in the image_classifier. 
- [.contiguous](https://github.com/pytorch/vision/blob/ec1c2a12cf00c6df83c7fb88f75b8117cda2f970/torchvision/transforms/functional.py#L170)
- [.to](https://github.com/pytorch/vision/blob/ec1c2a12cf00c6df83c7fb88f75b8117cda2f970/torchvision/transforms/functional.py#L172)
Both are removed by using [accimage path](https://github.com/pytorch/vision/blob/ec1c2a12cf00c6df83c7fb88f75b8117cda2f970/torchvision/transforms/functional.py#L157) instead. 

2. 1 is removed from [transforms.Normalize()](https://github.com/pytorch/serve/blob/master/ts/torch_handler/image_classifier.py#L25) in the image_classifier. 
- [.clone](https://github.com/pytorch/vision/blob/ec1c2a12cf00c6df83c7fb88f75b8117cda2f970/torchvision/transforms/functional_tensor.py#L950)
This can be removed by setting [inplace](https://github.com/pytorch/vision/blob/ec1c2a12cf00c6df83c7fb88f75b8117cda2f970/torchvision/transforms/functional_tensor.py#L949)=True

ts/torch_handler/image_classifier.py
min-jean-cho: Hi @HamidShojanazeri , please see my below comment on usage of accimage to remove 2 aten::copy ops. And it's correct that accimage is only used in vision handler.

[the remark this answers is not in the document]
min-jean-cho: Hi @HamidShojanazeri , please see my below response to Mark on usage of accimage to remove 2 aten::copy ops. And it's correct that accimage is only used in vision handler.

msaroufim:
@min-jean-cho it seems like accimage doesn't really have a maintainer as of today so it may be best to keep out those changes for now and just use the native backends in torchvision

For the other changes around removing unnecessary copies I'm all in and I think we should merge those asap

min-jean-cho:
> @min-jean-cho it seems like accimage doesn't really have a maintainer as of today so it may be best to keep out those changes for now and just use the native backends in torchvision
> 
> For the other changes around removing unnecessary copies I'm all in and I think we should merge those asap

Hi @msaroufim , I had a quick test without accimage but keeping other changes. But I observed there's little to no performance difference without accimage as most of copy ops were removed by use of accimage. While the inplace=True removes 1 copy ops per batch, it is compensated by an addition of 1 copy ops when images[i] = image. So the net gain is zero (i.e., same number of copy ops before and after). While the costly aten::stack is replaced, the performance gain is not as much as before with throughput of 1.88x when using accimage. Below is a sample result:

**pytorch_fp32_original_preprocessing**
**ts e2e throughput:** 85.98
**preprocessing_time_mean:** 1223.39

**pytorch_fp32_new_preprocessing**
**ts e2e throughput:** 85.76
**preprocessing_time_mean:** 1227.89

The profiling results are below. 
(1) original preprocessing
Screenshot (532)_LI

(2) new preprocessing
Screenshot (533)_LI

msaroufim:
May be worth looking into Pillow with SIMD https://github.com/pytorch/vision#image-backend

min-jean-cho:
> May be worth looking into Pillow with SIMD https://github.com/pytorch/vision#image-backend

Thanks for the pointer. I'll take a look at Pillow SIMD.

msaroufim:
@min-jean-cho to close on this issue I believe the right solution is to instead of a default integration author a tutorial, happy to feature you again as your blogs are fantastic. Let me know if you'd rather continue using this PR or leverage another one

min-jean-cho:
> @min-jean-cho to close on this issue I believe the right solution is to instead of a default integration author a tutorial, happy to feature you again as your blogs are fantastic. 

Thanks @msaroufim, sounds good - we can close this PR. I'd be happy to author a tutorial/blog on this topic of profiling with pytorch profiler to optimize torchserve image preprocessing, what do you think? Would this be published on https://pytorch.org/tutorials/ similar to our first joint blog https://pytorch.org/tutorials/intermediate/torchserve_with_ipex.html?

msaroufim:
As we discussed offline let's turn to a blog for this contribution and we can revisit later

PR #1555, opened 2022-04-08
Use sys.executable instead of hard coding

nijkah:
## Description

Predefined 'python' can be different from the command which is really executed.
So replace 'python' with sys.executable

Fixes #(issue)

## Type of change

## Checklist:

### DIFF
--- a/ts_scripts/install_dependencies.py
+++ b/ts_scripts/install_dependencies.py
@@ -1,7 +1,6 @@
 import os
 import platform
 import argparse
-from shutil import which
 import sys
 from pathlib import Path
 from print_env_info import run_and_parse_first_match
--- a/ts_scripts/install_dependencies.py
+++ b/ts_scripts/install_dependencies.py
@@ -1,6 +1,7 @@
 import os
 import platform
 import argparse
+from shutil import which
 import sys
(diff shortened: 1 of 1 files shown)


msaroufim (approved):

maaquib:
@nijkah Thanks for fixing this. Can you also verify the [sanity suite](https://github.com/pytorch/serve/blob/master/CONTRIBUTING.md) runs fine and then I can merge this

nijkah:
@maaquib I thought `python torchserve_sanity.py` had been done in `AWS CodeBuild CI Report`s.
Should I run it in my environment? I tested it in the past, but it requires quite a lot of resources compared to my environment.

HamidShojanazeri (approved):

ts_scripts/install_dependencies.py
```diff
@@ -1,6 +1,7 @@
 import os
 import platform
 import argparse
+from shutil import which
```
mreso, line 4: This seems to be an unused import

mreso (approved):

msaroufim (changes requested):
let
s address mathias feedback

PR #1562, opened 2022-04-13
docker nightly builds

### LINKED ISSUE #1558: Nightly build for Docker images
Nightly build for Docker images that would use `pip install torchserve-nightly` by default 

I'd rather create a different org from https://hub.docker.com/r/pytorch/torchserve called `torchserve-nightly`

msaroufim:
Images will be here: https://hub.docker.com/repository/docker/pytorch/torchserve-nightly
Tested here: https://github.com/msaroufim/serve/actions

Fixes #1558


docker/docker_nightly.py
```diff
@@ -0,0 +1,26 @@
...
+    # Build Nightly images and append the date in the name
+    os.system(f"./build_image.sh -bt dev -t pytorch/{cpu_version}")
+    os.system(f"./build_image.sh -bt dev -g -cv cu102 -t pytorch/{gpu_version}")
+
+    # Push Nightly images to official PyTorch Dockerhub account
+    for version in [cpu_version, gpu_version]:
+        os.system(f"docker push pytorch/{version}")
+    
+    # Tag images with latest and push those as well
+    os.system(f"docker tag pytorch/{cpu_version} pytorch/{project}:latest-cpu")
```
maaquib, line 22: NIT: Wondering if we should tag the nightlies as latest or the last stable release?

msaroufim: Since it's nightlies there's no guarantee of stability. For now I'm using a completely different project name `torchserve-nightly` so this doesn't mix up with our regular stable docker images https://hub.docker.com/repository/docker/pytorch/torchserve-nightly

maaquib (approved):

docker/docker_nightly.py
```diff
@@ -0,0 +1,26 @@
...
+    return today.strftime("%Y.%m.%d")
+
+if __name__ == "__main__":
+    project = "torchserve-nightly"
+    cpu_version = f"{project}:cpu-{get_nightly_version()}"
+    gpu_version = f"{project}:gpu-{get_nightly_version()}"
+
+    # Build Nightly images and append the date in the name
+    os.system(f"./build_image.sh -bt dev -t pytorch/{cpu_version}")
+    os.system(f"./build_image.sh -bt dev -g -cv cu102 -t pytorch/{gpu_version}")
```
mreso, line 15: Is there a reason we're building this with cu102 and not cu113?

.github/workflows/docker-nightly-buid.yml
```diff
@@ -0,0 +1,26 @@
+name: Push Docker Nightly
```
mreso, line 1: Nit: Filename is missing an 'l'

mreso (approved):

.github/workflows/docker-nightly-buid.yml
msaroufim: Ah whoops thank you for catching this

docker/docker_nightly.py
msaroufim: Our experience is generally it's been more stable but if that's no longer the case if we just update this line to `cu1113` and the nightly builds will work just fine

mreso:
Oh I see, I was just curious

PR #1564, opened 2022-04-15
Fix model-archiver doc link

maaquib:
## Description

Fix model-archiver doc link

## Type of change
Doc change


msaroufim (approved):

msaroufim:
Well actually this won't quite work, we need absolute links since that link is also used on pytorch.org/serve where it won't have access to the whole github repo.

I can add something like a linter to detect these issues proactively

Issue #1586, opened 2022-04-26
Flaky workflow tests on windows


msaroufim:
CI on Windows is broken because of an old version of Gradle https://github.com/pytorch/serve/runs/6167464335?check_suite_focus=true#step:6:5031

msaroufim:
Removing urgent since it seems flaky but this is kind of a weird flake

msaroufim:
Duplicate

PR #1601, opened 2022-05-02
Delete WorkflowTest.java

msaroufim:
Disabled flaky tests #1586

There are open issues to address them


lxning:
We should keep unit test.

Issue #1546, opened 2022-04-01
Accelerated preprocessing


msaroufim:
This recent PR by @min-jean-cho https://github.com/pytorch/serve/pull/1545 has made me realize how much performance we're leaving on the table when it comes to acceleration of preprocessing operations. We could more easily support more torchvision backends, faster tokenizers and add more hardware specific configurations to make it for developers to pick and choose.

This will also help our #1438 efforts

msaroufim:
For accelerated decoding, these are all the endpoints we can improve

```
(base) ➜  serve git:(decode) ✗ find . -name "*.py" | xargs grep "decode" . 
grep: .: Is a directory
./test/pytest/test_handler.py:    response = response.content.decode("utf-8")
./test/pytest/test_gRPC_inference_api.py:    prediction = response.prediction.decode('utf-8')
./ts_scripts/print_env_info.py:    output = output.decode(enc)
./ts_scripts/print_env_info.py:    err = err.decode(enc)
./ts_scripts/torchserve_grpc_client.py:        prediction = response.prediction.decode('utf-8')
./examples/nmt_transformer/model_handler_generalized.py:            decoded_text = text.decode('utf-8')
./examples/nmt_transformer/model_handler_generalized.py:            textInput.append(decoded_text)
./examples/text_to_speech_synthesizer/waveglow_handler.py:        text = text.decode('utf-8')
./examples/Workflows/dog_breed_classification/dog_breed_classification_handler.py:            cat_dog_response = row.get("cat_dog_classification").decode()
./examples/Workflows/dog_breed_classification/dog_breed_classification_handler.py:            input_data = row.get("pre_processing").decode()
./examples/Workflows/dog_breed_classification/cat_dog_classification_handler.py:            b64_data.append({"body": base64.b64decode(input_data)})
./examples/Workflows/dog_breed_classification/workflow_dog_breed_classification_handler.py:        b64_data.append(base64.b64encode(input_data).decode())
./examples/Huggingface_Transformers/Transformer_handler_generalized.py:                input_text = input_text.decode("utf-8")
./examples/Huggingface_Transformers/Transformer_handler_generalized.py:                    self.tokenizer.decode(input_ids_batch[i])
./examples/Huggingface_Transformers/Transformer_handler_generalized.py:                    self.tokenizer.decode(outputs[i], skip_special_tokens=True)
./examples/Huggingface_Transformers/Transformer_handler_generalized.py:            text = text.decode("utf-8")
[...]

PR #1641, opened 2022-05-19
[WIP] DALI integration

msaroufim:
## Description

Unfortunately DALI bundles its image decoder function in their pipeline language so we can't just use a part of DALI.
Also DALI insists on reading images from disk directly (there are workarounds) but outputs of a pipeline can only be read by implementing a data loader. Overall I can't see us integrating DALI in a handler but it should be possible with a custom python backend. Parking this work for now until I hopefully find something simpler.

#1546 

## Type of change

- [x] This change requires a documentation update - will require documentation update to discuss how to pick optimal settings for DALI

Will use Li's benchmarking suite before and after this change for image models to decide whether to ship or not

- [ ] Test A
- [ ] Test B

- Logs

## Checklist:

- [ ] Did you enable `pre-commit` to check and format your changes before pushing your changes?

### DIFF
--- a/ts/torch_handler/image_classifier.py
+++ b/ts/torch_handler/image_classifier.py
@@ -18,16 +18,12 @@ class ImageClassifier(VisionHandler):
     topk = 5
     # These are the standard Imagenet dimensions
     # and statistics
-    mean_ = [[[0.4850]],[[0.4560]],[[0.4060]]]
-    mean_ = torch.as_tensor(mean_)
-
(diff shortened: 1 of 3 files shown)


msaroufim:
Next steps 
1. Remove `torch.stack()` and `to(device)` operations in handler
2. Run regression suite
3. Look at performance improvements, if low then start work integrating custom backend for DALI

msaroufim:
Will continue work in my own private branch

PR #1661, opened 2022-05-31
debug standalone handler

### LINKED ISSUE #1538: Debugging standalone handlers
## 📚 Documentation

Related to #1504

Often when new users are debugging their model handler files, they'll make some change then package the model and then inspect logs in `model_log.log` while this is fast enough for advanced users for new users it's very confusing since you don't actually see an error in your console and there's a lag to package models especially if they're large.

Instead the insight is that model handlers are just python files, the way torchserve works is the Java frontend spawns a new process and runs the handler. But we could do that as well with a `main()` function

So assume your `handler.py` is a typical handler from `examples` but also now ends with the below

```python
if __name__ =="__main__":
  handler = CustomHandler()
  handler.initialize()
  x = torch.randn()
  x = handler.preprocess(x)
  x = handle.inference(x)
  x = handler.postprocess(x)
  print(x)
```

The nice thing about this approach you can now add `print()` statements, you can add a debugger and you can run a profiler on the python file like `scalene handler.py`

msaroufim:
## Description

This PR introduce the ability to debug standalone handlers, many customer issues are opened here which bugs in handler code instead of bugs in torchserve. The error message is scary looking and makes it seem like an internal torchserve failure `Backend worker monitoring thread interrupted or backend worker process died`

However because of the lengthy packaging and torchserve start process it makes interactive debugging a pain. 

Instead torchserve handlers are nothing but regular python files which can be run in the following way

```python
if __name__ == "__main__":
    handler = BaseHandler()
    handler.initialize()
    x = torch.randn()
    x = handler.preprocess(x)
    x = handler.inference(x)
    x = handler.postprocess(x)
    print(x)
```

However our handlers as is used relative imports which would make them fail unless they were called from a specific directory so we've replaced statements like `from ..utils.util import list_classes_from_module, load_label_mapping`

by 
`from ts.utils.util import list_classes_from_module, load_label_mapping` to make this work and we've added documentation to in `docs/FAQ.md` to describe this feature to users.

While I was here I also moved contractions to the `utils` folder since it's not a handler and remaining changes are done via `pre-commit`

Fixes #1538 

## Type of change

CI is enough, because if this didn't work it would break all of CI

## Checklist:

- [x] Did you have fun?

### DIFF
--- a/docs/FAQs.md
+++ b/docs/FAQs.md
@@ -174,3 +174,5 @@ if __name__ == "__main__":
     x = handler.postprocess(x)
     print(x)
 ```
+
+And then you can just run `python base_handler.py`
--- a/docs/FAQs.md
+++ b/docs/FAQs.md
@@ -116,7 +116,7 @@ Refer [default handlers](default_handlers.md#torchserve-default-inference-handle
 
 ### Is it possible to deploy Hugging Face models?
 Yes, you can deploy Hugging Face models using a custom handler.
-Refer [HuggingFace_Transformers](https://github.com/pytorch/serve/blob/master/examples/Huggingface_Transformers/README.md) for example. 
+Refer [HuggingFace_Transformers](https://github.com/pytorch/serve/blob/master/examples/Huggingface_Transformers/README.md) for example.
 
 ## Model-archiver
  Relevant documents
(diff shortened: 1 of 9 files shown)


mreso (changes requested):
When you run handler.initialize() a context is necessary. Is there a way to create a default context somehow? Or mock one for the testing purpose?

msaroufim:
> When you run handler.initialize() a context is necessary. Is there a way to create a default context somehow? Or mock one for the testing purpose?

Oh yeah great point, we can and here's an example https://github.com/pytorch/serve/blob/master/ts/torch_handler/unit_tests/test_utils/mock_context.py

Will update the docs to make this clearer

Issue #1631, opened 2022-05-14
Modularize `base_handler.py` into `handler.utils`


msaroufim:
## Modularize `base_handler.py` into `handler.utils`
This is a follow-up note on #1440 where we initially noted a few problems that were making it difficult to add code to TorchServe

Initially, the `base_handler` had a few simple responsibilities
1. Loading a model
2. Calling an inference
3. Registering some simple metrics

But over time, contributors found it easy to extend functionality in torchserve by extending the handler because it involved **python-only** changes and we've been reluctant to fully embrace this approach since more natively integrating within the `WorkloadManager.java` makes it so you can gate certain functionality behind a `config.properties` similar to https://github.com/pytorch/serve/pull/1319. However for more complex features the code review process becomes more of a challenge and grinds to a halt https://github.com/pytorch/serve/pull/1401

At the same time, we've had `examples/` grow in size and complexity like https://github.com/pytorch/serve/tree/master/examples/Huggingface_Transformers without having its improvements be easily usable by other parts of the code base

> So how can we encourage users to contribute their ideas to the easy-to-extend handlers without creating a giant monolithic file? How do we encourage modularity and code reuse?

## Describe the solution
Instead of designing things in the abstract, it's important to look at what contributors have actually added to handlers. Here's a non-exhaustive list

1. Environment variables: Added support for environment variable checking to gate certain behaviors https://github.com/pytorch/serve/blob/master/ts/torch_handler/base_handler.py#L25
2. Profiling: Added support for pytorch profiler https://github.com/pytorch/serve/blob/master/ts/torch_handler/base_handler.py#L218
3. Model interpretability: With captum https://github.com/pytorch/serve/blob/master/ts/torch_handler/base_handler.py#L287
4. Model optimization: With IPEX https://github.com/pytorch/serve/blob/master/ts/torch_handler/base_handler.py#L96
5. Error handling on batch processing
6. Utilities around pulling HuggingFace pretrained models: https://github.com/pytorch/serve/blob/master/examples/Huggingface_Transformers/Transformer_handler_generalized.py#L72
7. Model parallelism: https://github.com/pytorch/serve/blob/master/examples/Huggingface_Transformers/Transformer_handler_generalized.py#L84
8. Optimized preprocessing: via libraries like `DALI` or `accimage` https://github.com/pytorch/serve/pull/1545

[...]

Issue #1492, opened 2022-03-07
[RFC]: Metrics Refactoring


lxning:
### Current TorchServe has two mechanisms to emit metrics.

1. [Emit metrics to logs files in a StatsD like format by default](https://github.com/pytorch/serve/blob/master/docs/metrics.md#formatting) .

In this case, both frontend and backend metrics are recorded in log file. However, the logs format is not standard StatsD format. They miss the metric type information (ie. counter, gauge, timer and so on). Users have to write regex to parse the log to build dashboard. 

2. [Emit Prometheus formatted metrics](https://github.com/pytorch/serve/blob/master/docs/metrics_api.md#metrics-api).

In this case, existing TorchServe only emits 3 metrics.

- ts_inference_requests_total
- ts_inference_latency_microseconds
- ts_queue_latency_microseconds

Users are not able to get model metrics and system metrics via metrics endpoint.

### No central place to store  Metrics definition 
Existing TorchServe metrics definitions spread everywhere. It is difficult for users to know the available metrics.

### Re-Design

[TS_Metrics_Design.pdf](https://github.com/pytorch/serve/files/9892287/TS_Metrics_Design.pdf)

msaroufim:
For the requirements here need to make sure to include
* Adding full coverage for frontend and backend metrics with prometheus export
* Instructions to make it easy for anyone to export their own metrics to prometheus

thimabru1010:
Yes, I only can see these 3 metrics in prometheus. I tried to change model log behavior with a log4j2.xml file, but with no success. How do I get system metrics in prometheus?

jonhilgart22:
Custom metrics would be great. I'd like to add some histogram metrics for Prometheus to see latency percentiles.

sharvil:
Refactoring the existing metrics implementation is a welcome change. After reading through the proposed design, I recommend against introducing the `metrics.yaml` file.

In our case, we have model-specific performance metrics that we'd like to log and monitor. The ideal scenario is for each model archive to manage its own metrics. As those archives are loaded / unloaded they would register / unregister their metrics and those changes would be reflected in the metrics exposed by the frontend.

Defining a global metrics schema in `metrics.yaml` would require us to update the config separately every time a model is loaded or unloaded. In addition, doing so is repetitive since the proposed `MetricsCaching.addMetric` method already needs us to specify all the properties that would be in `metrics.yaml`.

sharvil:
[...]

PR #1664, opened 2022-06-01
Modularize ipex optimization in `base_handler.py` into `ts/utils/ipex_optimization.py`. 

min-jean-cho:
This PR

1. updates IPEX integration into TorchServe following #1631 . 

As described in #1631,
**Model optimization now**
```python
# in base_handler.py
if ipex_enabled:
    self.model = self.model.to(memory_format=torch.channels_last)
    self.model = ipex.optimize(self.model)

```
**Model optimization after**
```python
# in ts/utils/optimization.py
class Optimization():
  def __init__(model):
    self.model = model
  def optimize(self) -> torch.nn.Module:
    raise NotImplementedError("This is an abstract base class, you need to call or create your own runtime")

# in ts/utils/ipex_optimization.py
class IPEXOptimization(Optimization):
  def __init__(self, model, **kwargs)
    super().__init__(model)
  def optimize(self):
    self.model = ipex.optimize(self.model)
    return self.model 

# in base_handler.py
if ipex_enabled or onednn_graph_fusion_enabled:
    self.optimization = OPTIMIZATIONS["ipex"](self.model, ipex_enabled, onednn_graph_fusion_enabled)
    self.model = self.optimization.optimize()
```

2. Integrates IPEX optimization features bfloat16 auto mixed precision, oneDNN graph fusion, int8, etc into `ts/utils/ipex_optimization.py`

2.1 Refactors IPEX integration with more flexible interface for IPEX features. 
Let's demonstrate with an exemplary IPEX feature, int8. Currently to create IPEX int8 mar files, users have to follow step-by-step guidelines like the one given [here](https://github.com/pytorch/serve/tree/master/examples/intel_extension_for_pytorch#creating-and-exporting-int8-model-for-ipex). Now instead, users can simply provide .mar file without any ipex optimization as they usually would, and specify in `config.properties` as follows:
```
ipex_enable=true
ipex_dtype=int8
ipex_torchscript=true
ipex_input_tensor_shapes=1, 3, 224, 224
ipex_input_tensor_dtype=TYPE_FP32
```
`ts/utils/ipex_optimization.py` will then take care of the IPEX int8 optimization. 

With the base_handler modularized, future IPEX feature integration will all be located in (1) `ts/utils/ipex_optimization.py` with minor (if any) changes to the `base_handler.py`, and (2) `ConfigManager.java` as needed.


ts/utils/__init__.py
```diff
@@ -5,3 +5,14 @@
 """
 
 from . import timeit_decorator
+from .optimization import OPTIMIZATIONS
+from os.path import dirname, basename, isfile, join
+import glob
+
+modules = glob.glob(join(dirname(__file__), "*.py"))
```
msaroufim, line 12: add a comment explaining what this is trying to do

ts/utils/ipex_optimization.py
```diff
@@ -0,0 +1,144 @@
+from .optimization import optimization_registry, Optimization
```
msaroufim, line 1: haha I can see how the `base_handler.py` was constraining you

msaroufim (changes requested):
I am super glad to see this change in - I think what's left is
1. Tests for the utils functions
2. Not sure why the base_handler has more ipex related code than before now
3. Add comments to hard to understand parts especially the model registry
4. Update `CONTRIBUTING.md` or `docs/performance_guide.md` on how new optimizations should be added

This is a great contribution thank you for just going with it

ts/utils/ipex_optimization.py
```diff
@@ -0,0 +1,144 @@
...
+                    self.model(*x)
+                # conversion 
+                self.model = convert(self.model)
+    
+        # torchscript
+        if self.torchscript == "true":
+            x = self.get_dummy_tensors()
+            
+            if self.dtype == "float32" or self.dtype == "int8":
+                self.model = self.get_torchscripted_model(self.model, x)
```
HamidShojanazeri, line 139: Thanks @min-jean-cho for the PR. In the [current version](https://github.com/pytorch/serve/blob/master/ts/torch_handler/base_handler.py#L103), seems ipex optimizations are limited to ipex.optimize(). Can you please elaborate a bit more on the use of torchscript here? would torch scripting the ipex_optimized model help with the perf?

HamidShojanazeri: Also, the same for torch.cpu.amp.autocast()

frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
```diff
@@ -68,6 +68,16 @@ public final class ConfigManager {
...
+    private static final String TS_IPEX_DTYPE = "ipex_dtype";
+    private static final String TS_IPEX_CHANNEL_LAST = "ipex_channel_last";
+    
+    private static final String TS_IPEX_TORCHSCRIPT = "ipex_torchscript";
+    private static final String TS_IPEX_INPUT_TENSOR_SHAPES = "ipex_input_tensor_shapes";
+    private static final String TS_IPEX_INPUT_TENSOR_DTYPE = "ipex_input_tensor_dtype";
+    private static final String TS_IPEX_QSCHEME = "ipex_qscheme";
+    // oneDNN Graph fusion config option that can be set at config.properties 
+    private static final String TS_ONEDNN_GRAPH_FUSION_ENABLE = "onednn_graph_fusion_enable";
+    // launcher config option that can be set at config.properties
```
lxning, line 80: are these parameters more related to each model? If yes, they should be set in each model's configuration.

frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
```diff
@@ -699,7 +709,16 @@ public HashMap<String, String> getBackendConfiguration() {
...
+        
         config.put("TS_IPEX_ENABLE", prop.getProperty(TS_IPEX_ENABLE, "false"));
+        config.put("TS_IPEX_DTYPE", prop.getProperty(TS_IPEX_DTYPE, "float32"));
+        config.put("TS_IPEX_CHANNEL_LAST", prop.getProperty(TS_IPEX_CHANNEL_LAST, "true"));
+        config.put("TS_IPEX_TORCHSCRIPT", prop.getProperty(TS_IPEX_TORCHSCRIPT, "true"));
+        config.put("TS_IPEX_INPUT_TENSOR_SHAPES", prop.getProperty(TS_IPEX_INPUT_TENSOR_SHAPES, null));
+        config.put("TS_IPEX_INPUT_TENSOR_DTYPE", prop.getProperty(TS_IPEX_INPUT_TENSOR_DTYPE, null));
+        config.put("TS_IPEX_QSCHEME", prop.getProperty(TS_IPEX_QSCHEME, "per_tensor_affine"));
+        config.put("TS_ONEDNN_GRAPH_FUSION_ENABLE", prop.getProperty(TS_ONEDNN_GRAPH_FUSION_ENABLE, "false"));
+
```
lxning, line 721: Same here, they should be each model's config if each model can have different setting.

lxning: And also, as I commented in [issue1631](https://github.com/pytorch/serve/issues/1631#issuecomment-1129443165), it is not safe by setting environment. We need a secure solution.

msaroufim:
Hi @min-jean-cho I apologize but can you please pause your work here. After discussing with @lxning offline, we believe a better long term proposal would be to extend the message format we use between the frontend and the backend. The ConfigManager should really be for model specific configs that a user needs to care about

For example this what our current message protocol looks like https://github.com/pytorch/serve/blob/master/ts/protocol/otf_message_handler.py#L183-L193 - I'll take a look next week at extending it

min-jean-cho:
> Hi @min-jean-cho I apologize but can you please pause your work here. After discussing with @lxning offline, we believe a better long term proposal would be to extend the message format we use between the frontend and the backend. The ConfigManager should really be for model specific configs that a user needs to care about. Environment variables are indeed convenient but they don't expose the full power of torchserve as a multi model framework.
> 
> For example this what our current message protocol looks like https://github.com/pytorch/serve/blob/master/ts/protocol/otf_message_handler.py#L183-L193 - I'll take a look next week at extending it

Hi @msaroufim , no problem. Will stay tuned for an update on the long term proposal with extended message format. Thanks !

msaroufim:
Hi  @minjeanc, let's revisit this PR

Instead of using environment variables you can create your own config file called `ipex_config.json/yaml/txt/etc..` , pass it in using `--extra-files` when archiving a model https://github.com/pytorch/serve/tree/master/model-archiver#arguments - JSON is what we've mostly been using

And then from the handler check if that file exists (example: https://github.com/pytorch/serve/blob/master/ts/torch_handler/base_handler.py#L108), parse it and set all the configurations you need, this should also let you revert the changes to ConfigManager.java so even moving forward you should only need to write python code

min-jean-cho:
> Hi @minjeanc, let's revisit this PR
> 
> Instead of using environment variables you can create your own config file called `ipex_config.json/yaml/txt/etc..` , pass it in using `--extra-files` when archiving a model https://github.com/pytorch/serve/tree/master/model-archiver#arguments - JSON is what we've mostly been using
> 
> And then from the handler check if that file exists (example: https://github.com/pytorch/serve/blob/master/ts/torch_handler/base_handler.py#L108), parse it and set all the configurations you need, this should also let you revert the changes to ConfigManager.java so even moving forward you should only need to write python code

Thanks @msaroufim, I will have a look at this.

min-jean-cho:
> Instead of using environment variables you can create your own config file called ipex_config.json/yaml/txt/etc.. , pass it in using --extra-files when archiving a model https://github.com/pytorch/serve/tree/master/model-archiver#arguments - JSON is what we've mostly been using
> 
> And then from the handler check if that file exists (example: https://github.com/pytorch/serve/blob/master/ts/torch_handler/base_handler.py#L108), parse it and set all the configurations you need, this should also let you revert the changes to ConfigManager.java so even moving forward you should only need to write python code

Hi @msaroufim, have made some updates. Let me know what you think of the updates.

The updates are:
- add ipex optimization config and schema in `ts/torch_handler/utils/conf/ipex_config.py`
User can pass their ipex optimization config via `--extra_files ipex_config.yaml`
The choice of yaml vs. json was arbitrary -- happy to go with either option. 
The config file format can be documented in the ipex README.
In the future, other custom optimization config (and schema) can be added in `ts/torch_handler/utils/conf/custom_config.py`

- add ipex optimization in `ts/torch_handler/utils/optimization/ipex_optimization.py`
In the future, other custom optimization can be added in `ts/torch_handler/utils/optimization/custom_optimization.py`

- in the `base_handler`, ipex optimization is invoked following its config 
In the future, other custom optimization can similarly be invoked in the `base_handler`. 

The base_handler now looks like:
**Model optimization now**
```python
# in base_handler.py
if ipex_enabled:
    self.model = self.model.to(memory_format=torch.channels_last)
    self.model = ipex.optimize(self.model)
```

**Model optimization after**
```python
# in base_handler.py
if ipex_enabled:
    cfg_file_path = os.path.join(model_dir, "ipex_config.yaml")
    self.cfg = CONFIGURATIONS["ipex"](cfg_file_path)
    self.optimization = OPTIMIZATIONS["ipex"](self.cfg)

    self.model = self.optimization.optimize(self.model)
```

ts/torch_handler/utils/optimization/optimization.py
```diff
@@ -0,0 +1,19 @@
+from abc import abstractmethod
```
msaroufim, line 1: Cool! I need to sanity check this with #1857 specifically some optimizations my modify state but what I like about this is it makes it clear that optimizations are composable

ts/torch_handler/utils/conf/config.py
```diff
@@ -0,0 +1,63 @@
+from abc import abstractmethod
```
msaroufim, line 1: Note to self: we have yaml parsing in quite a few parts of our codebase now from `benchmarks` to the new #1492 so may make sense to unify this at some point

msaroufim (changes requested):

ts/torch_handler/utils/conf/__init__.py
```diff
@@ -0,0 +1,11 @@
+from .config import CONFIGURATIONS
+from os.path import dirname, basename, isfile, join
+import glob
+
```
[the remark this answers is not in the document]
min-jean-cho: The design is:

- all custom config files end in `config.py` (e.g., `ipex_config.py`, `custom_config.py`)
- all custom config classes in `config.py` end in `Conf` (e.g., `class IPEXConf`, `class CustomConf`)

The goal of this file is:
-  [`__import__`](https://github.com/pytorch/serve/pull/1664/files/13f23bd56a04ccf24f98396d08f8328a14db8651#diff-62d2a394cc3debb3b00a3b12890c68cd7aca25b90754d953c360682c3f05ce02R9) all custom config files  
- upon import they will be registered via [@configuration_registry](https://github.com/pytorch/serve/pull/1664/files/13f23bd56a04ccf24f98396d08f8328a14db8651#diff-5a0d965b22924e8389c28c797c7d3e043f63e385cfd6d8d59a7757d16c0582d1R75), which will be added to the [`CONFIGURATIONS`](https://github.com/pytorch/serve/pull/1664/files/13f23bd56a04ccf24f98396d08f8328a14db8651#diff-19f79b809f25fea8788bd9cfc1ec8e69c5ad0357707bf5d2d08c421dec823d71R13) dictionary whose key and name are class name and object respectively
- [`__all__`](https://github.com/pytorch/serve/pull/1664/files/13f23bd56a04ccf24f98396d08f8328a14db8651#diff-62d2a394cc3debb3b00a3b12890c68cd7aca25b90754d953c360682c3f05ce02R11) the custom `Conf` in `CONFIGURATIONS `will be exported upon `from ts.torch_handler.utils.conf import CONFIGURATIONS` in the base_handler. 
- the specific `IPEXConf` is accessed via `CONFIGURATIONS["ipex"]`

ts/torch_handler/utils/optimization/__init__.py
```diff
@@ -0,0 +1,11 @@
+from .optimization import OPTIMIZATIONS
```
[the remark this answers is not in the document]
min-jean-cho: Similarly as above:

- all custom optimization files end in `optimization.py` (e.g., `ipex_optimization.py`, `custom_optimzation.py`)
- all custom optimization classes in `optimization.py` end in `Optimization` (e.g., `class IPEXOptimization`, `class CustomOptimization`)

ts/torch_handler/utils/conf/ipex_config.py
```diff
@@ -0,0 +1,82 @@
...
+    assert data in QUANTIZATION_APPROACHES, "quantization approach {} is NOT supported".format(data)
+    return data 
+
+def _load_file_path(data):
+    if isinstance(data, str):
+        assert Path(data).exists(), "{} does not exist".format(data)
+        data = torch.load(data)
+    return data
+    
+quantization_schema = Schema({
```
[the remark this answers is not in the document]
min-jean-cho: Sure, I'm good with either option -- let me know if we would like to stick to `@dataclass`  for a unified config yaml file parsing

ts/torch_handler/utils/conf/__init__.py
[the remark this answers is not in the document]
msaroufim: Thank you I like this, might be worth adding a few coments directly in the code to explain how this works for someone new

Basically something like: "Hey I see you're changing this file, here's what you need to do to add a new optimization"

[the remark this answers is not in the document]
min-jean-cho: Sounds good -- there may be some few minor changes after switching to `@dataclass`.

ts/torch_handler/utils/conf/ipex_config.py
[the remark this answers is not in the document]
min-jean-cho: @msaroufim, have updated to `@dataclass ` , please have a look.

ts/torch_handler/utils/conf/__init__.py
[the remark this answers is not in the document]
min-jean-cho: @msaroufim , have added comments in the code throughout, please have a look.

ts/torch_handler/utils/optimization/ipex_optimization.py
```diff
@@ -0,0 +1,100 @@
...
+                else: # script 
+                    try:
+                        model = torch.jit.script(model)
+                    except:
+                        logger.error("TorchScript scripting the model failed. Make sure the model is scriptable.")
+                        exit(-1)
+                    
+                model = torch.jit.freeze(model)
+        
+        return model
```
min-jean-cho, line 100: cc. @jgong5, @jingxu10

ts/torch_handler/utils/optimization/ipex_optimization.py
```diff
@@ -0,0 +1,100 @@
...
+                        model = torch.jit.trace(model, example_inputs=self.torchscript_example_inputs)
+                    except:
+                        try: 
+                            model = torch.jit.trace(model, example_inputs=self.torchscript_example_inputs, check_trace=False, strict=False)
+                        except:
+                            logger.error("TorchScript tracing the model failed. Make sure the model is traceable.")
+                            exit(-1)
+                else: # script 
+                    try:
+                        model = torch.jit.script(model)
```
jgong5, line 93: Scripting mode with bf16 autocast is not supported on CPU yet. Also for int8, we only support tracing too. Is it OK to only support jit tracing in all cases?

jgong5 (commented):
Do we have UTs to cover various features added in this PR, e.g., inference with fp32/bf16 and int8 quantization?

ts/torch_handler/utils/optimization/ipex_optimization.py
```diff
@@ -0,0 +1,100 @@
...
+        # channel last 
+        if self.channels_last:
+            model = model.to(memory_format=torch.channels_last)
+        
+        # dtype 
+        if self.dtype == 'float32':
+            model = ipex.optimize(model, dtype=torch.float32)
+        elif self.dtype == 'bfloat16':
+            model = ipex.optimize(model, dtype=torch.bfloat16)
+        else: # int8
```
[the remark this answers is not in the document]
min-jean-cho: thanks @jgong5, have added a check to `quantization` mandatory for INT8, and not applicable for other dtypes (FP32, BF16). https://github.com/pytorch/serve/pull/1664/files#diff-5a0d965b22924e8389c28c797c7d3e043f63e385cfd6d8d59a7757d16c0582d1R89

ts/torch_handler/utils/optimization/ipex_optimization.py
min-jean-cho: thanks @jgong5, have only supported torchscript trace

min-jean-cho:
> Do we have UTs to cover various features added in this PR, e.g., inference with fp32/bf16 and int8 quantization?

@msaroufim , may I add IPEX-related UTs under this directory: https://github.com/pytorch/serve/tree/master/ts/torch_handler/unit_tests

msaroufim:
We can add unit tests to `test/pytest`. Also after chatting with @HamidShojanazeri offline we should split out optimizations into 2 different discussions
1. Optimizations done after training but before inference 
2. Optimizations during inference via a runtime

So for example 1 would be any optimization that works after you run `torch.save()` and 2 would be something like 
 `onnxruntime.InferenceSession(model_onnx_path, extra_args)`

The reason we need to make this distinction is optimizations in the style of 1 should not be happening in a handler, they should happen either as part of a seperate pre-inference step like https://github.com/msaroufim/torchprep or as an example with documentation. But optimisations in 2 have no choice but to be in a handler. Otherwise the most extreme case would be trying to get good out of box performance with some optimizations by having a full training loop happen in `_initialize()` and that just feels wrong

min-jean-cho:
> We can add unit tests to `test/pytest`. Also after chatting with @HamidShojanazeri offline we should split out optimizations into 2 different discussions
> 
> 1. Optimizations done after training but before inference
> 2. Optimizations during inference via a runtime
> 
> So for example 1 would be any optimization that works after you run `torch.save()` and 2 would be something like `onnxruntime.InferenceSession(model_onnx_path, extra_args)`
> 
> The reason we need to make this distinction is optimizations in the style of 1 should not be happening in a handler, they should happen either as part of a seperate pre-inference step like https://github.com/msaroufim/torchprep or as an example with documentation. But optimisations in 2 have no choice but to be in a handler. Otherwise the most extreme case would be trying to get good out of box performance with some optimizations by having a full training loop happen in `initialize()` and that just feels wrong

Thanks @msaroufim -- makes sense. I see most IPEX optimizations fit into (1). I see bf16 amp context manager `torch.cpu.amp.autocast` is necessary for (2) like here https://github.com/pytorch/serve/pull/1664/files#diff-d30e1f5ef9fd05e2ab9f652c116461632586cc132dde5d23ac4bc5cd19c799ccR196. cc. @jgong5  if any comments on this.

jgong5:
> Thanks @msaroufim -- makes sense. I see most IPEX optimizations fit into (1). I see bf16 amp context manager `torch.cpu.amp.autocast` is necessary for (2) like here https://github.com/pytorch/serve/pull/1664/files#diff-d30e1f5ef9fd05e2ab9f652c116461632586cc132dde5d23ac4bc5cd19c799ccR196. cc. @jgong5 if any comments on this.

I agree to categorize the optimizations into two parts. The (2) should be much lighter to apply than (1). But I think something different w.r.t. IPEX optimizations. Most IPEX optimizations fit (2) except for int8 calibration which would take longer time and fit (1) more. More details:
1. For FP32 and BF16, the extra steps are `ipex.optimize` and JIT tracing which are light.
2. For INT8, the extra steps are calibration, quantization and JIT tracing. The "calibration" part is a bit heavy while "quantization" is more like `ipex.optimize` and light-weight. But we can do the calibration step offline and save the int8 recipe as a json file to be loaded at runtime.

With this, can we still keep FP32/BF16 optimization in (2) and move int8 calibration step to (1)? We may still leave an option for users to do calibration at runtime if they don't want to do an offline preparation step, for the sake of flexible usage.

A separate question: does TorchServe support a JIT scripted module as input? In that case, the IPEX runtime optimization steps can be avoided for all data types.

ts/torch_handler/utils/optimization/ipex_optimization.py
```diff
@@ -0,0 +1,88 @@
...
+            from intel_extension_for_pytorch.quantization import prepare, convert
+            
+            if self.quantization_approach == 'static':
```
HamidShojanazeri, line 56: Thanks @min-jean-cho for the updates, I believe including quantization or tracing on the fly as part of serving may raise some concerns. For example quantization may result in accuracy drop that would be great if user can provide the quantized model to make sure we are not causing the drop on the fly without having the chance to check. 

Similarly tracing the model may fail as well for different sort of reasons. 

It might be safer if we leave the optimizations that changes the models weight the (the .pt)/ kernel fusions as an example as its been [here](https://github.com/pytorch/serve/tree/master/examples/intel_extension_for_pytorch#creating-and-exporting-int8-model-for-ipex) or in a separate optimization folder with automated scripts offerings that we could extend in future with other optimizations as well.

Then we can keep anything that does not apply changes to model weights/ kernels (fusions) such as core pinning and similar optimizations as part of the base_handler.

This would also let us keep the config files focused on runtime settings and not mixing it with non-runtime optimizations.

lxning:
@min-jean-cho Thank you for the contribution. Could you please add documentation and test log in this PR?

ts/torch_handler/utils/optimization/ipex_optimization.py
```diff
@@ -0,0 +1,46 @@
...
+    Args:
+        cfg (Conf): the optimization configuration.
+    """
+    def __init__(self, cfg):
+        super().__init__(cfg)
+        
+        self.dtype = cfg.dtype
+        self.channels_last = cfg.channels_last
+
+    def optimize(self, model):
```
min-jean-cho, line 23: @msaroufim, for FP32/BF16 eager, as  @jgong5 has suggested, I've kept  them as runtime optimizations as they are light-weight. For INT8 and JIT that include quantization and tracing, as @HamidShojanazeri has suggested, I'll keep them in a documentation. Are we all aligned on this?

min-jean-cho:
> @min-jean-cho Thank you for the contribution. Could you please add documentation and test log in this PR?

Thanks @lxning , documentation refers to an update to https://github.com/pytorch/serve/blob/master/examples/intel_extension_for_pytorch/README.md and/or new documentations to how to add custom config & optimization ? And test log refers to torchserve logs `./logs` ?

msaroufim:
closing for now, we can revisit this kind of refactor later

PR #1677, opened 2022-06-09
Support array of parameters

### LINKED ISSUE #1536: Process several images in one request
## Context

Cant send several images in a single request

first request
```
$ curl http://localhost:8080/predictions/resnet-18 -F 'data=@docs/images/dogs-before.jpg' -F 'data=@docs/images/kitten_small.jpg'
{
  "tabby": 0.7557169795036316,
  "tiger_cat": 0.14208266139030457,
  "Egyptian_cat": 0.059956926852464676,
  "lynx": 0.0372869074344635,
  "tiger": 0.0014757852768525481
```
second request

```
$curl http://localhost:8080/predictions/resnet-18 -F 'data=@docs/images/kitten_small.jpg' -F 'data=@docs/images/dogs-before.jpg'
{
  "toy_terrier": 0.09392645955085754,
  "bluetick": 0.07842477411031723,
  "Boston_bull": 0.05028736591339111,
  "kelpie": 0.043856214731931686,
  "German_short-haired_pointer": 0.043389711529016495
}
```
But I receive only one output for any request.
I thought that may be there are issues with batching, but batch_size =4, delay 5000, so after request I receive the json in 5 seconds

* torchserve version: master commit 6a191f822f2e759c2aad07b8f9eacbedcb039827
* torch-model-archiver version: 0.5.3
* torch version: 1.9.0+cu111
* torchvision version [if any]: 0.10.0+cu111
* torchtext version [if any]:  0.10.0
* torchaudio version [if any]: 0.9.0
* java version: openjdk 11.0.14 2022-01-18
* Operating System and version: openjdk 11.0.14 2022-01-18 

## Your Environment
pull fom master and run resnet-18 example 

## Expected Behavior
Expect to receive predictions for several images

## Current Behavior
It returns only one prediction

## Possible Solution
I found a unit test, and it works https://github.com/pytorch/serve/blob/master/ts/torch_handler/unit_tests/test_image_classifier.py#L30
So I don't know the solution. May be my expectations are wrong and it is correct behavior

jamesanto:
## Description

Supports array of parameters. It might be useful in cases where the model might expect a list of images, instead of a single image. It could also be used to control batching manually.

Fixes #1536 (Potentially/Partially ?)

## Type of change

Technically it doesn't break any API/interface. The service would behave differently if someone uses array parameters in requests.

Added unit test validating the expected request formats.

## Checklist:

- [x] Did you have fun?

### DIFF
--- a/ts/service.py
+++ b/ts/service.py
@@ -64,8 +64,25 @@ class Service(object):
             model_in = dict()
             # Parameter level headers are updated here. multipart/form-data can have multiple headers.
             for parameter in parameters:
-                model_in.update({parameter["name"]: parameter["value"]})
-                model_in_headers.update({parameter["name"]: {"content-type": parameter["contentType"]}})
+                parameter_name = parameter["name"]
+                if parameter_name.endswith("[]"):
+                    parameter_name = parameter_name[:len(parameter_name) - 2]
+                if parameter_name in model_in:
+                    parameter_value = model_in[parameter_name]
+                    if not isinstance(parameter_value, list):
+                        parameter_value = [parameter_value]
+                    parameter_value.append(parameter["value"])
+
+                    content_types = model_in_headers[parameter_name]['content-type']
+                    if not isinstance(content_types, list):
+                        content_types = [content_types]
+                    content_types.append(parameter["contentType"])
+                    header_value = {"content-type": content_types}
+                else:
+                    parameter_value = parameter["value"]
+                    header_value = {"content-type": parameter["contentType"]}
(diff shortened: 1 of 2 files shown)


msaroufim:
So I may need some more time to review and think about this one. Overall I think this makes sense to support video and time series use cases. As far as controlling the batch size though this probably isn't the right use case since we already support dynamic batching. Would you be interested in working on a video or time series example to make the case for this change?

msaroufim:
I believe per @lxning's comment this PR is not needed but feel free to reopen if you'd like to revisit

Issue #987, opened 2021-03-02
Built-in Handler for regression


Ark-kun:
## Is your feature request related to a problem? Please describe.

Every model ultimately transforms an input tensor (array of numbers) to an output tensor (another array of numbers). This is the general case, yet there is no handler that could handle such a model.

## Describe the solution

The base handler or a some new regression_handler should handle the case of predicting from a JSON list of numbers to a JSON list of numbers.

msaroufim:
@Ark-kun this is a a great idea - would you like to go ahead and make the PR for it?

PR #1682, opened 2022-06-13
feat(model-archiver): Enabled the built-in base handler

Ark-kun:
## Description

This handler is helpful for structural data and raw number arrays.
When serving the model, the base handler should be used with an appropriate service envelope ("body", "json" etc).

Fixes https://github.com/pytorch/serve/issues/987

## Type of change

## Checklist:

- [x] Did you have fun?

### DIFF
--- a/model-archiver/README.md
+++ b/model-archiver/README.md
@@ -139,6 +139,7 @@ A serialized file (.pt or .pth) should be a checkpoint in case of torchscript an
 ### Handler
 
 Handler can be TorchServe's inbuilt handler name or path to a py file to handle custom TorchServe inference logic. TorchServe supports the following handlers out of box:
+0. `base_handler`
 1. `image_classifier`
 2. `object_detector`
 3. `text_classifier`
--- a/model-archiver/model_archiver/model_packaging_utils.py
+++ b/model-archiver/model_archiver/model_packaging_utils.py
@@ -25,6 +25,7 @@ archiving_options = {
 
 
 model_handlers = {
+        'base_handler': 'tables',
         'text_classifier': 'text',
         'image_classifier': 'vision',
         'object_detector': 'vision',
--- a/docs/custom_service.md
+++ b/docs/custom_service.md
@@ -331,6 +331,7 @@ The above ModelHandler class should have the following methods with captum funct
 #### Extend default handlers
 
 TorchServe has following default handlers.
+- [base_handler](https://github.com/pytorch/serve/blob/master/ts/torch_handler/base_handler.py)
 - [image_classifier](https://github.com/pytorch/serve/blob/master/ts/torch_handler/image_classifier.py)
 - [image_segmenter](https://github.com/pytorch/serve/blob/master/ts/torch_handler/image_segmenter.py)
 - [object_detector](https://github.com/pytorch/serve/blob/master/ts/torch_handler/object_detector.py)


model-archiver/README.md
```diff
@@ -139,6 +139,7 @@ A serialized file (.pt or .pth) should be a checkpoint in case of torchscript an
 ### Handler
 
 Handler can be TorchServe's inbuilt handler name or path to a py file to handle custom TorchServe inference logic. TorchServe supports the following handlers out of box:
+0. `base_handler`
```
msaroufim, line 142: I don't believe this renders correctly a list, just start from 1

model-archiver/model_archiver/model_packaging_utils.py
```diff
@@ -25,6 +25,7 @@
 
 
 model_handlers = {
+        'base_handler': 'tables',
```
msaroufim, line 28: was this change tested? What's the intent?

msaroufim (changes requested):

lxning:
base handler can be either inherited by customized handler, or directly input as handler in model archiver.

model-archiver/README.md
Ark-kun: It seems to render correctly on GitHub: https://github.com/pytorch/serve/blob/de0868f53b36782f78da6491c3e77603b0a57ff6/docs/custom_service.md#extend-default-handlers

But I can change it to `1.` if needed.

model-archiver/model_archiver/model_packaging_utils.py
Ark-kun: There is only one place in the code where `model_handlers` is used.

The handler can either be a built-in handler (not copied to the model archive) or must be a python file (copied to the model archive). The behavior is decided based on the presence of the handler name in that dictionary.

It's only possible to use the base handler as a built-in handler. (Using the `base_handler.py` file does not work since that file uses relative imports that break when the file if copied to the model archive).

So, to make it possible to use the base_handler in model archive creator is to add it to the it to the `model_handlers` dict.

Ark-kun:
> base handler can be ... directly input as handler in model archiver.

I'm not sure that's the case:

The handler can either be a built-in handler (not copied to the model archive) or must be a python file (copied to the model archive). The behavior is decided based on the presence of the handler name in that dictionary.

It's only possible to use the base handler as a built-in handler. (Using the base_handler.py file does not work since that file uses relative imports that break when the file if copied to the model archive).

So, to make it possible to use the base_handler in model archive creator is to add it to the it to the model_handlers dict.

model-archiver/model_archiver/model_packaging_utils.py
msaroufim: Hmm I've been wanting to remove all relative imports in our handler so maybe this a good excuse to do so

Ark-kun: I think this PR would still be beneficial.
It's much easier to specify a built-in handler name vs putting a file inside a container.
It's really surprising to me that TorchServe supports serving advanced models out of the box (images, text), but does not support the basics (tabular models).

What concerns do you have about adding the `base_handler` to that dictionary?

msaroufim:
Hi @Ark-kun I apologize for my pontification, in hindsight this was a correct idea and I did end up implementing it when I needed to add ONNX support here https://github.com/pytorch/serve/blob/master/model-archiver/model_archiver/model_packaging_utils.py#L22 - closing for now but please let me know if you have any other feedback, will merge it much more quickly this time

PR #1691, opened 2022-06-17
Scriptable Tokenizer for Text Classification Example

mreso:
…ptable tokenizer into a single artifact

## Description

This example shows how to combine a model with a scriptable tokenizer and script the two together with TorchScript to achieve a single artifact.
The combination of model and tokenizer makes sure that the same tokenizer is used in training and inference.
Additionally, it decreases handler complexity and makes it easy to deploy these in the future C++ based backend.

Training the model depends on torchdata which is mentioned in the README. Inference does not require the package.

Fixes #(issue)

## Type of change

Test plan:
- Execute commands in examples/text_classification_with_scriptable_tokenzier/README.md
- As well as examples/text_classification/README.md (text_handler.py was slightly modified)

## Checklist:

- [X] Did you have fun?


examples/text_classification_with_scriptable_tokenzier/handler.py
```diff
@@ -0,0 +1,107 @@
...
+logger = logging.getLogger(__name__)
+
+def remove_html_tags(text):
+    """
+    Removes html tags
+    """
+    clean_text = CLEANUP_REGEX.sub("", text)
+    return clean_text
+
+class TextClassifier(BaseHandler, ABC):
```
msaroufim, line 26: Was the `text_handler` not the best option for you? Could you elaborate a bit more on why?

examples/text_classification_with_scriptable_tokenzier/train_model.py
```diff
@@ -0,0 +1,177 @@
...
+        loss, accuracy = evaluate()
+        print(f"Epoch = [{epoch}], loss = [{loss}], accuracy = [{accuracy}]")
+
+
+train()
+
+
+class TokenizerModelAdapter(nn.Module):
+    """
+    TokenizerModelAdapter moves input onto device and adds batch dimension
```
msaroufim, line 142: I thought line 2 said `DOES NOT SUPPORT BATCH!` could you elaborate a bit more?

examples/text_classification_with_scriptable_tokenzier/README.md
```diff
@@ -0,0 +1,50 @@
...
+    torch-model-archiver --model-name scriptable_tokenizer --version 1.0 --serialized-file model.pt --handler handler.py --extra-files "index_to_name.json"
+    ```
+       
+ * Register the model on TorchServe using the above model archive file and run a classification
+   
+    ```bash
+    mkdir model_store
+    mv scriptable_tokenizer.mar model_store/
+    torchserve --start --model-store model_store --models my_tc=scriptable_tokenizer.mar
+    curl http://127.0.0.1:8080/predictions/my_tc -T sample_text.txt
```
msaroufim, line 49: put the expected output as well

examples/text_classification_with_scriptable_tokenzier/README.md
```diff
@@ -0,0 +1,50 @@
+# Text Classfication using a Scriptable Tokenizer 
+
+This example shows how to combine a text classification model with a scriptable tokenizer into a single artifact and deploy it with TorchServe.
+The combination of model and tokenizer into a single artifact reduces the handler complexity and ensures that the same tokenizer is used during training and inference
```
msaroufim, line 4: >  into a single artifact reduces the handler complexity

are you referring to not needing to pass a tokenizer in extra files? 

> ensures that the same tokenizer is used during training and inference 

Could you elaborate a bit more on this, I'm not familiar with this general problem

examples/text_classification_with_scriptable_tokenzier/handler.py
```diff
@@ -0,0 +1,107 @@
...
+        text = line.get("data") or line.get("body")
+        if isinstance(text, (bytes, bytearray)):
+            text = text.decode('utf-8')
+
+        text = remove_html_tags(text)
+        text = text.lower()
+
+        return text
+
+    def inference(self, data, *args, **kwargs):
```
msaroufim, line 61: looks like optional args are not used?

examples/text_classification_with_scriptable_tokenzier/train_model.py
```diff
@@ -0,0 +1,177 @@
...
+DEVICE = torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
+
+PADDING_IDX = 1
+BOS_IDX = 0
+EOS_IDX = 2
+MAX_SEQ_LEN = 256
+XLMR_VOCAB_PATH = r"https://download.pytorch.org/models/text/xlmr.vocab.pt"
+XLMR_SPM_MODEL_PATH = r"https://download.pytorch.org/models/text/xlmr.sentencepiece.bpe.model"
+
+text_transform = T.Sequential(
```
msaroufim, line 28: 1-2 line comment would be helpful here

examples/text_classification_with_scriptable_tokenzier/train_model.py
```diff
@@ -0,0 +1,177 @@
...
+        Returns:
+            (Tensor): On device text tensor with batch dimension
+        """
+        tokens = F.to_tensor(tokens, padding_value=self._padding_idx).to(self._dummy_param.device)
+        if len(tokens.shape) == 2:
+            return tokens
+        else:
+            return tokens.unsqueeze(0)
+
+combi_model = T.Sequential(
```
msaroufim, line 165: 1-2 line comment would be helpful here

examples/text_classification_with_scriptable_tokenzier/train_model.py
```diff
@@ -0,0 +1,177 @@
...
+    text_transform,
+    TokenizerModelAdapter(PADDING_IDX),
+    model,
+)
+
+combi_model.eval()
+
+combi_model.to("cpu")
```
msaroufim, line 173: Is the training happening on GPU if available but then model scripted on CPU?

msaroufim (changes requested):
Thanks Matthias, super glad to see this work in. Left a bunch of feedback. The important ones is we need some test, updates to the README to clarify benefits to users and a few more comments in the train.py

examples/text_classification_with_scriptable_tokenzier/train_model.py
```diff
@@ -0,0 +1,177 @@
...
+class TokenizerModelAdapter(nn.Module):
+    """
+    TokenizerModelAdapter moves input onto device and adds batch dimension
+    """
+    def __init__(self, padding_idx):
+        super().__init__()
+        self._padding_idx = padding_idx
+        self._dummy_param = nn.Parameter(torch.empty(0))
+
+    def forward(self, tokens: Any) -> torch.Tensor:
```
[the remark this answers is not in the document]
mreso: This Any was necessary due to the scripting which insisted that the tokenizer return Any type. I will try to revisit that to make it narrower, but it might not be possible. In this case I will add a comment.

examples/text_classification_with_scriptable_tokenzier/train_model.py
```diff
@@ -0,0 +1,177 @@
...
+classifier_head = RobertaClassificationHead(num_classes=NUM_CLASSES, input_dim=INPUT_DIM)
+model = XLMR_BASE_ENCODER.get_model(head=classifier_head)
+model.to(DEVICE)
+
+LEARNING_RATE = 1e-5
+optim = AdamW(model.parameters(), lr=LEARNING_RATE)
+criteria = nn.CrossEntropyLoss()
+
+
+def train_step(input_tensor, target):
```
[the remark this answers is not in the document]
mreso: Same as above, took the text_classification example as a blueprint. I will need to upstream some changes to `text` to make this work.

examples/text_classification_with_scriptable_tokenzier/handler.py
mreso: The `text_handler` forces me to load a [source_vocab](https://github.com/pytorch/serve/blob/a4d5090e114cdbeddf5077a817a8cd02d129159e/ts/torch_handler/text_handler.py#L53) and [loads a tokenizer](https://github.com/pytorch/serve/blob/a4d5090e114cdbeddf5077a817a8cd02d129159e/ts/torch_handler/text_handler.py#L37) which is not necessary in my case.

examples/text_classification_with_scriptable_tokenzier/train_model.py
mreso: Sure, I will add a comment here. The model expects a batch dimension, or it fails with not matching tensor shaped. Is there a better way to solve this?

examples/text_classification_with_scriptable_tokenzier/train_model.py
mreso: Yes, for scripting the model should be moved to cpu. Will put a comment here.

examples/text_classification_with_scriptable_tokenzier/handler.py
mreso: These come from the BaseHandler. Removing them results in a linter error as the overriding method has a different signature. Any advice? Could not find a compelling solution other than disable one of the linter errors (unused variable or signature change).

examples/text_classification_with_scriptable_tokenzier/train_model.py
msaroufim: Yeah AFAIK only functorch makes it easy to have batch agnostic models so feel free to ignore my original point here

mreso:
Just adding the logs for running the unit test manually:
image

examples/dcgan_fashiongen/create_mar.sh
```diff
@@ -16,7 +16,7 @@ function cleanup {
 trap cleanup EXIT
 
 # Download and Extract model's source code
-sudo apt-get install zip unzip
+sudo apt-get install -y zip unzip
```
msaroufim, line 19: is this change needed?

msaroufim (approved):
LGTM - although I think we should revert the changes to `test/pytest` and `create_mar.sh` let's fix those properly in another PR pending review in the `CONTRIBUTING-dev.md` spec I shared

examples/dcgan_fashiongen/create_mar.sh
mreso: > I think we should revert the changes to test/pytest and create_mar.sh let's fix those properly in another PR
Sounds good, reverted the changes

lxning:
@mreso could you run sanity and regression test to make sure the pytest you added is verified?

mreso:
Thank @lxning! Here are the logs for the regression tests: https://gist.github.com/mreso/0bcaed895517f74a09cbd4e80621969c
Let me know if you need anything else!

PR #1709, opened 2022-06-24
Fixed broken path joins and unclosed files

### LINKED ISSUE #1686: Error in loading model in Windows os
I've noticed that there is an error in the code:
https://github.com/pytorch/serve/blob/a4d5090e114cdbeddf5077a817a8cd02d129159e/ts/model_loader.py#L92

There is an attempt to read the `MANIFEST.json` file, and there is use of the `/` charecter which used as a seperator in the Linux os, and in Windows there is use of the back-slash `\`, so this line creates always an invalid path so the loader does not find the manifest file.

I would change it to use the function `os.path.join` (which already in use) to write something like this:
```python
manifest_file = os.path.join(model_dir, "MAR-INF", "MANIFEST.json") 
```

I'm not changing this by myself in a PR beacause I don't know if there is another lines in the code which this error occurs.

### LINKED ISSUE #1708: Unclosed file descriptors in tests
Hey everyone!

As I reviewing the code as part of PR I'm working on, I found lines in the `pytest` tests that using the `open()` function without a `with`,  and the file never closed.

Example for such line: https://github.com/pytorch/serve/blob/873edc50cc6059c47830a5d30bbbdbe1e1a3d9bb/test/pytest/test_handler.py#L41

DPeled:
## Description

After a bug occurred to me while running over Windows, I found out that there are a lot places in the code where there is a path join using string concatenation and not with `os.path.join` which creates some broken paths and results some errors while serving models. In my ways of changing it, I've noticed that there are some bugs of unclosed file descriptors that were just left open, so I clodes them too.

Fixes #1708, fixes #1686 

## Type of change

## Checklist:

- [X] Did you have fun?


test/pytest/test_gRPC_inference_api.py
```diff
@@ -7,7 +7,7 @@
 import test_utils
 
 
-inference_data_json = "/../postman/inference_data.json"
+inference_data_json = "../postman/inference_data.json"
```
msaroufim, line 10: Might as well fix this too?

test/pytest/test_gRPC_management_apis.py
```diff
@@ -6,7 +6,7 @@
 from urllib import parse
 
 
-management_data_json = "/../postman/management_data.json"
+management_data_json = "../postman/management_data.json"
```
msaroufim, line 9: We can fix this too

test/pytest/test_metrics_kf.py
```diff
@@ -11,13 +11,14 @@
 import test_utils
 
 NUM_STARTUP_CFG = 0
-REPO_ROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)), "../../")
-snapshot_file = os.path.join(REPO_ROOT,"test/config_kf.properties")
+REPO_ROOT = os.path.normpath(os.path.join(os.path.dirname(os.path.abspath(__file__)), "../../"))
```
msaroufim, line 14: We can fix this too, maybe break apart into more than one line since it's getting hard to read

test/pytest/test_pytorch_profiler.py
```diff
@@ -15,15 +15,15 @@
 import test_utils
 from concurrent import futures
 
-REPO_ROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)), "../../")
-data_file_mnist = os.path.join(REPO_ROOT, "examples/image_classifier/mnist/test_data/1.png")
+REPO_ROOT = os.path.normpath(os.path.join(os.path.dirname(os.path.abspath(__file__)), "../../"))
```
msaroufim, line 18: We can improve this line too

test/pytest/test_utils.py
```diff
@@ -79,16 +79,16 @@ def unregister_model(model_name):
...
 
 def delete_mar_file_from_model_store(model_store=None, model_mar=None):
-    model_store = model_store if (model_store is not None) else f"{ROOT_DIR}/model_store/"
+    model_store = model_store if (model_store is not None) else os.path.join(ROOT_DIR, "model_store")
     if model_mar is not None:
         for f in glob.glob(path.join(model_store, model_mar + "*")):
             os.remove(f)
 
-environment_json = "/../postman/environment.json"
+environment_json = "../postman/environment.json"
```
msaroufim, line 87: Should we fix this?

msaroufim (changes requested):
Very cool thank you for improving the Torchserve user experience on Windows. I still see a few stragglers left. Might as well fix those as well and we should be able to easily merge this

test/pytest/test_gRPC_inference_api.py
lxning, line 10: Here also needs to be changed.

test/pytest/test_handler.py
```diff
@@ -7,12 +7,12 @@
 import ast 
 import pytest
 import torch
-REPO_ROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)), "../../")
-snapshot_file_kf = os.path.join(REPO_ROOT,"test/config_kf.properties")
-snapshot_file_tf = os.path.join(REPO_ROOT,"test/config_ts.properties")
-data_file_mnist = os.path.join(REPO_ROOT, 'examples/image_classifier/mnist/test_data/1.png')
-input_json_mnist = os.path.join(REPO_ROOT, "kubernetes/kserve/kf_request_json/v1/mnist.json")
-input_json_mmf = os.path.join(REPO_ROOT, "examples/MMF-activity-recognition/372CC.info.json")
+REPO_ROOT = os.path.normpath(os.path.join(os.path.dirname(os.path.abspath(__file__)), "../../"))
```
lxning, line 10: Same here.

msaroufim:
Hi @DPeled are you still interested in finishing this PR? Just a few minor comments left and we can merge

DPeled:
@msaroufim I'll work on it soon, I didn't have much time at the last weeks

test/pytest/test_gRPC_inference_api.py
DPeled: It works fine with the join it appears in later, as I tried with another example in python `os.join`

DPeled: it works well in joins, the output out put may seems weird, but it still work over Windows.
It would look like `Path\To\dir/../postman/infer.json` which works somehow with the simple `open` function of python

test/pytest/test_metrics_kf.py
DPeled:
```suggestion
REPO_ROOT = os.path.normpath(
    os.path.join(os.path.dirname(os.path.abspath(__file__)), "../../")
)
```
Somthing like that? I don't see any other way

test/pytest/test_pytorch_profiler.py
DPeled:
```suggestion
REPO_ROOT = os.path.normpath(
    os.path.join(os.path.dirname(os.path.abspath(__file__)), "../../")
)
```

Again, I can afford only that to stay readable and pythonic as well

test/pytest/test_utils.py
DPeled: The stating '/' should be removed as it can be messing with the split occurring in line 91.
Any other change to that line I don't see how it will be necessary.

DPeled:
@msaroufim @lxning I've read your review and I replied as well.
Please take a look hopefully you'll understand my answers and why I answered it.

msaroufim (approved):

test/pytest/test_metrics_kf.py
```diff
@@ -11,13 +11,16 @@
 import test_utils
 
 NUM_STARTUP_CFG = 0
-REPO_ROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)), "../../")
-snapshot_file = os.path.join(REPO_ROOT,"test/config_kf.properties")
+REPO_ROOT = os.path.normpath(
+    os.path.join(os.path.dirname(os.path.abspath(__file__)), "../../")
```
mreso, line 15: @DPeled Hi, I've seen the resolved conversation about this topic above but could you elaborate why e.g. [this line](https://github.com/pytorch/serve/blob/3c02fb67e12c82cb3231226a3a947ed93b519f4f/examples/text_to_speech_synthesizer/create_mar.py#L16) needs to be changes  from join('A','B/C') into join('A','B','C') but here we do not need to to join('A', '..','..')

mreso (approved):
LGTM
@msaroufim Should we suggest usage of pathlib instead of os.path for future developments?

msaroufim:
Probably a good boot camp task template we can setup @mreso

test/pytest/test_metrics_kf.py
DPeled: Of course!

Apparently the underlying code where the real accessing to that path is happened, Windows can understand that type of paths (like `C:\PATH\TO\SUBDIR\../../` which the join produce).

But, moreover, to eliminate that strange looking path, the `os.path.normpath` function takes this path and normalize it to look just like any traditional path in the current OS.

agunapal (approved):
Looks good

DPeled:
@msaroufim Hi is there something to do with the tasks that failed?
From the logs I can see that because some packages not installed or `torchserver.exe` was not recognized in windows.
It doesn't seems to fail as a result from my PR

msaroufim:
I kicked off CI again let's see what happens

msaroufim:
@DPeled can you please fix the lint error with `pre-commit` instructions are the bottom of the failed action. Also the GPU check needs to pass

maaquib (approved):

Issue #1736, opened 2022-07-12
torchserve version mismatching at `pytorch/torchserve:0.5.3-gpu`, `pytorch/torchserve:0.5.3-cpu`


klae01:
`pip list` in `pytorch/torchserve:0.5.3-gpu` return following. I found `torchserve` version is `0.6.0`.
Also `pytorch/torchserve:0.5.3-cpu` also have same issue

 `pytorch/torchserve:0.5.3-gpu` 
```
Package              Version
-------------------- ------------
captum               0.5.0
certifi              2022.5.18.1
charset-normalizer   2.0.12
cycler               0.11.0
enum-compat          0.0.3
fonttools            4.33.3
future               0.18.2
idna                 3.3
kiwisolver           1.4.2
matplotlib           3.5.2
numpy                1.22.4
packaging            21.3
Pillow               9.1.1
pip                  22.1.2
pkg_resources        0.0.0
psutil               5.9.1
pyparsing            3.0.9
python-dateutil      2.8.2
requests             2.28.0
setuptools           62.3.3
six                  1.16.0
torch                1.11.0+cu102
torch-model-archiver 0.6.0
torchserve           0.6.0
torchtext            0.12.0
torchvision          0.12.0+cu102
tqdm                 4.64.0
typing_extensions    4.2.0
urllib3              1.26.9
wheel                0.37.1
```

 `pytorch/torchserve:0.5.3-cpu`
```
Package              Version
-------------------- -----------
captum               0.5.0
certifi              2022.5.18.1
charset-normalizer   2.0.12
cycler               0.11.0
enum-compat          0.0.3
fonttools            4.33.3
future               0.18.2
idna                 3.3
kiwisolver           1.4.2
matplotlib           3.5.2
numpy                1.22.4
packaging            21.3
Pillow               9.1.1
pip                  22.1.2
pkg_resources        0.0.0
psutil               5.9.1
pyparsing            3.0.9
python-dateutil      2.8.2
requests             2.28.0
setuptools           62.3.3
six                  1.16.0
torch                1.11.0+cpu
torch-model-archiver 0.6.0
torchserve           0.6.0
torchtext            0.12.0
torchvision          0.12.0+cpu
tqdm                 4.64.0
typing_extensions    4.2.0
urllib3              1.26.9
wheel                0.37.1
```

msaroufim:
Ack - I've repro'ed and fixed this issue in a staging environment just need to confirm with rest of the team before I promote

[...]

PR #1742, opened 2022-07-18
install torchserve from repository (pytorch/torchserve:0.5.3-cpu version mismatching issue)

klae01:
## Description

In the current implementation, at the time of `docker build`, a Docker image with the latest version of `torchserve` installed is created regardless of the version of the repository.

I change implementations to install `torchserve` from source.

Fixes https://github.com/pytorch/serve/issues/1736

## Type of change

Compare the expected torchserve version and torchserve version of docker images obtained using `pip list|grep "torchserve "|awk '{printf $2}'`.

## Checklist:

- [x] Did you have fun?

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -76,7 +76,7 @@ RUN python -m pip install --no-cache-dir captum torchtext
 
 COPY . tmp/serve
 
-RUN python -m pip install --no-cache-dir tmp/serve/ tmp/serve/model-archiver
+RUN python -m pip install --no-cache-dir tmp/serve/ tmp/serve/model-archiver tmp/serve/workflow-archiver
 
 # Final image for production
 FROM ${BASE_IMAGE} AS runtime-image
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -39,7 +39,8 @@ RUN --mount=type=cache,id=apt-dev,target=/var/cache/apt \
     && rm -rf /var/lib/apt/lists/* \
     && cd /tmp \
     && curl -O https://bootstrap.pypa.io/get-pip.py \
-    && python3.8 get-pip.py
+    && python3.8 get-pip.py \
+    && rm -rf /tmp/*
 
 
 RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.8 1 \ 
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -96,12 +101,12 @@ COPY --chown=model-server --from=compile-image /home/venv /home/venv
 
 ENV PATH="/home/venv/bin:$PATH"
 
-COPY dockerd-entrypoint.sh /usr/local/bin/dockerd-entrypoint.sh
+COPY docker/dockerd-entrypoint.sh /usr/local/bin/dockerd-entrypoint.sh
```
msaroufim, line 104: are you expecting users to instead run this dockerfile from `serve` instead of `docker`?

docker/Dockerfile
```diff
@@ -71,7 +72,11 @@ RUN TORCH_VER=$(curl --silent --location https://pypi.org/pypi/torch/json | pyth
     else \
         python -m pip install --no-cache-dir torch==$TORCH_VER+cpu torchvision==$TORCH_VISION_VER+cpu -f https://download.pytorch.org/whl/torch_stable.html; \
     fi
-RUN python -m pip install -U setuptools && python -m pip install --no-cache-dir captum torchtext torchserve torch-model-archiver
+RUN python -m pip install --no-cache-dir captum torchtext
+
+COPY . tmp/serve
+
+RUN python -m pip install --no-cache-dir tmp/serve/ tmp/serve/model-archiver
```
msaroufim, line 79: So in our `Dockerfile.dev` we instead do this https://github.com/pytorch/serve/blob/master/docker/Dockerfile.dev#L61-L63

Any thoughts on what's better?

msaroufim (changes requested):

docker/Dockerfile
klae01: https://github.com/klae01/serve-dockerfile-update/blob/71e962cc2cb38e74da033ce5ed3e02434cb5906c/docker/build_image.sh#L111

I change provided `context` as `serve` (not only `docker`) while docker build.
so, still docker build at `docker`

docker/Dockerfile
klae01: torchserve & model-archiver & workflow-archiver are install at following line.
https://github.com/pytorch/serve/blob/0db95ee361d7f7a758a9e39f7cbc71e8c9340a4a/docker/Dockerfile.dev#L68

https://github.com/pytorch/serve/blob/master/ts_scripts/install_from_src.py

It seems simpler to call pip directly instead of 'install_from_src.py'.
From my observation, the installation step from the source seems to override the existing package. So uninstall in `install_from_src.py` also seems unnecessary.

In my personal opinion, it seems convenient to build a different version with the latest version of `Dockerfile.dev`, but I think it's reasonable to build only with the code from the (old) branch. If there is a serious error, it seems necessary to update the `Dockerfile` by hotfix.

msaroufim (approved):

msaroufim:
Actually after looking at this more closely this is fine, we recommend our dev images for developers

PR #1746, opened 2022-07-20
Skip Flaky Java Windows Tests

msaroufim:
## Description

We have about ~14/192 tests that break constantly on Windows CI. There are probably more flaky ones we can fix in the future but at least with these fixed we can expect the Windows CI to turn green. The way to properly fix this would also involve mocking TS more so we don't actually have to run an inference.

We haven't had the bandwidth to fix the root cause for a few months so for now it may be best to skip and revisit as a better engineering task.

Not sure if this is the best way to skip tests using `testng` in `Java` so open to feedback

The CI report will clearly state which tests passed, failed or were skipped

## Type of change

- [x] After change
Validate Windows CI below

- [x] Test B
There's tons of examples of broken windows tests  - just browse https://github.com/pytorch/serve/actions/workflows/ci_cpu.yml

Some tests that fail frequently are
* https://github.com/pytorch/serve/runs/7206779451?check_suite_focus=true
* https://github.com/pytorch/serve/runs/7186923365?check_suite_focus=true
* https://github.com/pytorch/serve/runs/7349804414?check_suite_focus=true


msaroufim:
Found most of them!

Screen Shot 2022-07-20 at 2 46 32 PM

rohithkrn:
In python, I know we can implement a common decorator like `@skipIfWindows` and annotate the tests. Not sure if there's a similar functionality in Java.

msaroufim:
> In python, I know we can implement a common decorator like `@skipIfWindows` and annotate the tests. Not sure if there's a similar functionality in Java.

Oh yeah I love pytest where this kind of thing is easy

Even JUnit has a nice feature for it with `@DisabledOnOs(WINDOWS)` but this is the best I could find with `testng`

rohithkrn (approved):

frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java
```diff
@@ -1133,7 +1134,11 @@ public void testModelWithInvalidCustomPythonDependency()
     @Test(
             alwaysRun = true,
             dependsOnMethods = {"testModelWithInvalidCustomPythonDependency"})
-    public void testLoadingMemoryError() throws InterruptedException {
+    public void testLoadingMemoryError() throws Exception {
```
mreso, line 1137: throws allows multiple exception types to be reported. Is there a reason why you changed the type to "Exception" instead of "InterruptedException, SkipException"?

mreso (changes requested):
See comment

frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java
msaroufim: Thanks for the tip, yeah fixed that just now

mreso (approved):
LGTM

msaroufim:
@rohithkrn @mreso we're all green

Screen Shot 2022-07-29 at 1 53 15 PM

PR #1775, opened 2022-07-31
Deprecate neuron Dockerfile in favor in DLC

rohithkrn:
This change is to deprecate neuron dockerfile as it is currently broken. Instead, it's ideal at this to point use containers provided by [aws/deep-learning-containers](https://github.com/aws/deep-learning-containers/blob/master/available_images.md#neuron-inference-containers) that are more reliable.

I also learnt that this was created to run performance tests but it was never run/tested.

Packages that are missing in this Dockerfile
- `python3.8, python3.8-dev, python3.8-venv, python3.8-distutils`. It installs `python3-dev`
- `gnupg`

I tried installing these packages and ran the container but it did not detect neuron devices inside the container probably because of missing drivers.

Ran the docker container like below following https://awsdocs-neuron.readthedocs-hosted.com/en/latest/neuron-deploy/dlc-then-ec2-devflow.html#
`docker run -it --name ts-jdk17 --rm --shm-size=1g -p 8080:8080 -p 8081:8081 -p 8082:8082 -e "AWS_NEURON_VISIBLE_DEVICES=ALL" -v $HOME/torchserve:/workspace ts-neuron-jdk17`
 Inside the container `neuron-ls` does not return anything.


docker/Dockerfile.neuron.dev
```diff
@@ -1,13 +1,18 @@
+## THIS DOCKERFILE HAS BEEN DEPRECATED ##
```
msaroufim, line 1: I'd rather we delete this file and leverage another container to describe neuron instructions. We could make neuron an argument of either `Dockerfile` or `Dockerfile.dev`. We shouldn't expect people to read comments in a Dockerfile

lxning:
@rohithkrn please highlight deprecation message in [README.md](https://github.com/pytorch/serve/blob/master/docker/README.md)

rohithkrn:
> @rohithkrn please highlight deprecation message in [README.md](https://github.com/pytorch/serve/blob/master/docker/README.md)

@lxning Updated. Thanks

docker/README.md
```diff
@@ -1,3 +1,6 @@
+### Deprecation notice:
```
maaquib, line 1: I would suggest moving this down. This is the main docker README and might be confusing to customers who might not have any idea about neuron

rohithkrn: In my experience I have seen these kind of deprecation notices put at the top. Also provided link to the specific dockerfile to not confuse users. I don't have a strong opinion on this though. Do you suggest putting it at the end?

maaquib: Not a big deal. I was just concerned about customers being confused about neuron devices given we don't have any other documentation for it.

maaquib (approved):

msaroufim (approved):

Issue #1776, opened 2022-08-01
Handle invalid json formatted requests without killing the worker process


duk0011:
### 🚀 The feature

**Current Setup:**
Currently, the TorchServe backend worker process dies whenever it receives an invalid json formatted requests.

**Feature Requested:**
Instead of killing the backend worker process, TorchServe could handle this failure gracefully within **try** block, log the failing payload and send back any relevant error code (may be 400?). 

**Test Query:** "Blue shirt!

**Point of failure within Torchserve:**
https://github.com/pytorch/serve/blob/master/ts/protocol/otf_message_handler.py#L314

**Failure Traceback:** 
2022-07-27T15:30:41,172 [INFO ] W-9000-el1001_1.0-stdout MODEL_LOG - Backend worker process died.
2022-07-27T15:30:41,172 [INFO ] W-9000-el1001_1.0-stdout MODEL_LOG - Traceback (most recent call last):
2022-07-27T15:30:41,172 [INFO ] W-9000-el1001_1.0-stdout MODEL_LOG -   File "/home/venv/lib/python3.8/site-packages/ts/model_service_worker.py", line 210, in <module>
2022-07-27T15:30:41,172 [INFO ] W-9000-el1001_1.0-stdout MODEL_LOG -     worker.run_server()
2022-07-27T15:30:41,172 [INFO ] W-9000-el1001_1.0-stdout MODEL_LOG -   File "/home/venv/lib/python3.8/site-packages/ts/model_service_worker.py", line 181, in run_server
2022-07-27T15:30:41,172 [INFO ] W-9000-el1001_1.0-stdout MODEL_LOG -     self.handle_connection(cl_socket)
2022-07-27T15:30:41,172 [INFO ] W-9000-el1001_1.0-stdout MODEL_LOG -   File "/home/venv/lib/python3.8/site-packages/ts/model_service_worker.py", line 132, in handle_connection
2022-07-27T15:30:41,172 [INFO ] W-9000-el1001_1.0-stdout MODEL_LOG -     cmd, msg = retrieve_msg(cl_socket)
2022-07-27T15:30:41,172 [INFO ] W-9000-el1001_1.0-stdout MODEL_LOG -   File "/home/venv/lib/python3.8/site-packages/ts/protocol/otf_message_handler.py", line 36, in retrieve_msg
2022-07-27T15:30:41,172 [INFO ] W-9000-el1001_1.0-stdout MODEL_LOG -     msg = _retrieve_inference_msg(conn)
[...]

PR #1789, opened 2022-08-10
Gracefully handle decoding exceptions in python backend

msaroufim:
Fixes https://github.com/pytorch/serve/issues/1776

Test: https://colab.research.google.com/drive/1eKMiheF50YYGtPmi0WqVkeRqLk8Gbwg5?usp=sharing (would probably be better to start writing unit tests for the python backend when we have time)

The meaningful code is at the very end of the file in the `try, except` block


ts/protocol/otf_message_handler.py
```diff
@@ -310,10 +313,21 @@ def _retrieve_input_data(conn):
...
-    if content_type == "application/json" and (decode_req is None or decode_req == "true"):
-        model_input["value"] = json.loads(value.decode("utf-8"))
-    elif content_type.startswith("text") and (decode_req is None or decode_req == "true"):
-        model_input["value"] = value.decode("utf-8")
+    if content_type == "application/json" and (
+        decode_req is None or decode_req == "true"
+    ):
+        try:
+            model_input["value"] = json.loads(value.decode("utf-8"))
+        except (JSONDecodeError, UnicodeDecodeError) as e:
```
msaroufim, line 321: The `try` and `except` statements are the meaningful changes in this PR

ts/protocol/otf_message_handler.py
```diff
@@ -310,10 +312,21 @@ def _retrieve_input_data(conn):
...
+            model_input["value"] = json.loads(value.decode("utf-8"))
+        except Exception as e:
+            logging.info(f"Exception {e} thrown on value: {value}")
+    elif content_type.startswith("text") and (
+        decode_req is None or decode_req == "true"
+    ):
+        try:
+            model_input["value"] = value.decode("utf-8")
+        except Exception as e:
+            logging.info(f"Exception {e} thrown on value: {value}")
```
agunapal, line 328: If there is an error, what should value be populated as?

msaroufim: Made this a bit clearer now, I'd like to return nothing if there's any issue

msaroufim:
@rohithkrn @lxning GPU CI seems to be frozen now

ts/protocol/otf_message_handler.py
```diff
@@ -310,10 +312,26 @@ def _retrieve_input_data(conn):
...
+            return model_input
+
+        except Exception as e:
+            logging.exception(e)
+
     else:
         model_input["value"] = value
-    return model_input
+        return model_input
+    return None
```
maaquib, line 337: Not sure if returning `None` is the right approach here. The calling method `_retrieve_request` expects a `None` response when all the data from the socket has been read. In this case we would just end up sending an empty array to the handler which can be confusing for customers who did send an invalid json.
I think we should throw an exception here and have it handled up the stack so that a proper `400 Bad Request` error response can be created and returned to the customer

msaroufim: I feel like 2 is probably the easiest. What it'll do is make the handler crash instead of the worker which feels more appropriate

msaroufim: @maaquib just made the change, is thatt what you meant?

ts/protocol/otf_message_handler.py
```diff
@@ -310,10 +312,27 @@ def _retrieve_input_data(conn):
...
-        model_input["value"] = value.decode("utf-8")
+    if content_type == "application/json" and (
+        decode_req is None or decode_req == "true"
+    ):
+        try:
+            model_input["value"] = json.loads(value.decode("utf-8"))
+
+        except Exception as e:
+            model_input["value"] = value
+            logging.exception(e)
```
maaquib, line 323:
```suggestion
            logger.warning("Failed json decoding of input data: {}".format(e))
```

ts/protocol/otf_message_handler.py
```diff
@@ -310,10 +312,27 @@ def _retrieve_input_data(conn):
...
+            
+    elif content_type.startswith("text") and (
+        decode_req is None or decode_req == "true"
+    ):
+        try:
+            model_input["value"] = value.decode("utf-8")
+
+        except Exception as e:
+            model_input["value"] = value
+            logging.exception(e)
```
maaquib, line 333:
```suggestion
            logging.warning("Failed utf-8 decoding of input data. Forwarding encoded payload", exc_info=True)
```

maaquib (commented):
This might be a bit better

ts/protocol/otf_message_handler.py
maaquib: @msaroufim  Added a couple of suggestions to make the log more direct

ts/protocol/otf_message_handler.py
msaroufim: Hmm but the issue here is you won't know what the payload was. The benefit of `logging.exception()` is that it'll print a stack trace

An alternative I was thinking of was

```python
logging.warning(f"Decoding failed with error {e} on payload {value})
```

ts/protocol/otf_message_handler.py
msaroufim: hmm looks like everything broke, will take a look tmrw

maaquib (approved):

agunapal (approved):
looks good

Issue #1797, opened 2022-08-15
Efficient and Consistent Batching


msaroufim:
### 🚀 The feature

I've had this conversation offline a few times so I figured I would summarize it a single thread to explain how batching works in TorchServe today and how to make it more efficient

### How batching works today

Today our batching is controlled via 2 parameters in `config.properties`, `batch_size` and `max_batch_delay`

So let's say you set the `batch_size` to 8 and `max_batch_delay` to 10ms then every inference request will be captured in the batch and if the `max_batch_delay` is exceeded then whatever requests are in that batch will be sent over

Then from inside a handler we run code that looks like

```python
for request in requests:
  next_input = to_tensor(request)
  batch = torch.cat((batch, next_input), 0)
  batch.to("cuda")
```

### Problems with current approach

1. We first instantiate tensors on CPU and then move to GPU which means our bottleneck will be memory bandwidth and not computation on GPU
2. Per request to tensor transformation is not batched which unless we're using CUDA graphs will be slow
3. `torch.cat` is slow because for every new request we are resizing a tensor and instead to instantiate one large tensor once and then use indexing notation to fill it up

### Solution

1. Instantiate a single tensor on GPU
2. Run `to_tensor()` operations on the entire requests object at once instead of iterating in a for loop
3. Use indexing notation to fill out elements from a batch and optionally use CUDA graphs to make this more efficient

Once we settle on this more performant approach we can then consolidate the different batching implementations that users need to develop from within a handler into a single base handler

### Motivation, pitch

The benefit of the new approach is that reduces the time it takes to run our `preprocess()` function and will directly translate in a lower end to end inference time for any consumer of our models.

ozancaglayan:
Hello @msaroufim . A quick question that boggles my mind on this: what happens if N different users pushes requests at more or less the same time to the API and their requests are internally batched? How do the N users get the N responses that correspond to their requests in the end?

msaroufim:
That's a good question so right now TorchServe would return N requests back to each of the N users and a good long-term solution for this would be us to add a middleware layer that would keep track of who made the request in a batch and send it back to them only.

But a good workaround I've seen others use is this, each of the `preprocess()`, `inference()` and `postprocess()` have the following signature `f(list : List) -> List` so you can not just return the inference request but also add in a `Dict` in the response that would map each response to a requester and then on the requester side look for that response ID.

PR #1798, opened 2022-08-16
Trigger Benchmark using Keyword in PR

msaroufim:
This PR checks if the keyword at the top shows up anywhere in the PR description and if it does it triggers our benchmark suite. This is useful for times when you need to check if a given change improves performance or not without setting up a manual suite each time. Hopefully this will make it easier to easily measure and more quickly integrate performance related changes like #1631 and #1797 

Test is below

h/t @xuzhao9

### DIFF
--- a/.github/workflows/benchmark_nightly_cpu.yml
+++ b/.github/workflows/benchmark_nightly_cpu.yml
@@ -4,10 +4,13 @@ on:
   # run every day at 2:15am
   schedule:
     - cron:  '15 02 * * *'
+  pull_request:
+
 jobs:
   nightly:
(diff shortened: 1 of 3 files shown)


msaroufim:
## What benchmarks will look like without keyword

Screen Shot 2022-08-16 at 3 20 12 PM

## What benchmarks will look like with `RUN_BENCHMARK` keyword

rohithkrn:
LGTM, wondering if we can document this as a note somewhere, May be in `CONTRIBUTING.MD`?

rohithkrn (approved):

agunapal (approved):

PR #1802, opened 2022-08-17
Util fixes

ozancaglayan:
## Description

- Formatting fixes
- Removes the spurious `isinstance()` check (#1020)
- Fix format string error for `PredictionException`

## Type of change


ts/utils/util.py
```diff
@@ -81,11 +96,14 @@ def map_class_to_label(probs, mapping=None, lbl_classes=None):
...
+
 class PredictionException(Exception):
     def __init__(self, message, error_code=500):
         self.message = message
         self.error_code = error_code
         super().__init__(message)
 
     def __str__(self):
-        return "message : error_code".format(message=self.message, error_code=self.error_code)
+        return "{message} : {error_code}".format(
```
msaroufim, line 107: since we're here do you mind changing this to an `f` string

ts/utils/util.py
```diff
@@ -22,51 +23,65 @@ def list_classes_from_module(module, parent_class=None):
...
+
 def load_label_mapping(mapping_file_path):
     """
     Load a JSON mapping { class ID -> friendly class name }.
     Used in BaseHandler.
     """
     if not os.path.isfile(mapping_file_path):
-        logger.warning('Missing the index_to_name.json file. Inference output will not include class name.')
+        logger.warning(
+            "'index_to_name.json' file is missing. Inference output will not include class name."
```
msaroufim, line 48: we can improve this to show the mapping file path

ts/utils/util.py
```diff
@@ -22,51 +23,65 @@ def list_classes_from_module(module, parent_class=None):
...
+        mapping = {str(k): v for k, v in enumerate(mapping["object_type_names"])}
         return mapping
 
     for key, value in mapping.items():
         new_value = value
         if isinstance(new_value, list):
             new_value = value[-1]
         if not isinstance(new_value, str):
-            raise Exception('labels in index_to_name must be either str or [str]')
+            raise Exception("labels in index_to_name must be either str or [str]")
```
msaroufim, line 71: how about List[str] instead of [str]

msaroufim (changes requested):
Thank you! Since we're here might as well improve some of the logging messages otherwise easy to merge

ts/utils/util.py
ozancaglayan: yeah sure, I was wondering whether you are being conservative for earlier python versions, otherwise I find f-strings much more concise. Does this mean that, I should replace `%s` style substitutions with f-strings if I hit them?

ts/utils/util.py
ozancaglayan: done

ts/utils/util.py
ozancaglayan: done

ts/utils/util.py
msaroufim: Yeah for security reasons we support only Python 3.8+. I've been wanting to do all these upgrades but it's easier to manage if we do it per new file that someone happened to touch

msaroufim (approved):

ozancaglayan:
should we merge this or do you want to wait for a 2nd approval?

rohithkrn (approved):

PR #1811, opened 2022-08-24
Temporarily reverting PR 1798 

agunapal:
## Description

There is a failure in BenchMark CI https://github.com/pytorch/serve/actions/runs/2908587948

This happened because multiple instances of benchmark CI runs were running concurrently after the commit in PR 1798

Fixes #(issue)

## Type of change

Run benchmark CI on nightly basis only.
Also, added a None type check to avoid a crash.

## Checklist:

- [ ] Did you have fun?
- [ Yes] Have you made corresponding changes to the documentation?

### DIFF
--- a/benchmarks/benchmark-ab.py
+++ b/benchmarks/benchmark-ab.py
@@ -398,7 +398,9 @@ def prepare_common_dependency():
         execution_params["config_properties"],
         os.path.join(execution_params["tmp_dir"], "benchmark", "conf"),
     )
-    shutil.copyfile(input, os.path.join(execution_params["tmp_dir"], "benchmark", "input"))
+    shutil.copyfile(
+        input, os.path.join(execution_params["tmp_dir"], "benchmark", "input")
+    )
 
 
 def getAPIS():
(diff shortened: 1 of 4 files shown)


benchmarks/benchmark-ab.py
```diff
@@ -505,11 +507,16 @@ def generate_csv_output():
     artifacts["TS latency P90"] = extract_entity(data, "90%", -1)
     artifacts["TS latency P99"] = extract_entity(data, "99%", -1)
     artifacts["TS latency mean"] = extract_entity(data, "Time per request:.*mean\)", -3)
-    artifacts["TS error rate"] = (
-        int(artifacts["TS failed requests"]) / execution_params["requests"] * 100
-    )
+    if isinstance(artifacts["TS failed requests"], type(None)):
+        artifacts["TS error rate"] = None
```
lxning, line 511: artifacts["TS error rate"] should be number

agunapal: changed to 0.0

msaroufim (approved):

lxning (approved):

msaroufim:
Let's revisit why jobs weren't queued, I don't get why the job would fail. We can potentially also only reenable this change for GPU since it only takes 30min whereas CPU takes 12h so we'd need to reduce the time there first

PR #1814, opened 2022-08-24
cpp backend logging and load model

maaquib:
## Description
* Folly log integration: I tried to implement a factory pattern to keep logging decoupled to some extent but Generics and Macros were giving a lot of trouble. So I have kept it tightly coupled to folly logging for the time being. Will do [type erasures](https://www.artima.com/articles/on-the-tension-between-object-oriented-and-generic-programming-in-c) to fix this once the other functionality has been implemented
* Added encoding for successful load model response
* TODO: Add more tests for OTF protocol and logging macros

### Create sample request for testing
```python
import struct

buf = bytearray()
buf += b'L'
model_name = b'mnist'
buf+=struct.pack("!i", len(model_name))
buf+=model_name
model_path=b'/tmp/models/b946065d4921408b88bc89f0b1bc5cd6'
buf+=struct.pack("!i", len(model_path))
buf+=model_path
#Batch Size
buf+=struct.pack("!i",1)
handler=b'BaseHandler'
buf+=struct.pack("!i", len(handler))
buf+=handler
#GPU ID
buf+=struct.pack("!i",-1)
#Envelope
envelope=b''
buf+=struct.pack("!i", len(envelope))
buf+=envelope
buf+=struct.pack("!?", True)

with open('req.bin', 'rb') as f:
    f.write(buf)
```

### Test happy case
```bash
$TS1> cat logging.config
INFO:default; default=file:path=/tmp/ts.log,async=false
$TS1> ./build.sh && ./_build/src/backends/model_worker_socket --model_dir /tmp/models/b946065d4921408b88bc89f0b1bc5cd6 --logger_config_path logging.config
$TS2> nc 127.0.0.1 9000 < req.bin > response.txt
```

### Test response
```python
import struct

with open('response.txt', 'rb') as fb:
    status_code = struct.unpack("!i", fb.read(4))
    length = struct.unpack("!i", fb.read(4))
    message = fb.read(length[0])
    no_pred = struct.unpack("!i", fb.read(4))

status_code, length, message, no_pred
```

### Model dir
```bash
> tree /tmp/models/b946065d4921408b88bc89f0b1bc5cd6/
/tmp/models/b946065d4921408b88bc89f0b1bc5cd6/
├── MAR-INF
│   └── MANIFEST.json
└── mnist_cnn.pt
> cat /tmp/models/b946065d4921408b88bc89f0b1bc5cd6/MAR-INF/MANIFEST.json     
{
  "createdOn": "23/08/2022 11:51:46",
  "runtime": "python",
  "model": {
    "modelName": "mnist",
    "serializedFile": "mnist_cnn.pt",
    "handler": "BaseHandler",
    "modelFile": "mnist.py",
    "modelVersion": "1.0"
  },
  "archiverVersion": "0.5.3"
}
```


cpp/src/backends/protocol/otf_message.hh
```diff
@@ -28,7 +28,9 @@ namespace torchserve {
 
   class OTFMessage {
     public:
-    static byte_buffer CreateLoadModelResponse(StatusCode code, const std::string& message);
+    static bool SendAll(Socket conn, char *data, size_t length);
+    static void CreateLoadModelResponse(std::unique_ptr<torchserve::LoadModelResponse> response, char* data);
```
rohithkrn, line 32: I think `EncodeLoadModelResponse` makes more sense since we are just encoding an existing response.

cpp/src/backends/protocol/otf_message.cc
```diff
@@ -1,21 +1,50 @@
...
+    p += sizeof(no_predict);
+  }
+
+  bool OTFMessage::SendLoadModelResponse(Socket client_socket_, std::unique_ptr<torchserve::LoadModelResponse> response) {
+    char *data = new char[sizeof(LoadModelResponse)];
+    torchserve::OTFMessage::CreateLoadModelResponse(std::move(response), data);
+    if(!torchserve::OTFMessage::SendAll(client_socket_, data, sizeof(LoadModelResponse))) {
+      return false;
+    }
+    delete[] data;
```
rohithkrn, line 46: If `SendAll` fails for some reason, we will miss deleting this data?

maaquib: Fixed

cpp/src/backends/protocol/otf_message.hh
maaquib: Refactored

cpp/src/backends/protocol/otf_message.cc
```diff
@@ -1,30 +1,41 @@
...
+    // Expectation from frontend deserializer is a -1
+    // at the end of a LoadModelResponse
+    int32_t no_predict = htonl(response->predictions);
+    memcpy(p, &no_predict, sizeof(no_predict));
+    p += sizeof(no_predict);
   }
 
-  char OTFMessage::RetrieveCmd(Socket conn) {
+  bool OTFMessage::SendLoadModelResponse(const ISocket& client_socket_, std::unique_ptr<torchserve::LoadModelResponse> response) {
+    char *data = new char[sizeof(LoadModelResponse)];
```
rohithkrn, line 25: does the `sizeof` operator on `LoadModelResponse` object always yield the right size of the data? In the case of inference response, it did not return the expected response probably because it holds pointers to the data instead of the actual data.

cpp/src/backends/protocol/otf_message.cc
```diff
@@ -1,30 +1,41 @@
...
-    };
-    std::byte msg[sizeof(LoadModelResponse)];
-    std::memcpy(msg, &response, sizeof(LoadModelResponse));
-    byte_buffer response_byte_buffer;
-    std::copy(response_byte_buffer.begin(), response_byte_buffer.end(), msg);
-    return response_byte_buffer;
+  void OTFMessage::EncodeLoadModelResponse(std::unique_ptr<torchserve::LoadModelResponse> response, char* data) {
+    char* p = data;
+    // Serialize response code
+    int32_t s_code = htonl(response->code);
```
rohithkrn, line 7: nit: do you think we can move adding integer to data pointer to a util function and re-use later below?

cpp/src/backends/protocol/otf_message.cc
maaquib: Will fix this

cpp/src/backends/protocol/otf_message.cc
maaquib: Done

rohithkrn:
Merged as part of https://github.com/pytorch/serve/pull/1817. Closing this.

maaquib:
Changes merged as part of https://github.com/pytorch/serve/pull/1817

PR #1829, opened 2022-08-30
Fix broken link in Getting Started to Torch Model archiver

Alwahsh:
## Description

Lots of links in Getting started are not working, probably because it was moved into docs. This fixes links to point to the correct files.

## Type of change

No Logic Change - Just a documentation update.

N/A

## Checklist:

N.A

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -24,12 +24,12 @@
 
     #### For Windows
 
-    Refer to the documentation [here](docs/torchserve_on_win_native.md).
+    Refer to the documentation [here](torchserve_on_win_native.md).
 
 2. Install torchserve, torch-model-archiver and torch-workflow-archiver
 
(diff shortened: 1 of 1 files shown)


msaroufim:
I think I'd like us to stop merging individual PRs to fix links. The real solution is to integrate this into our linter github action so if you'd like to pick that up I would be happy to give you pointers

msaroufim:
I think I'd like us to stop merging individual PRs to fix links. The real solution is to integrate this into our linter github action so if you'd like to pick that up I would be happy to give you pointers

msaroufim:
I think I'd like us to stop merging individual PRs to fix links. The real solution is to integrate this into our linter github action so if you'd like to pick that up I would be happy to give you pointers

msaroufim:
Closing as this work will be subsumed by #1855

PR #1837, opened 2022-09-01
Reverted RUN_BENCHMARK revert

### LINKED ISSUE #1834: Nightly CPU benchmark failed
Commit b62e5d7ab23e295c333a0422023edfba4f579e6a daily scheduled [CI run](https://github.com/pytorch/serve/actions/runs/2961077745) failed, please check why

msaroufim:
## Description

Fixes #1834 

We had initially reverted the ability to run benchmarks with a keyword because we believed that it was causing issues when concurrent benchmarks were ran. I suspect that wasn't quite the problem so decided to retrigger benchmarks and fix the issue which I believe was using None when there were no failed requests. If the benchmark CI passes below then it means the fix worked but I do unfortunately need to wait 16h to confirm. 

## Type of change

Before: #1834 
After: benchmark CI signals below

The CI signals will also confirm whether the benchmark script is flaky, which we cannot afford because it's 16h long and too slow to rerun

### DIFF
--- a/.github/workflows/benchmark_nightly_cpu.yml
+++ b/.github/workflows/benchmark_nightly_cpu.yml
@@ -1,9 +1,10 @@
 name: Benchmark torchserve cpu nightly
 
+on:
   # run every day at 2:15am
-schedule:
-  - cron:  '15 02 * * *'
-pull_request:
+  schedule:
+    - cron:  '15 02 * * *'
+  pull_request:
 
 jobs:
   nightly:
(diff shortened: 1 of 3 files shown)


msaroufim:
Also looking at the actions tab now we can see the scheduled action and the queued action and we can confirm whether the timeouts do actually happen or whether we can queue benchmark jobs just fine

Screen Shot 2022-08-31 at 10 15 54 PM

PR #1794, opened 2022-08-15
Fixes dead links

NeoKish:
## Description

Please read our [CONTRIBUTING.md](https://github.com/pytorch/serve/blob/master/CONTRIBUTING.md) prior to creating your first pull request.

Please include a summary of the feature or issue being fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes #pytorch/pytorch.github.io#1034

PR #1855, opened 2022-09-10
Add linting steps to doc-automation workflow

sadra-barikbin:
spellcheck and linkcheck was added to workflow yml file.

## Checklist:

- [x] Did you have fun?


.github/workflows/doc-automation.yml
```diff
@@ -21,6 +21,10 @@ jobs:
         cd ./docs/sphinx
         make html
       id: build
+    - name: Check links in all markdown files
```
msaroufim, line 24: So the problem with this approach is `doc-automation` runs on merge to master because we can't change our documentation based on whatever commits anyone makes

My suggestion would be to make this work on `lint.yml`. 

Also we need to make sure that the if the output of those scripts is not empty then the action needs to fail (only in lint.yml). Also if you are running the markdown link checker did it print broken links that you were then able to fix? 

Finally for broken links in general I think it's fine to fix them all at once in a single PR but for typos in documentation I don't think those are critical to fix in batch and we should instead just show users typos in the files that they changed only (I do this already to check for linting so just borrow same idea)

ts_scripts/install_utils
```diff
@@ -303,4 +303,6 @@ run_markdown_link_checker(){
   done
   set -e
   exit $STATUS
-}
\ No newline at end of file
+}
+
+"$@"
```
msaroufim, line 308: what's the goal of this change?

msaroufim (changes requested):

.github/workflows/doc-automation.yml
sadra-barikbin: So you want both spellchecking and link-checking to be in `lint.yml`. As for link-checking, you want the whole documentation to be checked, but for spellchecking just the modified files. Am I correct?
For link-checking, how is it to use `gaurav-nelson/github-action-markdown-link-check` as a step in in the `build` job?

For spellchecking we could add a dedicated job to `lint` workflow for it to run in parallel or do it as a step in the same `build` job.

msaroufim:
Thanks @sadra-barikbin this is a good step forward, to merge this here's what's left
1. The link checker script is working so can you fix all those links? Otherwise all future PRs will will temporarily fail this check until someone does do this. Also check that the output is correct for example I saw some errors pointed out in the ipex example that seemed just fine to me
2. For the spellchecking, it should only run on the `md` files that were changed in the PR which you can get like so https://github.com/pytorch/serve/blob/788ede8c2b032a5e46fdceb1e5e1c5f82367b9cd/.github/workflows/lint.yml#L28 and when the check does fail you should tell people what they need to do either fix the typos or add a new word in the dictionary

msaroufim:
I noticed you're fixing some links, an important detail is for links in docs folder you need absolute links if you're linking to anything outside of docs otherwise the generated link will be broken

sadra-barikbin:
Regarding those links, link-checker didn't raise an error. Where exactly you mean those links would break?

msaroufim:
Here's a good example https://github.com/pytorch/serve/pull/1794/files

sadra-barikbin:
I did. Some link errors need your review. Also I couldn't fix links in `examples/intel_extension_for_pytorch`. Do you have any idea?

sadra-barikbin:
Is spellcheck really a necessary thing? New words, sequences etc. must get added to its dictionary again and again. We could fix all typos now and remove this step. From now on, typos would most likely be fixed during PR reviews.

msaroufim:
Spellchecking is definitely lower priority than finding broken links we can finish that first

sadra-barikbin:
@msaroufim any blocker?

msaroufim:
The Lint jobs are both red, they should be green for us to merge this specific change. Spellchecking needs to be progressive (only apply to changed files) as we discussed or not used

Also should resolve the merge conflict

sadra-barikbin:
Some links need your review:
- I could not find a file named `Transformer_handler_generalized_v2.py` in the repo. referenced in `kubernetes/kserve/kf_request_json/v2/bert/README.md`. By the way there is `examples/Huggingface_Transformers/Transformer_handler_generalized.py`
- `[input](10000)`s cause error in `benchmarks/sample_report.md` and I don't know their purpose.
- I can't find `https://github.com/pytorch/serve/tree/master/test/benchmark`, referenced in `CONTRIBUTING.md`
- I can't find `https://github.com/pytorch/serve/settings/actions/runners` referenced in `benchmarks/README.md`

sadra-barikbin:
The last commit is about a link which due to a bug in `link-check` (which `markdown-link-check` makes use of), action raises an error. I filed an issue for them but for now I did a workaround for that link.

msaroufim:
Great to see the lint spellcheck now green

For the links here are some answers to your specific questions
* Looks like upstream the file changed, I believe this is the correct one https://github.com/kserve/kserve/blob/master/docs/samples/v1beta1/torchserve/v2/bert/sequence_classification/Transformer_kserve_handler.py - this is great though already showing the value of your linter
* We can make these not links https://github.com/pytorch/serve/blob/master/benchmarks/sample_report.md
* Reference this instead https://github.com/pytorch/serve/tree/master/benchmarks
* Remove runners link since that's only visible to admins of the repo so replace with some phrase like "check the runners tab on Github"

msaroufim (approved):
Heroic effort thank you!

sadra-barikbin:
@msaroufim , This repo isn't actively maintained?

msaroufim:
It is, folks are just busy with other stuff @lxning @namannandan @agunapal @rohithkrn can one of you please take a look?

.github/workflows/lint.yml
```diff
@@ -43,3 +49,33 @@ jobs:
...
+          files: | 
+            **/*.md
+
+      - name: Check spellings
+        run: |
+          sources=""
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            sources+=" -S $file"
+          done
+          pyspelling -c $GITHUB_WORKSPACE/ts_scripts/spellcheck_conf/spellcheck.yaml --name Markdown $sources
```
namannandan, line 75: [wordlist.dic](https://github.com/pytorch/serve/blob/master/ts_scripts/spellcheck_conf/wordlist.dic) is currently tracked by git. Since we've updated the `wordlist.txt` file, should we include the updated `wordlist.dic` file as well in this commit? If it is auto-generated each time `pyspelling` is run, we could `git rm` the file and not track it in future by including it in `.gitignore`

ts_scripts/spellcheck_conf/wordlist.txt
```diff
@@ -615,4 +615,272 @@ scarpe
...
+FIxvz
+GOR
+InferenceService
+JCSEk
+JmOXoSsAAKK
+Knative
+NOPQVgsIcW
+PNnKQ
+PV
+QHbuDCgr
```
namannandan, line 804: Nit: Where do we see this string in the repo? Was wondering how this and a few other such strings got included here?

namannandan (approved):

msaroufim:
@sadra-barikbin please address the last piece of feedback from @namannandan, get the spellcheck job to be green and we should be good to merge this

.github/workflows/lint.yml
```diff
@@ -20,18 +20,24 @@ jobs:
...
         run: |
           pip install pre-commit
-          pre-commit install      
+          pre-commit install
 
-      - name: Get specific changed files
-        id: changed-files-specific
-        uses: tj-actions/changed-files@v20.1
+      - name: Check links in all markdown files
+        uses: gaurav-nelson/github-action-markdown-link-check@1.0.14
```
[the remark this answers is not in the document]
sadra-barikbin: No. I change it to 1.0.13

.github/workflows/lint.yml
sadra-barikbin: I agree. I do it

ts_scripts/spellcheck_conf/wordlist.txt
sadra-barikbin: In URLs, keys and tokens. I added a step to spellcheck to remove urls and fenced codes.

sadra-barikbin:
Done

PR #1857, opened 2022-09-12
Add ONNX and ORT support + Docs for TensorRT

msaroufim:
EDIT 11/4: To make this easier to merge I've cut the scope to only ONNX and ORT, will revisit the rest after pytorch/tensorrt when it gets official pypi binaries but this is now ready for review

EDIT 11/6: I'm going to write a brief doc page on using different optimization runtimes

EDIT 11/8: Addressed most feedback, rest will be addressed in future work

## This PR
* a `--serialized-file` that's in `.onnx` format, which will be correctly loaded by the base handler using an `ort.InferenceSession()` https://pytorch.org/tutorials/advanced/super_resolution_with_onnxruntime.html
* TensorRT format is a `.ts` extension which can be loaded via `torch.jit.load()` https://pytorch.org/TensorRT/getting_started/getting_started_with_python_api.html#getting-started-with-python-api
* A hardcoded `model_config.json` similar to `index_to_name.json` which would have model specific information which you can get access via `model_config.get("property"). For now I'm not using any special configs
* Starting with torch 1.12 nvfuser is now the default backend for torchscript so that's a free performance enhancement we got where we don't need to do anything

## Open question
* Should we do the conversion from torch to ONNX in the `initialize()`? My gut is no, we don't ask users to train models in `initialize()` they should prepare their models and once prepared used `serve` but I could be convinced otherwise - EDIT: NO
* Should I put a schema on `model_config.json`? I'd like to wait and see what schema gets chosen for #1664 and then decide
* There are other ways of designing the same thing but I felt like this was the simplest, I thought about the here https://gist.github.com/msaroufim/c5aa505fdd7053ca184da921536c26f1

## Future work
* Add AITemplate support https://ai.facebook.com/blog/gpu-inference-engine-nvidia-amd-open-source/?utm_source=twitter&utm_medium=organic_social&utm_campaign=blog
* nvfuser is actually already supported and starting 1.12 has become the default fusing backend for `torch.jit.script()`


model-archiver/model_archiver/arg_parser.py
```diff
@@ -19,101 +18,130 @@ class ArgParser(object):
...
+            "--handler",
+            required=True,
+            dest="handler",
+            type=str,
+            default=None,
+            help="TorchServe's default handler name\n"
+            " or Handler path to handle custom inference logic.",
+        )
+
+        parser_export.add_argument(
```
HamidShojanazeri, line 71: Thanks @msaroufim for the PR, so we are thinking of passing the model_config file optional as part of extra file here?

ts/torch_handler/base_handler.py
```diff
@@ -33,6 +35,14 @@
             "IPEX is enabled but intel-extension-for-pytorch is not installed. Proceeding without IPEX."
         )
 
+try:
```
HamidShojanazeri, line 38: should we guard it behind some sort of condition? maybe moving it after we found serialized model has .onnx extension?

ts/torch_handler/base_handler.py
```diff
@@ -78,38 +89,66 @@ def initialize(self, context):
...
-            if not os.path.isfile(model_pt_path):
-                raise RuntimeError("Missing the model.pt file")
+            self.model.eval()
+
+        elif self.model_pt_path.endswith(".pt"):
+            logger.info("Loading torchscript model")
+            self.model = self._load_torchscript_model(self.model_pt_path)
+            self.model.eval()
+
+        elif self.model_pt_path.endswith(".ts") and tensorrt_enabled:
```
HamidShojanazeri, line 113: TorchTensort module could also be saved as .pt extension as well. https://github.com/pytorch/TensorRT/blob/master/examples/fx/fx2trt_example.py#L138, We might be able to use the model_config to specify the runtime. It may be an extra step for user but it can keep the code and logic clear for user and avoid any potential error while deploying the model.

HamidShojanazeri:
We also would need to think of preprocessing for onnx, as the input need to be converted to numpy. We might be able to have separate util files for onnx, trt or future backend or alternatively have one backend utils to keep all the related helper functions there. This may help to keep the base_handler cleaner and more maintainable. cc @lxning on this as well.

model-archiver/model_archiver/arg_parser.py
msaroufim: Yup that's exactly right

ts/torch_handler/base_handler.py
msaroufim: Ok yeah that makes sense I'll move tensor RT out of packaging and into the model config

model-archiver/model_archiver/arg_parser.py
```diff
@@ -19,101 +18,130 @@ class ArgParser(object):
     @staticmethod
     def export_model_args_parser():
```
lxning, line 19: model archiver needs to be extended to support new runtime types: eg. onnx, tensorRT

requirements/developer.txt
```diff
@@ -13,4 +13,5 @@ pygit2==1.6.1
 pyspelling
 pre-commit
 twine
+onnxruntime
```
lxning, line 16: it should be dynamically loaded based on model runtime type.

ts/torch_handler/base_handler.py
```diff
@@ -33,6 +35,14 @@
             "IPEX is enabled but intel-extension-for-pytorch is not installed. Proceeding without IPEX."
         )
 
+try:
+    import onnxruntime
+
+    onnx_enabled = True
```
lxning, line 41: it should be based on model runtime type definition, not hard coded.

msaroufim:
Just a quick update, the main challenge for this PR will be correctly packaging the installation scripts and I'm seeing all sorts of basic issues even with basic models to setup some E2E test. Some relevant issues
* https://github.com/pytorch/TensorRT/issues/1233#issuecomment-1274037109
* https://github.com/pytorch/ort/issues/150

requirements/developer.txt
msaroufim: I think we should do that for our common dependencies but for our developer dependencies we should install everything so we can run tests in CI

ts/torch_handler/base_handler.py
msaroufim: I'll work on tensorRT in another PR

ts/torch_handler/base_handler.py
msaroufim: great idea, done

msaroufim:
And here's an example of an inference running from my logs. Repro is in `test/pytest/test_onnx.py`



```
(base) ubuntu@ip-172-31-17-70:~$ curl -X POST http://127.0.0.1:8080/predictions/onnx --data-binary '1'
[
  -0.28003010153770447
](base) ubuntu@ip-172-31-17-70:~$ 
```

ts/torch_handler/base_handler.py
```diff
@@ -235,6 +307,8 @@ def handle(self, data, context):
             if self._is_describe():
                 output = [self.describe_handle()]
             else:
+                # What's a better way to do this
+                data = torch.randn(1, 1)
```
msaroufim, line 311: Looking for feedback on this, not sure how I can curl data that I don't want to be preprocessed

ts/torch_handler/base_handler.py
```diff
@@ -78,38 +81,101 @@ def initialize(self, context):
...
+        # For TensorRT support follow instructions here: https://pytorch.org/TensorRT/getting_started/getting_started_with_python_api.html#getting-started-with-python-api
+        elif self.model_pt_path.endswith(".pt"):
+            self.model = self._load_torchscript_model(self.model_pt_path)
+            self.model.eval()
+
+        # Convert your model by following instructions: https://pytorch.org/tutorials/advanced/super_resolution_with_onnxruntime.html
+        elif self.model_pt_path.endswith(".onnx") and onnx_enabled:
+            # self.model = self._load_onnx_model(self.model_pt_path)
+            providers = (
+                ["CUDAExecutionProvider", "CPUExecutionProvider"]
```
HamidShojanazeri, line 121: I wonder if it should be "CUDAExecutionProvider" instead of "  ["CUDAExecutionProvider", "CPUExecutionProvider"]"

ts/torch_handler/base_handler.py
```diff
@@ -78,38 +81,101 @@ def initialize(self, context):
...
-        logger.debug("Model file %s loaded successfully", model_pt_path)
+        logger.debug("Model file %s loaded successfully", self.model_pt_path)
 
         # Load class mapping for classifiers
         mapping_file_path = os.path.join(model_dir, "index_to_name.json")
         self.mapping = load_label_mapping(mapping_file_path)
 
         self.initialized = True
 
+    def _load_onnx_model(self, model_onnx_path):
```
HamidShojanazeri, line 151: are we going to replace the above Lines 120-137 with this function?

ts/torch_handler/base_handler.py
```diff
@@ -183,9 +250,14 @@ def inference(self, data, *args, **kwargs):
         Returns:
             Torch Tensor : The Predicted Torch Tensor is returned in this function.
         """
-        marshalled_data = data.to(self.device)
         with torch.no_grad():
-            results = self.model(marshalled_data, *args, **kwargs)
+            if hasattr(self.model, "run"):
+                data = data.to(torch.float32).cpu().numpy()
+                # TODO: Should we make this "modelInput configurable", feels complicated
+                results = self.model.run(None, {"modelInput": data})[0]
```
HamidShojanazeri, line 257: I think we should leave this part to user, and have documents that ask user to implement the inference or maybe a utility like

```
output =[node.name for node in model.graph.output]

input_all = [node.name for node in model.graph.input]
input_initializer =  [node.name for node in model.graph.initializer]
input = list(set(input_all)  - set(input_initializer))
```
may help to then assert if the passed input matches the expected ones?

ts/torch_handler/base_handler.py
HamidShojanazeri: Not sure what we are trying to do here?

HamidShojanazeri (commented):
Thanks @msaroufim , I added some comments, but more generally I wonder if we have plans to move all the backend initialization logics + utilities into a backend specific files / utils. This would make the base_handler easier to maintain and more readable. Maybe its a good boot camp task?

ts/torch_handler/base_handler.py
msaroufim: No it's both because you may need to fallback to CPU for some operations, the tutorial I link in tests shows this very pattern

ts/torch_handler/base_handler.py
msaroufim: Yeah this is a typo, will fix

ts/torch_handler/base_handler.py
msaroufim: So the base handler does preprocess data because it can't make assumptions about the data type

So I'd like to essentially curl a torch tensor for the inference directly and I'm not sure what the right way to do that is. One workaround is I can inherit from this handler and in my test have a preprocesser but if I put the test here I believe it will be more general and safer

ts/torch_handler/base_handler.py
msaroufim: Yeah interesting idea, lemme take a look at this

setup.py
```diff
@@ -157,25 +158,29 @@ def run(self):
...
+            "build_frontend": BuildFrontEnd,
+            "build_plugins": BuildPlugins,
+            "build_py": BuildPy,
         },
         install_requires=requirements,
-        entry_points={'console_scripts': ['torchserve=ts.model_server:start',]},
+        extras_require={
+            "onnx": ["numpy", "onnx","onnx-runtime"],
+            "ipex": ["intel_extension_for_pytorch"],
+        },
```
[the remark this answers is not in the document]
msaroufim: Sure, this optional dependency pattern is one I've seen used by Ray for example the idea is you could run

```
pip install torchserve[onnx]
pip install ipex
```

To get the right dependencies

Now as far as putting things in `requirements-dev.txt` my take is everything that needs to be tested should be there. We can otherwise potentially explore `requirements-test.txt` where there should be no size limits

requirements/developer.txt
```diff
@@ -14,3 +14,6 @@ pyspelling
 pre-commit
 twine
 intel_extension_for_pytorch; sys_platform != 'win32' and sys_platform != 'darwin'
+onnx
+onnx-runtime
+numpy
```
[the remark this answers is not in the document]
msaroufim: That's true, I guess for dev dependencies that seems fine though. I can remove these if we disable the ONNX tests by default but I don't believe we should either

ts/torch_handler/base_handler.py
```diff
@@ -78,38 +81,101 @@ def initialize(self, context):
...
 
         model_dir = properties.get("model_dir")
-        model_pt_path = None
+        self.model_pt_path = None
         if "serializedFile" in self.manifest["model"]:
             serialized_file = self.manifest["model"]["serializedFile"]
-            model_pt_path = os.path.join(model_dir, serialized_file)
+            self.model_pt_path = os.path.join(model_dir, serialized_file)
+
+        if self.model_pt_path.endswith("onnx"):
```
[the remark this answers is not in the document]
msaroufim: Ah I see I like that idea is interesting because it makes it clear that the backends are completely different, this will make testing much easier on my end and checking end to end coverage much easier like for example does batching work? Does captum work with ONNX? etc..

I guess the tradeoff is because is we'll need another layer of handlers then we won't be able to reuse any code

ts/torch_handler/base_handler.py
msaroufim: Another idea would be to not run the inference at all in the base handler

I can recommend users create their own inference function where the inference happens so that the scope of ONNX support really means initializing the model

msaroufim:
Discussed offline with Li and Hamid
* Disable test by default in CI
* Refactor base handler
* Remove requirements from developer dependencies
* Add documentation to describe how people should pre and post process

Future
* automation of pre and post processing

msaroufim:
@HamidShojanazeri @lxning I made all the changes we discussed 


## Logs

```
ubuntu@ip-172-31-17-70:~/serve/test/pytest$ curl -X POST http://127.0.0.1:8080/predictions/onnx --data-binary '1'
Prediction Succeeded
```

## Pytest

```
(ort) ubuntu@ip-172-31-17-70:~/serve/test/pytest$ pytest test_onnx.py 
==================================== test session starts =====================================
platform linux -- Python 3.8.13, pytest-7.2.0, pluggy-1.0.0
rootdir: /home/ubuntu/serve
plugins: mock-3.10.0, cov-4.0.0
collected 5 items                                                                            

test_onnx.py .....                                                                     [100%]

===================================== 5 passed in 1.17s ======================================
```

docs/performance_guide.md
```diff
@@ -5,6 +5,16 @@ There are many tricks to optimize PyTorch models for production including but no
...
+1. Package serialized ONNX weights `torch-model-archiver --serialized-file model.onnx ...`
+2. Load those weights from `base_handler.py` using `ort_session = ort.InferenceSession(self.model_pt_path, providers=providers, sess_options=sess_options)` which supports reasonable defaults for both CPU and GPU inference
+3. Allow you define custom pre and post processing functions to pass in data in the format your onnx model expects
+
+### TensorRT and NVfuser support
+TorchServe also already supports models optimized via TensorRT. To leverage the TensorRT runtime you can follow examples here https://github.com/pytorch/TensorRT and once you're done you'll have serialized weights which you can load with `torch.jit.load()` https://pytorch.org/TensorRT/getting_started/getting_started_with_python_api.html#getting-started-with-python-api so for all intents and purposes after a conversion there is no difference in how PyTorch treats a Torchscript model vs a TensorRT model. The benefit of using `torch.jit.load()` is that it's now also the default fuser in PyTorch starting from PyTorch `1.12`
```
HamidShojanazeri, line 19: I believe TorchTRT through [FX path](https://github.com/pytorch/TensorRT/blob/master/examples/fx/fx2trt_example.py) can also let user to save the optimized model in .pt as well. It may worth mentioning the FX path here and TorchTRT as the preferred way/ new/ available option to go.

HamidShojanazeri (approved):
Thanks @msaroufim , LGTM. I also suggest for next set of PRs maybe having linting/ formatting as a separate PR makes easier to focus on the changes.

lxning (approved):

PR #1858, opened 2022-09-12
mypy type annotations

msaroufim:
mypy reduces bugs, makes code safer and makes it easier to ramp up. It's already used heavily in the vast majority of pytorch projects

Added support for a single file as an example of things running but over time we should increase type coverage. For example this PR found issues in metric implementation that invalidates Liskov so worth using this more.

Already found 1 bug thanks to this PR

```python
    def get_request_id(self, idx: int = 0) -> Optional[str]:
        if self.request_ids is None:
            return None
        return self.request_ids.get(idx)
```


msaroufim:
The type annotations seem to pass just fine https://github.com/pytorch/serve/actions/runs/3177448549/jobs/5177898647

agunapal (approved):

.github/workflows/mypy.yml
```diff
@@ -0,0 +1,38 @@
+name: Lint
```
mreso, line 1: Is there a reason why this is not another step in the lint.yml like spellcheck?
In the checks tab we now have 2 Lint workflows. See screenshot:
image

mreso (approved):
Beside the two Lint workflows (see comment) this LGTM

.github/workflows/mypy.yml
msaroufim: makes sense, fixed - will merge once I confirm CI signal is green

PR #1866, opened 2022-09-21
Validate torchserve process in pid file before starting

### LINKED ISSUE #1862: valid ts process during start
### 🚀 The feature

Torchserve needs to verify the [pid](https://github.com/pytorch/serve/blob/master/ts/model_server.py#L39) is a real torchserve process. 

### Motivation, pitch

This feature is able to avoid torchserve restart failure if system reuses a pid for the other application.

### Alternatives

_No response_

### Additional context

_No response_

rohithkrn:
## Description

Fixes #1862 - There's an edge case where if `.model_server.pid` file is not deleted when torchserve stops abruptly and another process with same pid is started before torchserve is attempted to start again at which point it falsely assumes that the new pid of another process is torchserve process and exits with message `"TorchServe is already running, please use torchserve --stop to stop TorchServe."` This additional check validates that the process is actually a torchserve process and if not it will start torchserve. 
Testing:
* Usual path - verified that torchserve start and stop work as before
* Edge case path - manually changed the pid file to replace pid with another process and verified that torchserve started without complaining that torchserve is already running.

### DIFF
--- a/ts/model_server.py
+++ b/ts/model_server.py
@@ -3,16 +3,19 @@ File to define the entry point to Model Server
 """
 
 import os
+import platform
 import re
 import subprocess
 import sys
 import tempfile
 from builtins import str
-import platform
 
 import psutil
-from ts.version import __version__
+
 from ts.arg_parser import ArgParser
+from ts.version import __version__
+
+TS_NAMESPACE = "org.pytorch.serve.ModelServer"
 
 
 def start():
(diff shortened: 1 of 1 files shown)


agunapal (approved):
looks good

ts/model_server.py
```diff
@@ -26,7 +29,13 @@ def start():
     if os.path.isfile(pid_file):
         with open(pid_file, "r") as f:
             pid = int(f.readline())
-
+            try:
+                process_from_pid_file = psutil.Process(pid)
+                pid = pid if TS_NAMESPACE in process_from_pid_file.cmdline() else None
```
msaroufim, line 34: What does a `None` pid mean?

rohithkrn:
> Could you please add some detail as to how you verified this fix worked?

@msaroufim updated the PR description to reflect the details

msaroufim (approved):

lxning (approved):

PR #1869, opened 2022-09-21
Support Ubuntu 20.04 GPU in Docker build

### LINKED ISSUE #1865: images with Ubuntu 20.04
### 🚀 The feature

images with Ubuntu 20.04 

### Motivation, pitch

Currently images are all ubuntu 18.04. if we can have the flexibility to build ubuntu20.04-based image that'll be great. e.g. ubuntu18 ships with gcc7, which don't support C++17 quite well. ubuntu20.04 ship with gcc9 with very good C++17 support and we also see some performance gains

### Alternatives

_No response_

### Additional context

_No response_

msaroufim:
## Description

This PR adds support for building Ubuntu 20.04 images, it works in a backwards compatible manner. If we're willing to break backwards compatibility I think we can build a much more usable version using `argparse`, `f-strings` in python but we don't have to do that now. I personally find bash more confusing than its worth.

To repro `build_image.sh -g -cv cu116 -ub ubuntu20.04`

Fixes #1865 

## Type of change

Change is backwards compatible

https://gist.github.com/msaroufim/f5ccb1feac7604245396cd3fe2bce6c7

## Checklist:

- [x] Did you have fun?


agunapal (approved):

agunapal (commented):
If someone specifies CUDA 10.2 with ubuntu 20, do we want to give this message

else
            echo "CUDA version not supported"
            exit 1

rohithkrn (commented):
I wonder if we test our docker container changes somehow? If not, we should probably come up with a set of tests to verify functionality.

rohithkrn (approved):

msaroufim:
> I wonder if we test our docker container changes somehow? If not, we should probably come up with a set of tests to verify functionality.

It's something we've talked about before but never got the change to prioritize. One layer of tests is our nightly builds which will break if the default configuration is wrong but that doesns't test the more exotic configurations

msaroufim:
I see this error in CI which does not seem related

```
Run python torchserve_sanity.py
  File "torchserve_sanity.py", line 5, in <module>
    from ts_scripts.sanity_utils import test_sanity
  File "/home/runner/work/serve/serve/ts_scripts/sanity_utils.py", line 3, in <module>
    import nvgpu
ModuleNotFoundError: No module named 'nvgpu'
Error: Process completed with exit code 1.
```

msaroufim:
Something weird happening with CI - will retry again tomorrow. I don't think these are related to my change since docker images are not used in CI

PR #1874, opened 2022-09-23
Metrics helper classes implementation for C++ backend

namannandan:
#### This PR implements the following Metrics helper classes in C++ for Torchserve backend
- Units
- Metric

#### Testing
- Each of the classes have their associated unit tests

#### Reference design
- https://github.com/pytorch/serve/issues/1492


cpp/src/backends/metrics/metric.cc
```diff
@@ -0,0 +1,61 @@
...
+#include "src/backends/metrics/units.hh"
+
+namespace torchserve {
+    template <typename ValueType>
+    Metric<ValueType>::Metric(
+            const std::string& name, const MetricType& type, const ValueType& value,
+            const std::string& unit, const std::vector<Dimension>& dimensions,
+            const std::string& request_id, const bool& is_updated) :
+            name{name}, type{type}, value{value}, unit{Units::GetUnitMapping(unit)},
+            dimensions{dimensions}, request_id{request_id}, is_updated{is_updated} {}
```
lxning, line 14: Pls take a look at [prometheus client API](https://github.com/jupp0r/prometheus-cpp).

cpp/src/utils/CMakeLists.txt
```diff
@@ -4,9 +4,13 @@ list(APPEND TS_UTILS_SOURCE_FILES ${TS_UTILS_SRC_DIR}/config.cc)
...
 list(APPEND TS_UTILS_SOURCE_FILES ${TS_UTILS_SRC_DIR}/model_archive.cc)
 list(APPEND TS_UTILS_SOURCE_FILES ${TS_UTILS_SRC_DIR}/logging.cc)
+list(APPEND TS_UTILS_SOURCE_FILES ${TS_UTILS_SRC_DIR}/metrics/units.cc)
+list(APPEND TS_UTILS_SOURCE_FILES ${TS_UTILS_SRC_DIR}/metrics/dimension.cc)
+list(APPEND TS_UTILS_SOURCE_FILES ${TS_UTILS_SRC_DIR}/metrics/metric.cc)
 add_library(ts_utils SHARED ${TS_UTILS_SOURCE_FILES})
 target_include_directories(ts_utils PUBLIC ${TS_UTILS_SRC_DIR})
-target_link_libraries(ts_utils ${FOLLY_LIBRARIES} ${CMAKE_DL_LIBS})
+target_include_directories(ts_utils PRIVATE ${Boost_INCLUDE_DIRS})
+target_link_libraries(ts_utils ${FOLLY_LIBRARIES} ${CMAKE_DL_LIBS} ${Boost_LIBRARIES})
```
lxning, line 13: why boost is needed here?

cpp/src/utils/metrics/dimension.hh
```diff
@@ -0,0 +1,19 @@
...
+    class Dimension {
+        public:
+        Dimension(const std::string& name, const std::string& value);
+        const std::string GetName() const;
+        const std::string GetValue() const;
+
+        private:
+        const std::string name;
+        const std::string value;
+    };
```
lxning, line 16: please address my[ previous comments](https://github.com/pytorch/serve/pull/1874#discussion_r983866145).

cpp/src/utils/metrics/metric.hh
```diff
@@ -0,0 +1,50 @@
...
+        uint64_t GetTimestamp() const;
+        const std::string GetRequestId() const;
+
+        private:
+        const double value;
+        const uint64_t timestamp;
+        const std::string request_id;
+    };
+
+    class Metric {
```
lxning, line 29: As we sync up last week, we need first define torchserve Metric API so that we are able to  extend it to different implementation such as frontend prometheus and backend log metrics.

Here the code is more related to backend. it is difficult to later extend to frontend. So please define metric api first, and then implement backend metrics.

cpp/src/backends/CMakeLists.txt
```diff
@@ -27,28 +27,28 @@ set(TS_BACKENDS_TORCH_SCRIPTED_SOURCE_FILES "")
...
+target_include_directories(ts_backends_torch_scripted PUBLIC
   ${TS_BACKENDS_TORCH_SCRIPTED_SRC_DIR} ${TS_BACKENDS_TORCH_SCRIPTED_SRC_DIR}/handler ${TORCH_INCLUDE_DIRS})
-target_link_libraries(ts_backends_torch_scripted PUBLIC ts_utils ts_backends_core ${TORCH_LIBRARIES}) 
+target_link_libraries(ts_backends_torch_scripted PUBLIC ts_utils ts_backends_core ${TORCH_LIBRARIES})
 install(TARGETS ts_backends_torch_scripted DESTINATION ${torchserve_cpp_SOURCE_DIR}/_build/libs)
 
 # build library ts_backend_torch_deploy
 #set(TS_BACKENDS_TORCH_DEPLOY_SOURCE_FILES "")
 #add_library(ts_backends_torch_deploy SHARED ${TS_BACKENDS_TORCH_DEPLOY_SOURCE_FILES})
 #target_include_directories(ts_backends_torch_deploy PUBLIC ${TS_BACKENDS_TORCH_DEPLOY_SRC_DIR})
```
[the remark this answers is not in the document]
namannandan: I have rebased my dev branch and removed all code changes to this file, but since I had touched this file in a previous commit, the [pre-commit hook](https://github.com/pytorch/serve/blob/be283977488ea186ac631634b2ed6a40f72a227d/.pre-commit-config.yaml#L17) to remove trailing whitespaces is still running on this file.

cpp/src/utils/CMakeLists.txt
namannandan: Good catch. Boost is not required in this PR, so I'll remove it. I had initially included it to fetch the `hostname` in a platform independent manner while generating the metic log. Will consider including it if required in a subsequent PR when implementing the `STDOUTMetricEmitter` class.

cpp/src/utils/metrics/dimension.hh
namannandan: I do agree with your comment that a `Dimension` has a name and can take on multiple values. The intention of defining this class was to use it in the context of a particular metric instance. For example, if we have a metric that tracks the inference latency for a fastrcnn model with batch size 2, then the dimensions for that specific metric instance could be something like `{"model_name": "fastrcnn", "batch_size": "2"}`. Example usage in [testcase](https://github.com/pytorch/serve/pull/1874/files#diff-f399ff4bf1dde7c7aa221ae2e2441f29aa62b6c1f41f3938f5afb7744350ead7R43). Also, I believe that we will not know all the possible dimension values during initialization time, since we don't enforce that in the metrics configuration. Please let me know your thoughts on this.

cpp/src/utils/metrics/metric.hh
namannandan: I believe we are on the same page. The class I'm defining here is going to be used to store the metric attributes `(name, unit, type, dimensions and values)` for a single instance of a particular metric. The plan is to add a `MetricCache` class which will hold several of these `Metric` objects. The modularity to emit metrics either as a log or to prometheus or a different metric framework in future will come from the `MetricEmitter` handle in the `MetricCache` instance. For example, the backend `MetricCache` instance will use the `STDOUTMetricEmitter` and the frontend `MetricCache` instance will use the `PrometheusMetricEmitter`. Please let me know let me know your thoughts on this.

cpp/src/utils/metrics/dimension.hh
lxning: It is impossible to know all the possible values of a dimension. As an API, the data structure should be able to reflect this use case, eg. vector.

cpp/src/utils/metrics/metric.hh
lxning: sounds good.

cpp/src/utils/metrics/metric.hh
```diff
@@ -0,0 +1,50 @@
...
+        const std::vector<MetricValue> GetValues() const;
+        void Update(const double& value, const std::string& request_id = "");
+        void Reset();
+
+        private:
+        const std::string name;
+        const std::string unit;
+        const MetricType type;
+        const std::vector<Dimension> dimensions;
+        std::vector<MetricValue> values;
```
lxning, line 46: The data structure can be simplified and also prepared for sample data point fast updating, eg hashmap.

cpp/src/utils/metrics/dimension.hh
namannandan: That's true, for storing dimension names and values within a `Metric` object, the `Dimension` object may not be absolutely necessary. I was also considering its use in the context of supporting the `MetricCache` API, i.e `Metric addMetric(metricName, unit, dimensions, metricType);` and `Metric getMetric(metricName, dimensions);`. In both of these cases, the `dimensions` argument will need to be a vector of `<dimension_name:single_dimension_value>` tuples. So here, it may be clear to the user, if we express the `dimensions` input argument to be of type `std::vector<Dimension>`. What do you think?

cpp/src/utils/metrics/metric.hh
namannandan: Agree that it may be more efficient to use a hashmap to store the metric values, but a couple of quick questions on the suggested data structure `ConcurrentHashMap<Vector, Sample>` to be used inside the `Metric` object:

1. Since `Vector` is mutable, is it valid to use it as a key in a hashmap?
2. If we plan to store the `Sample` value associated with each possible combination of dimension values in a single `Metric` object, how would we support the `Metric` API `void Update(value);`? Everytime `Update` is called, the caller will need to supply the dimension values as well otherwise we would not know which `Sample` to update. Instead, if we store the value associated with a single metric with specific combination of dimension values in a single `Metric` object, wouldn't that make the implementation simpler to support the `MetricCache` APIs `Metric addMetric(metricName, unit, dimensions, metricType);`, `Metric getMetric(metricName, dimensions);` and also the `Metric` API `void Update(value);`? Although the tradeoff here is that we will use more space by duplicating metric attributes across `Metric` objects.

cpp/src/utils/CMakeLists.txt
namannandan: On second thought, Boost is required in this PR since I've included implementation for the `Emitter`.

cpp/src/utils/metrics/metric.hh
namannandan: Updated implementation to follow the API defined in metrics redesign RFC: https://github.com/pytorch/serve/issues/1492

cpp/src/utils/metrics/metric_cache.hh
```diff
@@ -0,0 +1,19 @@
...
+
+#include <string>
+#include <map>
+
+#include "src/utils/metrics/metric.hh"
+
+namespace torchserve {
+    class MetricCache {
+        public:
+        virtual Metric AddMetric(const std::string& name, const std::string& unit,
```
namannandan, line 12: `LogMetricCache` implementation for backed which will derive from `MetricCache` will support the `AddMetric` and `GetMetric` API by:

1. Storing allowed metric names and corresponding unit and dimension names obtained from the metrics configuration.
2. Storing a map of type `std::map<std::string, LogMetric>`. Here, the key will be a string consisting of metric name, unit, dimension name and value corresponding to each dimension.

rohithkrn: are we going to have other implementations other than `LogMetricCache` in the future?

cpp/src/utils/metrics/log_metric.hh
```diff
@@ -0,0 +1,33 @@
...
+#include <tuple>
+#include <vector>
+#include <map>
+
+#include "src/utils/metrics/metric.hh"
+#include "src/utils/metrics/emitter.hh"
+
+namespace torchserve {
+    // Value entry for log metric consisting of <metric_value, timestamp, request_id>
+    typedef std::tuple<double, std::uint64_t, std::string> LogMetricValue;
```
rohithkrn, line 14: nit: `using LogMetricValue = std::tuple<double, std::uint64_t, std::string>` is more modern?

cpp/src/utils/metrics/metric_cache.hh
namannandan: Yes, we can have a `PrometheusMetricCache` which the frontend can use to store references to prometheus client metric objects.

cpp/src/utils/metrics/log_metric.hh
namannandan: Good point. Will update.

namannandan:
Discussion on design updates and implementation details are tracked in this [document](https://docs.google.com/document/d/1DrInbQx7SFP9YG2xPS7Ca6s6qV5hf9_cjm7wic5RZP0/edit?usp=sharing).

cpp/src/utils/metrics/log_metric.cc
```diff
@@ -0,0 +1,89 @@
...
+                }
+            }
+            std::string error_message = "Provide dimension values corresponding to dimension names: "
+                                        + dimension_names_string;
+            throw std::invalid_argument(error_message);
+        }
+    }
+
+    void TSLogMetric::ValidateMetricValue(const double& value) {
+        if(type == MetricType::COUNTER && value < 0.0) {
```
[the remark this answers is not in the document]
namannandan: A counter metric value can only increase or be reset to 0 on restart as per the prometheus metric definition: https://prometheus.io/docs/concepts/metric_types/#counter.
The prometheus client metric implementation also has this check: https://github.com/jupp0r/prometheus-cpp/blob/871e782d57fec2fba1f00e1c4ed44cc1fc4a0f66/core/src/counter.cc#L8

Do we want to include this check here or let the frontend handle it?

cpp/src/utils/metrics/metric.hh
```diff
@@ -0,0 +1,32 @@
...
+        IMetric(const MetricType& type, const std::string& name, const std::string& unit,
+                const std::set<std::string>& dimension_names) :
+                type{type}, name{name}, unit{Units::GetUnitMapping(unit)}, dimension_names{dimension_names} {}
+        virtual void AddOrUpdate(const std::map<std::string, std::string>& dimension_values, const double& value) = 0;
+
+        protected:
+        const MetricType type;
+        const std::string name;
+        const std::string unit;
+        const std::set<std::string> dimension_names;
```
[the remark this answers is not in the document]
namannandan: I believe `std::set` will maintain its elements in sorted order (ascending by default) but may not be the same order as in the metrics configuration file. This will not matter for the backend metrics implementation in this PR since it does not make any assumptions about the order in which the dimension names or values are provided. `AddOrUpdate` method accepts a map consisting of keys as dimension names and values as dimension values.

Do we want the implementation to use the metric dimension order specified in the configuration file?

cpp/src/utils/metrics/log_metric.hh
```diff
@@ -0,0 +1,28 @@
...
+
+#include "src/utils/metrics/metric.hh"
+
+namespace torchserve {
+    class TSLogMetric : public IMetric {
+        public:
+        TSLogMetric(const MetricType& type, const std::string& name, const std::string& unit,
+                    const std::set<std::string>& dimension_names) :
+                    IMetric(type, name, unit, dimension_names) {}
+        void AddOrUpdate(const std::map<std::string, std::string>& dimension_values, const double& value);
```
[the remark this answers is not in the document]
namannandan: Yes, the argument here is a map with key as dimension name and value as dimension value.

Do we want this argument to be a list of dimension values instead, with order assumed from configuration file?

cpp/src/utils/metrics/log_metric.cc
```diff
@@ -0,0 +1,89 @@
...
+                std::chrono::system_clock::now().time_since_epoch()).count();
+
+        if (request_id.empty()) {
+            TS_LOGF(INFO, "[METRICS]{}.{}:{}|#{}|#hostname:{},{}",
+                    name, unit, value, dimension_values_string, hostname, timestamp);
+        }
+        else {
+            TS_LOGF(INFO, "[METRICS]{}.{}:{}|#{}|#hostname:{},{},{}",
+                    name, unit, value, dimension_values_string, hostname, timestamp, request_id);
+        }
```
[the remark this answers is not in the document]
namannandan: This is intended to be used by backend. I included the `hostname` and `timestamp` to align with the current python implementation for backend metrics: https://github.com/pytorch/serve/blob/877c7b6026a2957e10a5565cfc836e305f8799ff/ts/metrics/metric.py#L77
which is invoked here: https://github.com/pytorch/serve/blob/d258da88fb4b1c528b4a5ffdfaa1f2a4ef3fecab/ts/service.py#L140

Do we want to exclude these in the CPP backend metrics implementation?

cpp/src/utils/metrics/metric.hh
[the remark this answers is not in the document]
lxning: The backend should also keep the same order defined in the metrics configuration b/c backend metrics needs to be injected into prometheus metrics by frontend.

cpp/src/utils/metrics/log_metric.hh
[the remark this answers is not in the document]
lxning: The variable name should reflect its meaning. That's my comments.

cpp/src/utils/metrics/log_metric.cc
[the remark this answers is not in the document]
lxning: I see.

cpp/src/utils/metrics/log_metric.cc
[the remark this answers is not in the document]
lxning: ok, then keep the same pattern as the existing python backend,

cpp/src/utils/metrics/log_metric.hh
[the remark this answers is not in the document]
namannandan: Got it, makes sense. Also, considering the fact that we want to retain order in which dimension names are specified in the metrics configuration, the `dimension_values` argument can be a vector of strings corresponding to the dimension names. Will update implementation.

rohithkrn (approved):

cpp/test/utils/metrics/log_metric_test.cc
```diff
@@ -0,0 +1,150 @@
...
+#include <vector>
+
+#include "src/utils/logging.hh"
+
+namespace torchserve {
+class TSLogMetricTest : public ::testing::Test {
+ protected:
+  const std::string logfile_path{"test/resources/logging/test.log"};
+  const std::string logger_config_path_str{
+      "test/resources/logging/log_to_file.config"};
```
lxning, line 19: where is the log_to_file.config?

cpp/test/utils/metrics/log_metric_test.cc
```diff
@@ -0,0 +1,150 @@
...
+
+TEST_F(TSLogMetricTest, TestTSLogMetricEmitWithoutRequestId) {
+  TSLogMetric test_metric(MetricType::COUNTER, metric_name, "ms",
+                          metric_dimension_names);
+  test_metric.AddOrUpdate(metric_dimension_values, 1.5);
+  const std::vector<std::string> metric_logs = GetMetricLogs();
+  const std::string expected_metric_log_pattern =
+      "^.*\\[METRICS\\]test_metric\\.Milliseconds\\:1\\.5\\|#level\\:model,"
+      "model_name\\:test_model\\"
+      "|#hostname\\:.*,[0-9]+$";
```
lxning, line 134: [This](https://github.com/pytorch/serve/blob/master/frontend/server/src/main/java/org/pytorch/serve/metrics/Metric.java#L15) is how frontend parse backend metrics log. Please follow the same regex.

cpp/test/utils/metrics/log_metric_test.cc
```diff
@@ -0,0 +1,150 @@
...
+#include <regex>
+#include <string>
+#include <vector>
+
+#include "src/utils/logging.hh"
+
+namespace torchserve {
+class TSLogMetricTest : public ::testing::Test {
+ protected:
+  const std::string logfile_path{"test/resources/logging/test.log"};
```
lxning, line 17: it is not a good way to write a log in resource. Backend log is stdio log.

cpp/test/utils/metrics/log_metric_test.cc
namannandan: It is here: https://github.com/pytorch/serve/blob/cpp_backend/cpp/test/resources/logging/log_to_file.config
and is used for the `logging` unit tests. I've re-used the same here.

cpp/test/utils/metrics/log_metric_test.cc
namannandan: Will update the log path to `test/utils/metrics/metrics_test.log`

namannandan: Also, I considered the option of capturing the `stdout` for the unit test, but it appears that `gtest` does not offer any public API for capturing test `stdout` from what I could find. Although it does have an internal API for capturing `stdout` which may not be advisable to use in the test code:

1. https://github.com/google/googletest/blob/191ca1f3a9262b90a586ae2c2e8c742c3d867801/googletest/src/gtest-port.cc#L1117
2. https://stackoverflow.com/questions/3803465/how-to-capture-stdout-stderr-with-googletest

Therefore instead, I used log to file and then parse the file to validate the log output.

cpp/test/utils/metrics/log_metric_test.cc
namannandan: Good pint, will update unit tests to use same metric log regex as frontend.

PR #1887, opened 2022-10-04
Bug fix in cpp integration

rohithkrn:
## Description

This change contains fix for minor bugs in java - cpp integration. It also replaces use of glog with folly logging in one of the files.

### DIFF
--- a/frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerLifeCycle.java
+++ b/frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerLifeCycle.java
@@ -94,7 +94,6 @@ public class WorkerLifeCycle {
                 startWorkerCPP(port, "LDP");
                 break;
             default:
-                logger.info("starting python worker, runtime: {}", model.getRuntimeType());
                 startWorkerPython(port);
                 break;
         }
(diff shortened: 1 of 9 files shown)


frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerLifeCycle.java
```diff
@@ -204,9 +204,9 @@ private void startWorkerCPP(int port, String runtimeType)
         }
 
         argl.add(cppBackendBin.getAbsolutePath());
-        argl.add("--sock_type");
+        argl.add("--socket_type");
```
maaquib, line 207: Revert. This is used in python backend. I will update this in the cpp part in my next pr

maaquib (changes requested):

lxning (approved):

maaquib (approved):

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerLifeCycle.java
rohithkrn: done

PR #1890, opened 2022-10-05
cpp backend with Java integ

maaquib:
## Description

Fixes #(issue)

## Type of change

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

## Checklist:

- [ ] Did you have fun?

### DIFF
--- a/cpp/src/backends/process/model_worker.cc
+++ b/cpp/src/backends/process/model_worker.cc
@@ -13,6 +13,7 @@ namespace torchserve {
     const std::string& model_dir) {
     unsigned short socket_family;
     socket_family = AF_INET;
+    socket_type_ = socket_type;
     if (socket_type == "unix") {
       socket_family = AF_UNIX;
       if (socket_name.empty()) {
--- a/cpp/src/backends/process/model_worker_socket.cc
+++ b/cpp/src/backends/process/model_worker_socket.cc
@@ -5,8 +5,8 @@
 #include "src/utils/logging.hh"
 #include "src/backends/process/model_worker.hh"
 
-DEFINE_string(socket_type, "tcp", "socket type");
-DEFINE_string(socket_name, "", "socket name for uds");
+DEFINE_string(sock_type, "tcp", "socket type");
+DEFINE_string(sock_name, "", "socket name for uds");
 DEFINE_string(host, "127.0.0.1", "");
 DEFINE_string(port, "9000", "");
 DEFINE_string(runtime_type, "LSP", "model runtime type");
@@ -22,7 +22,7 @@ int main(int argc, char* argv[]) {
 
   torchserve::SocketServer server = torchserve::SocketServer::GetInstance();
 
-  server.Initialize(FLAGS_socket_type, FLAGS_socket_name, 
+  server.Initialize(FLAGS_sock_type, FLAGS_sock_name,
   FLAGS_host, FLAGS_port, FLAGS_runtime_type, FLAGS_device_type, FLAGS_model_dir);
 
   server.Run();
--- a/cpp/build.sh
+++ b/cpp/build.sh
@@ -136,7 +136,7 @@ function install_folly() {
 function install_kineto() {
   if [ "$PLATFORM" = "Linux" ]; then
     echo -e "${COLOR_GREEN}[ INFO ] Skip install kineto on Linux ${COLOR_OFF}"
-  elif  "$PLATFORM" = "Mac" ]; then
+  elif [ "$PLATFORM" = "Mac" ]; then
     KINETO_SRC_DIR=$BASE_DIR/third-party/kineto
 
     if [ ! -d "$KINETO_SRC_DIR" ] ; then


frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerLifeCycle.java
```diff
@@ -19,6 +21,8 @@
 public class WorkerLifeCycle {
 
     private static final Logger logger = LoggerFactory.getLogger(WorkerLifeCycle.class);
+    private static final Pattern PATTERN =
```
rohithkrn, line 24: Can we give more descriptive name to `PATTERN`?

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerLifeCycle.java
```diff
@@ -335,10 +339,11 @@ public void run() {
                         continue;
                     }
 
-                    if ("Torch worker started.".equals(result)) {
+                    Matcher matcher = PATTERN.matcher(result);
```
rohithkrn, line 342: Can we move `matcher` inside the `if` where it's used?

cpp/src/backends/process/model_worker.cc
```diff
@@ -73,17 +75,18 @@ namespace torchserve {
       sock_addr.sin_port = port_;
       sock_addr.sin_addr.s_addr = inet_addr(socket_name_.c_str());
       srv_sock_address = reinterpret_cast<sockaddr*>(&sock_addr);
+      name_len = sizeof(*srv_sock_address);
```
rohithkrn, line 78: curious why we are using size of `srv_sock_address` here and size of `sock_addr` in unix path? does it matter?

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerLifeCycle.java
maaquib: The `matcher` is used in the `else if` clause and I didn't want to do the matching twice. Hence declared it outside

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerLifeCycle.java
maaquib: Done

cpp/src/backends/process/model_worker.cc
maaquib: In case of `uds` casting to `sockaddr` and using `sizeof(*srv_sock_address)` was truncating the file name to 14 chars. This was not a problem with `tcp`

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerLifeCycle.java
```diff
@@ -335,10 +339,11 @@ public void run() {
...
                     }
 
-                    if ("Torch worker started.".equals(result)) {
+                    Matcher matcher = PID_LOG_PATTERN.matcher(result);
+                    if (result.endsWith("Torch worker started.")) {
                         lifeCycle.setSuccess(true);
-                    } else if (result.startsWith("[PID]")) {
-                        lifeCycle.setPid(Integer.parseInt(result.substring("[PID]".length())));
+                    } else if (matcher.matches()) {
+                        lifeCycle.setPid(Integer.parseInt(matcher.group(1)));
```
lxning, line 346: Why log record ending with "Torch worker started." not work in cpp?

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerLifeCycle.java
```diff
@@ -19,6 +21,8 @@
 public class WorkerLifeCycle {
 
     private static final Logger logger = LoggerFactory.getLogger(WorkerLifeCycle.class);
+    private static final Pattern PID_LOG_PATTERN =
+            Pattern.compile(".*\\[PID\\](\\d+)$");
```
lxning, line 25: Please avoid regex as possible as you can.

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerLifeCycle.java
maaquib: Since we are using folly for backend logging the only out of the box formatter is glog ([doc](https://github.com/facebook/folly/blob/main/folly/logging/docs/LogHandlers.md#formatter)). That adds some additional info to the beginning of the log line which has to be accounted for

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerLifeCycle.java
maaquib: Same response as above

cpp/src/backends/process/model_worker_socket.cc
```diff
@@ -5,8 +5,8 @@
 #include "src/utils/logging.hh"
 #include "src/backends/process/model_worker.hh"
 
-DEFINE_string(socket_type, "tcp", "socket type");
-DEFINE_string(socket_name, "", "socket name for uds");
+DEFINE_string(sock_type, "tcp", "socket type");
```
mreso, line 8: Why are we changing this to the shorter form? (socket -> sock)

cpp/src/backends/process/model_worker.cc
```diff
@@ -1,5 +1,8 @@
 #include "src/backends/process/model_worker.hh"
 
+namespace fs = std::filesystem;
```
rohithkrn, line 3: not sure if we need this, I see this is also used in logging_test.cc, wondering if there's a way to define this at single place.

rohithkrn (approved):

mreso (approved):
LGTM

PR #1940, opened 2022-11-02
Add an example with torchdata and torchserve

PratsBhatt:
## Description
The pull request provides a simple example of using torchdata with torchserve.
It uses MNIST as the dataset and task to be solved.
The current example builds on top of the already provided example of MNIST.

Fixes #(issue)

## Type of change
The current pull request adds an example w.r.t torchdata and torchserve for MNIST model.
It adds an `inference.py` script that takes care of loading the MNIST dataset and do REST calls to torchserver. It adds a new `mnist_handler.py` script which adds a preprocessing step to convert the payload of the REST request to tensor as well as to output a class number once inference request is finished.

The output of the `inference.py` looks as the following.

```bash
2022-11-03 02:10:38.996234 - Model prediction Class 1 True Class: tensor([1], dtype=torch.uint8)
2022-11-03 02:10:38.996327 - Model prediction Class 8 True Class: tensor([3], dtype=torch.uint8)
2022-11-03 02:10:38.996433 - Model prediction Class 4 True Class: tensor([4], dtype=torch.uint8)
2022-11-03 02:10:38.996578 - Model prediction Class 7 True Class: tensor([7], dtype=torch.uint8)
2022-11-03 02:10:38.996699 - Model prediction Class 2 True Class: tensor([2], dtype=torch.uint8)
2022-11-03 02:10:38.996821 - Model prediction Class 3 True Class: tensor([3], dtype=torch.uint8)
2022-11-03 02:10:38.996943 - Model prediction Class 7 True Class: tensor([7], dtype=torch.uint8)
2022-11-03 02:10:38.997034 - Model prediction Class 4 True Class: tensor([4], dtype=torch.uint8)
2022-11-03 02:10:38.997124 - Model prediction Class 5 True Class: tensor([5], dtype=torch.uint8)
2022-11-03 02:10:38.997222 - Model prediction Class 9 True Class: tensor([9], dtype=torch.uint8)
2022-11-03 02:10:38.997305 - Model prediction Class 0 True Class: tensor([0], dtype=torch.uint8)
```

## Checklist:

- [x] Did you have fun?

### DIFF
--- a/examples/image_classifier/mnist/torchdata/inference.py
+++ b/examples/image_classifier/mnist/torchdata/inference.py
@@ -1,0 +1,76 @@
+import asyncio
+import datetime
+
+import aiohttp
+import matplotlib.pyplot as plt
+import numpy as np
+import torch
+from torchvision import datasets, transforms
+
+# Batch_Size to kept 1 and letting torchserver to do batch inference
+BATCH_SIZE = 1
+# Torchserver model URl
+MODEL_URL = "http://127.0.0.1:8080/predictions/mnist"
+# Total number of inference calls to send in a single batch
+BATCH_TOTAL_TASKS = 100
+# Stop sending infrence request after making TOTAL_INFERENCE_CALLS
+TOTAL_INFERENCE_CALLS = BATCH_TOTAL_TASKS * 2
+
+#transform the images to tensor. Normalize the images as well.
+image_transform = transforms.Compose([
+        transforms.ToTensor(),
+        transforms.Normalize((0.1307,), (0.3081,))
+    ])
+
+# Load the dataset and set train flag to False.
+testset = datasets.MNIST('./MNIST_dataset', download=True, train=False, transform=image_transform)
+
+# Creating the dataloader.
+inference_dataset = torch.utils.data.DataLoader(testset, batch_size=BATCH_SIZE, shuffle=True)
+
+# Loop through the dataset
+async def iterate_dataset():
+    aiohttp_client_session = aiohttp.ClientSession()
+    tasks = []
+    total_completed_tasks = 0
+    for batch, true_label in iter(inference_dataset):
+        if len(tasks) <= BATCH_TOTAL_TASKS:
+            tasks.append(
+                send_inference_request(
+                    aiohttp_client_session, batch, true_label
+                )
+            )
+        else:
+            # Once we've TOTAL_TASKS send all inference request in short time to allow
+            # torchserver to do batch inference
+            print(
+                f"Total inference calls sent so far {len(tasks)}."
+            )
+            result = await asyncio.gather(*tasks)
+            if total_completed_tasks > TOTAL_INFERENCE_CALLS:
+                print(
+                    f"Finished sending maximum number of inference calls {TOTAL_INFERENCE_CALLS}. Stopping..."
+                )
+            tasks = []
+            tasks.append(
+                send_inference_request(
+                    aiohttp_client_session, batch, true_label
+                )
+            )
+
+        total_completed_tasks = total_completed_tasks + 1
+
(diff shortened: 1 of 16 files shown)


examples/image_classifier/mnist/torchdata/inference.py
```diff
@@ -0,0 +1,76 @@
...
+image_transform = transforms.Compose([
+        transforms.ToTensor(),
+        transforms.Normalize((0.1307,), (0.3081,))
+    ])
+
+# Load the dataset and set train flag to False.
+testset = datasets.MNIST('./MNIST_dataset', download=True, train=False, transform=image_transform)
+
+# Creating the dataloader.
+inference_dataset = torch.utils.data.DataLoader(testset, batch_size=BATCH_SIZE, shuffle=True)
```
msaroufim, line 29: Overall I think the example looks good, it's integrating a torchvision dataset but it's not quite clearly a torchdata integration

Specifically I was hoping we could create some toy torchdata dataset directly without leveraging torchvision. I believe this change would be minor to your code but if it isn't I'm happy to merge this if you have bandwidth to work on the more vanilla torchdata integration

agunapal:
@PratsBhatt Thanks for taking this up. Overall it looks good, but In this example, we would want to explicitly make use of TorchData features. You could take a look at this example in TorchData and see if you can modify your current example with with. https://github.com/pytorch/data/blob/main/examples/vision/imagefolder.py

examples/image_classifier/mnist/torchdata/inference.py
ejguan: I agree. @NivekT  Wondering do you have existing pipeline that they can take a reference. for vision benchmarking?

NivekT: For DataPipe reference:
1. Here is the [`torchvision` implementation of loading MNIST](https://github.com/pytorch/vision/blob/main/torchvision/prototype/datasets/_builtin/mnist.py) - it might be too complicated; one option is to use that here (similar to `datasets.MNIST`)
2. A standalone, common example is something like this:
```python
dp = FileLister(str(root), masks=[f"archive_{args.archive_size}*.tar"])
dp = dp.shuffle(buffer_size=10000)
dp = FileOpener(dp, mode="b")
dp = TarArchiveLoader(dp, mode="r:")
dp = dp.shuffle(buffer_size=archive_size)
dp = dp.sharding_filter()
dp = dp.map(pil_loader).map(pil_transformation)
# dp = dp.map(tensor_loader).map(tensor_transformation)  # Alternate - convert image to tensor then transform
```

Separately, I think we should use `DataLoader2` instead of the old version in the example. @ejguan WDYT?

PratsBhatt: Thank you @msaroufim , @ejguan and @NivekT for your comments and guidance. I have incorporated the required changes. Looking forward to your feedback. Thank you once again.

msaroufim (approved):
Amazing thank you for the quick turnaround

examples/image_classifier/mnist/torchdata/inference.py
agunapal: Thanks for the quick turnaround @PratsBhatt . Looks good. I am approving it. Minor feedback: Please link the example here since its a few levels deep and might be missed by others.  https://github.com/pytorch/serve/blob/master/examples/README.md

agunapal (approved):
Looks good. 
Minor feedback: Please link the example here since its a few levels deep and might be missed by others.
 https://github.com/pytorch/serve/blob/master/examples/README.md

PratsBhatt:
Thank you @agunapal and @msaroufim , I have implemented the code changes. Looking forward to merging the PR.

PR #1943, opened 2022-11-03
minor doc fix for dev installation

GavinPHR:
## Description

The `getting_started.md` doc should have `--environment=dev` appended to `python ./ts_scripts/install_from_src.py`.
It took me a while to find out why my changes are not being reflected when developing, this small change should make life a bit easier.


msaroufim:
I'm conflicted about this one, on one hand I agree for the same reason but on the other is the getting started really supposed to be for contributors? I was thinking the CONTRIBUTING.md would be a better place for this kind of stuff but I'm certainly open to move things around if you feel like it'd make things more discoverable

GavinPHR:
> I'm conflicted about this one, on one hand I agree for the same reason but on the other is the getting started really supposed to be for contributors? I was thinking the CONTRIBUTING.md would be a better place for this kind of stuff but I'm certainly open to move things around if you feel like it'd make things more discoverable

I agree, `CONTRIBUTING.md` is the better place and it has the correct commands already.
I think it can be nice to remove the `Install TorchServe for development` section from `getting_started.md`, and instead provide a link to `CONTRIBUTING.md`.

msaroufim:
Wanna just change this PR to do that?

docs/getting_started.md
```diff
@@ -7,33 +7,33 @@
     Note: For Conda, Python 3.8 is required to run Torchserve.
 
     #### For Debian Based Systems/ MacOS
-    
+
```
GavinPHR, line 10: pre-commit hook removed all the trailing white spaces

maaquib (approved):

msaroufim (approved):

PR #1958, opened 2022-11-09
Feat: Dali integration

jagadeeshi2i:
## Description

Adds Nvidia DALI preprocessing to Vision Handler for image and video processing.

Fixes #(issue)

## Type of change

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

## Checklist:

- [ ] Did you have fun?


msaroufim:
We're moving away from using environment variables to gate behavior everywhere, my preference would be to check if dali_config.json was passed in as an extra file in archiver to gate import statements. You can import things either in __init__() or initialize(). Otherwise this looks awesome thanks!

jagadeeshi2i:
Thanks for your suggestions @msaroufim. Will add the necessary changes.

examples/nvidia_dali/serialiize_dali_pipeline.py
```diff
@@ -0,0 +1,55 @@
...
+        std=[0.229 * 255, 0.224 * 255, 0.225 * 255],
+    )
+    return normalized
+
+
+def main(filename):
+    config_file = open(args.config)
+    config = json.load(config_file)
+    batch_size = config["batch_size"]
+    num_threads = config["num_threads"]
```
msaroufim, line 44: What's a good way to set this? Since each pipeline will run in a worker.

Also should pipelines be shared across workers?

examples/nvidia_dali/serialiize_dali_pipeline.py
```diff
@@ -0,0 +1,55 @@
...
+        mean=[0.485 * 255, 0.456 * 255, 0.406 * 255],
+        std=[0.229 * 255, 0.224 * 255, 0.225 * 255],
+    )
+    return normalized
+
+
+def main(filename):
+    config_file = open(args.config)
+    config = json.load(config_file)
+    batch_size = config["batch_size"]
```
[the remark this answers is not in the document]
jagadeeshi2i: Here the batch_size is the max batch_size so this will for for smaller inputs. Yes this should be same as TS batch_size.

examples/nvidia_dali/serialiize_dali_pipeline.py
jagadeeshi2i: The above file serializes the pipeline and only the deserialized file is passed to MAR file along with dali_config.json.

examples/nvidia_dali/README.md
```diff
@@ -0,0 +1,79 @@
...
+mkdir model_store
+mv resnet-18.mar model_store/
+```
+
+### Set environment for DALI
+
+Run the following command in your terminal to set the environment variable for DALI_PREPROCESSING
+
+```bash
+export DALI_PREPROCESSING=true
```
mreso: @msaroufim already mentioned that we are trying to avoid env variables. I might even go a bit further and would say that this could be implemented as a custom handler which inherits from VisionHandler that only overrides preprocessing + initialize. That way we keep Vision handler simple and avoid any gating.

ts/torch_handler/vision_handler.py
```diff
@@ -3,26 +3,87 @@
...
 
+if (
+    "DALI_PREPROCESSING" in os.environ
+    and os.environ["DALI_PREPROCESSING"].lower() == "true"
+):
+    from nvidia.dali.pipeline import Pipeline
+    from nvidia.dali.plugin.pytorch import DALIGenericIterator, LastBatchPolicy
+
 
 class VisionHandler(BaseHandler, ABC):
```
mreso, line 20: As mentioned above, I would prefer a custom handler as dali needs additional packages through a requirements.txt + additional file containing serialized pipeline.

mreso (changes requested):
Great PR! My preference would be a custom handler though. That way the integration would be far more convenient as not env variable needs to be set etc.

agunapal:
Hi @jagadeeshi2i  It will good to have this PR integrated. Can you please create a custom handler as suggested above and re-submit

examples/nvidia_dali/dali_config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "batch_size" : 5,
+  "num_threads" : 2,
+  "device_id" : 0
```
lxning, line 4: Does these parameter work?

jagadeeshi2i: Yes by default they hold negative values and will throw errors like `RuntimeError: Device index must not be negative`

examples/nvidia_dali/custom_handler.py
```diff
@@ -0,0 +1,81 @@
...
+        num_threads = self.dali_configs["num_threads"]
+        device_id = self.dali_configs["device_id"]
+
+        input_byte_arrays = [i["body"] if "body" in i else i["data"] for i in data]
+        for byte_array in input_byte_arrays:
+            np_image = np.frombuffer(byte_array, dtype=np.uint8)
+            batch_tensor.append(np_image)  # we can use numpy
+
+        filename = os.path.join(self.model_dir, self.dali_file[0])
+        pipe = Pipeline.deserialize(filename=filename)
```
mreso, line 61: How long does it take to initialize the pipeline? Could we move this to initialize() to perform this step once and then just store it in self.pipe?

examples/nvidia_dali/custom_handler.py
```diff
@@ -0,0 +1,81 @@
...
+        Args:
+            data (List): Input data from the request is in the form of a Tensor
+
+        Returns:
+            list : The preprocess function returns the input image as a list of float tensors.
+        """
+        batch_tensor = []
+        batch_size = self.dali_configs["batch_size"]
+        num_threads = self.dali_configs["num_threads"]
+        device_id = self.dali_configs["device_id"]
```
mreso, line 53: Would it make sense to couple device id with the device the model is placed or are we expecting to usually run preproc on a different/centralized device?

test/pytest/test_dali_preprocess.py
```diff
@@ -0,0 +1,54 @@
...
+
+    handler.initialize(ctx)
+    data = {}
+
+    with open(Path(CURR_FILE_PATH) / "test_data" / file, "rb") as image:
+        image_file = image.read()
+        byte_array_type = bytearray(image_file)
+        data["body"] = byte_array_type
+
+    x = handler.dali_preprocess([data])
```
mreso, line 49: Can you update the unit test to reflect the cutom handler? Also the model is not downloaded automatically. You can e.g. use a fixture to download it an remove it afterwards.

mreso (approved):
Thanks for this PR! Please address the comments before merging.

examples/nvidia_dali/custom_handler.py
jagadeeshi2i: Yes, we can put pre-processing in a separate GPU.

jagadeeshi2i:
@lxning @chauhang Pls approve the PR.

msaroufim (approved):

PR #1960, opened 2022-11-09
`torch.compile()` support

msaroufim:
Just went ahead and added `torch.compile` support - to make it happen I made a few changes
1. Added a tutorial
2. In the base handler check if _dynamo is present and if yes enable pt 2.0
3. If pt 2.0 is enabled and a user passed in a `compile.json` file to specify the backend then model compilation is enabled
4. Enabled some reasonable defaults for mode that reduces overhead - aka cuda graphs
5. For convenience I also changed `install_dependencies.py` to include a mode that installs nightly torch

## Logs

```
2022-12-05T23:56:53,168 [INFO ] W-9000-densenet161_1.0-stdout MODEL_LOG - Listening on port: /tmp/.ts.sock.9000
2022-12-05T23:56:53,172 [INFO ] W-9000-densenet161_1.0-stdout MODEL_LOG - Successfully loaded /home/ubuntu/serve/ts/configs/metrics.yaml.
2022-12-05T23:56:53,172 [INFO ] W-9000-densenet161_1.0-stdout MODEL_LOG - [PID]15049
2022-12-05T23:56:53,172 [INFO ] W-9000-densenet161_1.0-stdout MODEL_LOG - Torch worker started.
2022-12-05T23:56:53,173 [INFO ] W-9000-densenet161_1.0-stdout MODEL_LOG - Python runtime: 3.8.13
2022-12-05T23:56:53,187 [INFO ] W-9000-densenet161_1.0-stdout MODEL_LOG - Connection accepted: /tmp/.ts.sock.9000.
2022-12-05T23:56:53,228 [INFO ] W-9000-densenet161_1.0-stdout MODEL_LOG - model_name: densenet161, batchSize: 1
2022-12-05T23:56:57,592 [INFO ] W-9000-densenet161_1.0-stdout MODEL_LOG - Compiled model with backend inductor
```


examples/pt2/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+As an example let's expand our getting started gui  de with the only difference being passing in the extra `compile.json` file
+
+```
+mkdir model_store
+torch-model-archiver --model-name densenet161 --version 1.0 --model-file ./serve/examples/image_classifier/densenet_161/model.py --export-path model_store --extra-files ./serve/examples/image_classifier/index_to_name.json,./serve/examples/image_classifier/compile.json --handler image_classifier
+torchserve --start --ncs --model-store model_store --models densenet161.mar
+```
+
+The exact same approach works with any other model, what's going on is the below
```
HamidShojanazeri, line 28: @msaroufim Training part may need a short explanation to make it clear it is not related to inference and I am guessing without the optimized training we can still do the optimized inference. Maybe we can just keep it for inference here and blow that you are explaining mention the Training optimizations as well.

ts/torch_handler/base_handler.py
```diff
@@ -24,6 +29,33 @@
...
 
+# Possible values for backend in utils.py
+def check_pt2_enabled():
+    try:
+        import torch._dynamo
+        pt2_enabled = True
+        if torch.cuda.is_available():
+            # If Ampere enable tensor cores and ideally get yourself an A10G or A100
+            if torch.cuda.get_device_capability() >= (8, 0):
+                torch.backends.cuda.matmul.allow_tf32 = True
```
HamidShojanazeri, line 40: Adding comment to explain why its needed

ts/torch_handler/base_handler.py
```diff
@@ -138,6 +171,21 @@ def initialize(self, context):
...
 
+        self.model.eval()
+        optimization_config = os.path.join(model_dir, "compile.json")
+        backend = load_compiler_config(optimization_config)
+
+        # PT 2.0 support is opt in
+        if check_pt2_enabled() and backend:
+            # Compilation will delay your model initialization
+            try:
+                self.model = torch.compile(self.model, backend=backend)
```
HamidShojanazeri, line 182: QQ, we are assuring the numerical stability so user wouldn't expect any accuracy drop/ difference?

ts/torch_handler/base_handler.py
msaroufim: I thought this would be sufficient `# If Ampere enable tensor cores and ideally get yourself an A10G or A100` - lemme know what's not clear

ts/torch_handler/base_handler.py
msaroufim: Good question, in general we do guarantee it but there will be small numerical differences but we have tools to debug https://pytorch.org/docs/master/dynamo/troubleshooting.html#accuracy-debugging

examples/pt2/README.md
```diff
@@ -0,0 +1,49 @@
+## PyTorch 2.x integration
+
+PyTorch 2.0 brings more compiler options to PyTorch, for you that should mean better perf either in the form of lower latency or lower memory consumption. Integrating PyTorch 2.0 is fairly trivial but for now the support will be experimental until the PyTorch 1.14 release.
```
mreso, line 3: Maybe: ...until the official release. 
As there will technically be no 1.14 release?

examples/pt2/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+# 5. Compile the module and then run inferences with it
+opt_mod = torch.compile(mod)
+```
+
+torchserve takes care of 4 and 5 for you while the remaining steps are your responsibility. You can do the exact same thing on the vast majority of TIMM or HuggingFace models.
+
+## Next steps
+
+For now PyTorch 2.0 has mostly been focused on accelerating training so production grade applications should instead opt for TensorRT for accelerated inference performance which is also natively supported in torchserve. We just wanted to make it really easy for users to experiment with the PyTorch 2.0 stack. You can learn more here https://github.com/pytorch/serve/blob/master/docs/performance_guide.md
```
mreso, line 49: Should we mention the combination of TorchDynamo + TRT backend here? That way the model does not need to be a full graph but would still use TRT for speed?

ts/utils/util.py
```diff
@@ -38,6 +53,26 @@ def list_classes_from_module(module, parent_class=None):
     return classes
 
 
+def load_compiler_config(config_file_path):
+    """
+    Load a compiler {compiler_name -> compiler }
+    Can be extended to also support kwargs for ONNX and TensorRT
+    """
+    if not os.path.isfile(config_file_path):
+        logger.warning(f"{config_file_path} is missing. PT 2.0 will not be used")
```
mreso, line 62: Should we use info here as users who do not provide the file might be confused by the warning?

examples/pt2/README.md
msaroufim: Yeah lemme make that clearer

ts_scripts/install_dependencies.py
```diff
@@ -55,6 +55,10 @@ def install_python_packages(self, cuda_version, requirements_file_path):
         # developer.txt also installs packages from common.txt
         os.system(f"{sys.executable} -m pip install -U -r {requirements_file_path}")
         # If conda is available install conda-build package
+        if nightly:
```
mreso, line 58: Can we if-else this with line 53? Otherwise, we install torch and then reinstall as nightlies.

examples/pt2/README.md
msaroufim: Maybe? At a high level that makes sense but I've heard anecdotally that the TRT backend when not outputting full graphs isn't all that great. To turn that anecdote into something more concrete I'd need to run some benchmarks which I don't have all that much time for. I can omit this entire section if you think it's too confusing but my heuristic was that TensorRT is a more stable alternative.

ts_scripts/install_dependencies.py
msaroufim: Not sure I can unfortunately without significantly changing how our requirements.txt is setup - a lot of them explicitly pick a torch version

examples/pt2/README.md
mreso: Yeah, in that case it's fine as is. Without proper benchmarks we should not recommend, even though it would enable folks who have no full graph to run TRT.

ts_scripts/install_dependencies.py
mreso: I see, remember we had this discussion before.... then maybe just clarify in a comment why and vow to improve this in the future and then see you in clowntown! :-D

ts/torch_handler/base_handler.py
msaroufim: Added more detail

HamidShojanazeri (approved):

lxning (approved):

PR #1963, opened 2022-11-09
Changed version number for release

agunapal:
## Description

Updating version number for release

NA


workflow-archiver/workflow_archiver/version.txt
```diff
@@ -1 +1 @@
-0.2.4
+0.2.5
```
msaroufim, line 1: Should we just change this to 0.6.1 it's very confusing to me why workflow archiver has a different version from all the other packages

agunapal: I agree with you @msaroufim .  @lxning Is that fine?

lxning (approved):

workflow-archiver/workflow_archiver/version.txt
lxning: we need completely redesign workflow soon. So I don't it is necessary to change at this moment.

Issue #1959, opened 2022-11-09
Support py3.10 torchserve binaries


rohithkrn:
### 🚀 The feature

Release py3.10 binaries for torchserve and model-archiver

### Motivation, pitch

There's a strong customer demand for py3.10 [DLAMI](https://docs.aws.amazon.com/dlami/latest/devguide/what-is-dlami.html) which consumes torchserve packages.

msaroufim:
Do we really need libgit? Seems like we introduced it primarily here https://github.com/pytorch/serve/blob/master/ts_scripts/install_from_src.py#L34 to validate a source installation but we can also just print the output of `ts.__version__`

PR #1999, opened 2022-11-23
Update dependencies for python3.10

rohithkrn:
## Description

Update dependencies for python3.10 support.

Had to install rust in dependencies since tokenizers pip wheel would fail without it.
```
copying py_src/tokenizers/processors/init.pyi -> build/lib.macosx-12-arm64-cpython-39/tokenizers/processors
copying py_src/tokenizers/trainers/init.pyi -> build/lib.macosx-12-arm64-cpython-39/tokenizers/trainers
copying py_src/tokenizers/tools/visualizer-styles.css -> build/lib.macosx-12-arm64-cpython-39/tokenizers/tools
running build_ext
running build_rust
error: can't find Rust compiler
```

Issue: https://github.com/pytorch/serve/issues/1959

### DIFF
--- a/ts_scripts/install_dependencies.py
+++ b/ts_scripts/install_dependencies.py
@@ -68,7 +68,9 @@ class Common:
         pass
 
     def install_rust(self):
-        os.system("curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh")
+        os.system(
+            "curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y"
+        )
 
 
 class Linux(Common):
--- a/ts_scripts/install_dependencies.py
+++ b/ts_scripts/install_dependencies.py
@@ -67,6 +67,9 @@ class Common:
     def install_wget(self):
         pass
 
+    def install_rust(self):
+        os.system("curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh")
+
 
 class Linux(Common):
     def __init__(self):
@@ -148,6 +151,7 @@ def install_dependencies(cuda_version=None):
         system.install_wget()
         system.install_nodejs()
         system.install_node_packages()
+        system.install_rust()
 
     if platform.system() == "Linux" and args.environment == "dev":
         system.install_libgit2()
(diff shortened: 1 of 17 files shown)


ts_scripts/install_dependencies.py
```diff
@@ -148,6 +153,7 @@ def install_dependencies(cuda_version=None):
         system.install_wget()
         system.install_nodejs()
         system.install_node_packages()
+        system.install_rust()
```
msaroufim, line 156: why do we need rust installed and how big is this dependency?

lxning (commented):
could you please add test log?

lxning:
I tried this PR on ubuntu, it does not work.

```
git diff ts_scripts/install_dependencies.py
diff --git a/ts_scripts/install_dependencies.py b/ts_scripts/install_dependencies.py
index 1b0484c4..2fdc6caa 100644
--- a/ts_scripts/install_dependencies.py
+++ b/ts_scripts/install_dependencies.py
@@ -67,6 +67,10 @@ class Common:
     def install_wget(self):
         pass

+    def install_rust(self):
+        os.system(
+            "curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y"
+        )

 class Linux(Common):
     def __init__(self):
@@ -148,6 +152,7 @@ def install_dependencies(cuda_version=None):
         system.install_wget()
         system.install_nodejs()
         system.install_node_packages()
+        system.install_rust()

     if platform.system() == "Linux" and args.environment == "dev":
         system.install_libgit2()
```
python ts_scripts/install_dependencies.py --environment dev
...
 If you are using an outdated pip version, it is possible a prebuilt wheel is available for this package but pip is not able to install from it. Installing from the wheel would avoid the need for a Rust compiler.

      To update pip, run:

          pip install --upgrade pip

      and then retry package installation.

      If you did intend to build this package from source, try installing a Rust compiler from your system package manager and ensure it is on the PATH during installation. Alternatively, rustup (available at https://rustup.rs) is the recommended way to download and update the Rust compiler toolchain.
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for tokenizers

agunapal:
I was looking into this. transformers version 4.11.0 doesn't work with Python 3.10, Do we have a dependency for using the version 4.11.0 of transformers  @HamidShojanazeri @msaroufim

agunapal:
I tried upgrading pip. That didnt work.

lxning:
Py310 is fixed in https://github.com/pytorch/serve/pull/2031. so close this one.

PR #2002, opened 2022-11-23
[WIP] Better transformer HF example

HamidShojanazeri:
## Description
This PR will add the Better Transformer optimization to the current HF example. This optimization does not change anything in the model just map the layer names from HF to match expected names in NN.Transformer module where we have faster kernels.

Fixes #(issue)

## Type of change

- [ x] New feature (non-breaking change which adds functionality)

This requires PT 1.13, will upload the regression test logs.

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

### DIFF
--- a/examples/Huggingface_Transformers/Transformer_handler_generalized.py
+++ b/examples/Huggingface_Transformers/Transformer_handler_generalized.py
@@ -64,7 +64,6 @@ class TransformersSeqClassifierHandler(BaseHandler, ABC):
                 model_dir, "libpyt_fastertransformer.so"
             )
             torch.classes.load_library(faster_transformer_complied_path)
-
         # Loading the model and tokenizer from checkpoint and config files based on the user's choice of mode
         # further setup config can be added.
         if self.setup_config["save_mode"] == "torchscript":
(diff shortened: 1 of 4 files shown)


msaroufim (changes requested):
Looks good, left some minor nits. Also please run the lint and spellchecking jobs before checking this in

examples/Huggingface_Transformers/Transformer_handler_generalized.py
```diff
@@ -82,6 +81,23 @@ def initialize(self, ctx):
...
+            #Using the Better Transformer integration to speedup the inference
+            if self.setup_config["BetterTransformer"]:
+                try:
+                    from optimum.bettertransformer import BetterTransformer
+                    is_optimum_available = True
+                except ImportError as error:
+                logger.warning(
+                    "HuggingFace Optimum is not installed. Proceeding without BetterTransformer"
+                )
+                if is_optimum_available:
```
msaroufim, line 93: This works but I think it's better to avoid deep nesting like this, it's more common to do 

```
try: 

except Error1:

except Error2:
```

or 

```
try:

except Error1, Error2:
```

msaroufim (changes requested):

msaroufim (approved):

examples/Huggingface_Transformers/setup_config.json
```diff
@@ -8,5 +8,6 @@
  "captum_explanation":true,
  "embedding_name": "bert",
  "FasterTransformer":false,
+ "BetterTransformer":true,
```
msaroufim, line 11: if model conversion not guaranteed to succeed maybe default false is better

msaroufim (approved):
Assuming all CI is green we're good to go

lxning (approved):

Issue #2063, opened 2023-01-09
Remove dependency on future


AndreasBergmeier6176:
### 🐛 Describe the bug

There is an advisory due to the future package: https://github.com/advisories/GHSA-v3c5-jqr6-7qm8
It seems to me like the _future_ package serves no purpose anymore. See discussion in https://github.com/PythonCharmers/python-future/issues/612.
Is it possible to solve the security problem by just removing the dependency on future and doing a new serve release?

In our project executing `poetry show --tree --why future` we get the output:
```
torchserve 0.6.1 TorchServe is a tool for serving neural net models for inference
└── future *
```

### Error logs

See above links.

### Installation instructions

poetry

### Model Packaing

NA

### config.properties

NA

### Versions

0.6.1

### Repro instructions

NA

### Possible Solution

Perhaps remove `future` line from dependencies.

msaroufim:
Yeah this makes sense, we are doing a security centric release next so this is timely

PR #2082, opened 2023-01-19
Deprecate future package and drop Python2 support

namannandan:
## Description

Remove dependence on the `future` package and references to `Python2` in order to resolve security vulnerabilities.

Unregister `resnet-18` model after testing load from snapshot in sanity test to avoid subsequent sanity runs from failing.

Fixes #([2063](https://github.com/pytorch/serve/issues/2063))

## Type of change

- [ ] Sanity test
CPU: [cpu_sanity_log.txt](https://github.com/pytorch/serve/files/10484508/cpu_sanity_log.txt)
GPU: [gpu_sanity_log.txt](https://github.com/pytorch/serve/files/10484510/gpu_sanity_log.txt)
- [ ] Regression test
CPU: [cpu_regression_log.txt](https://github.com/pytorch/serve/files/10484511/cpu_regression_log.txt)
GPU: [gpu_regression_log.txt](https://github.com/pytorch/serve/files/10484514/gpu_regression_log.txt)
- [ ] Benchmark
CPU: [cpu_benchmark_log.txt](https://github.com/pytorch/serve/files/10484517/cpu_benchmark_log.txt)
GPU: [gpu_benchmark_log.txt](https://github.com/pytorch/serve/files/10484519/gpu_benchmark_log.txt)
- [ ] Conda binary testing
[Build and install torchserve conda binaries](https://github.com/pytorch/serve/blob/master/binaries/README.md) resulting in the following conda environment:
```
(ts-conda-test) $ which torchserve
/home/ubuntu/anaconda3/envs/ts-conda-test/bin/torchserve
(ts-conda-test) $ which torch-model-archiver
/home/ubuntu/anaconda3/envs/ts-conda-test/bin/torch-model-archiver
(ts-conda-test) $ which torch-workflow-archiver
/home/ubuntu/anaconda3/envs/ts-conda-test/bin/torch-workflow-archiver
(ts-conda-test) $ pip freeze | grep future
(ts-conda-test) $ python torchserve_sanity.py &> conda_bin_sanity_log.txt
```
CPU: [cpu_conda_bin_sanity_log.txt](https://github.com/pytorch/serve/files/10484520/cpu_conda_bin_sanity_log.txt)
GPU: [gpu_conda_bin_sanity_log.txt](https://github.com/pytorch/serve/files/10484522/gpu_conda_bin_sanity_log.txt)

The conda packages were verified by running the `test_torchserve`, `test_modelarchiver`, `test_workflow_archiver`, `test_sanity` and `test_workflow_sanity` methods in the [torchserve_sanity](https://github.com/pytorch/serve/blob/master/torchserve_sanity.py) script  

## Checklist:

- [x] Did you have fun?

### DIFF
--- a/binaries/conda/torch-model-archiver/meta.yaml
+++ b/binaries/conda/torch-model-archiver/meta.yaml
@@ -14,7 +14,6 @@ requirements:
     - python
     - pillow
     - psutil
-    - future
 
 build:
   script: "{{ PYTHON }} -m pip install . -vv"
--- a/frontend/archive/src/main/java/org/pytorch/serve/archive/model/Manifest.java
+++ b/frontend/archive/src/main/java/org/pytorch/serve/archive/model/Manifest.java
@@ -127,8 +127,6 @@ public class Manifest {
     public enum RuntimeType {
         @SerializedName("python")
         PYTHON("python"),
-        @SerializedName("python2")
-        PYTHON2("python2"),
         @SerializedName("python3")
         PYTHON3("python3");
 
--- a/model-archiver/README.md
+++ b/model-archiver/README.md
@@ -58,7 +58,7 @@ torch-model-archiver --model-name densenet161 --version 1.0 --model-file example
 $ torch-model-archiver -h
 usage: torch-model-archiver [-h] --model-name MODEL_NAME  --version MODEL_VERSION_NUMBER
                       --model-file MODEL_FILE_PATH --serialized-file MODEL_SERIALIZED_PATH
-                      --handler HANDLER [--runtime {python,python2,python3}]
+                      --handler HANDLER [--runtime {python,python3}]
                       [--export-path EXPORT_PATH] [-f] [--requirements-file]
 
 Model Archiver Tool
@@ -83,11 +83,11 @@ optional arguments:
                         file path to handle custom TorchServe inference logic.
   --extra-files EXTRA_FILES
                         Comma separated path to extra dependency files.
-  --runtime {python,python2,python3}
+  --runtime {python,python3}
                         The runtime specifies which language to run your
                         inference code on. The default runtime is
                         RuntimeType.PYTHON. At the present moment we support
-                        the following runtimes python, python2, python3
+                        the following runtimes python, python3
   --export-path EXPORT_PATH
                         Path where the exported .mar file will be saved. This
                         is an optional parameter. If --export-path is not
--- a/ts_scripts/sanity_utils.py
+++ b/ts_scripts/sanity_utils.py
@@ -1,16 +1,17 @@
+import glob
 import os
 import sys
-import nvgpu
-import glob
-from ts_scripts import marsgen as mg
 
+import nvgpu
+
+from ts_scripts import marsgen as mg
 
 REPO_ROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..")
 sys.path.append(REPO_ROOT)
 
 from ts_scripts import tsutils as ts
-from ts_scripts.tsutils import generate_grpc_client_stubs
 from ts_scripts import utils
+from ts_scripts.tsutils import generate_grpc_client_stubs
 
 
 def run_markdown_link_checker():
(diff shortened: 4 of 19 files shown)


ts_scripts/sanity_utils.py
```diff
@@ -184,6 +184,13 @@ def test_sanity():
         print(f"## Failed to run inference on {resnet18_model['name']} model")
         sys.exit(1)
 
+    response = ts.unregister_model(resnet18_model["name"])
```
msaroufim, line 187: are changes to this file related to this PR?

binaries/conda/torch-model-archiver/meta.yaml
```diff
@@ -14,7 +14,6 @@ requirements:
     - python
     - pillow
     - psutil
-    - future
```
msaroufim, line 17: can you test that the conda binaries still work by creating them? The pypi ones are tested in CI just not these https://github.com/pytorch/serve/tree/master/binaries/conda

msaroufim (changes requested):

ts_scripts/sanity_utils.py
namannandan: I observed that if I run sanity tests back to back, the registration for `resnet-18` version `1` fails because the model already exits. In the sanity test logic, we clean up all other models, so I added this segment to unregister the `resnet-18` model as well after testing load from snapshot.

binaries/conda/torch-model-archiver/meta.yaml
namannandan: Thanks for pointing that out, will test the conda binaries as well.

ts_scripts/sanity_utils.py
namannandan: Would it be better to include this change in a separate PR? CI should still work without this change too.

msaroufim: Na I think making small improvements as you see them is a good idea, please keep the change

binaries/conda/torch-model-archiver/meta.yaml
namannandan: Added details in the summary about conda package testing.

msaroufim (approved):

lxning:
@namannandan please also add test log  for gpu.

frontend/archive/src/main/java/org/pytorch/serve/archive/model/Manifest.java
```diff
@@ -127,8 +127,6 @@ public void setEnvelope(String envelope) {
     public enum RuntimeType {
         @SerializedName("python")
```
agunapal, line 128: @namannandan Could you please clarify if python here  and in subsequent files in java means python3?

mreso: Given that we drop support for Python 2, would it make sense to consolidate the python runtimes into one and rely on the configuration to figure out the right executable?
https://github.com/pytorch/serve/blob/2d1bd2649583c5a505840dcb1fbf9f812ced08fd/frontend/server/src/main/java/org/pytorch/serve/util/messages/EnvironmentUtils.java#L63

model-archiver/README.md
```diff
@@ -58,7 +58,7 @@ torch-model-archiver --model-name densenet161 --version 1.0 --model-file example
 $ torch-model-archiver -h
 usage: torch-model-archiver [-h] --model-name MODEL_NAME  --version MODEL_VERSION_NUMBER
                       --model-file MODEL_FILE_PATH --serialized-file MODEL_SERIALIZED_PATH
-                      --handler HANDLER [--runtime {python,python2,python3}]
+                      --handler HANDLER [--runtime {python,python3}]
```
mreso, line 61: For a period of time we could just map python3 to python with a deprecation warning and instruction to select executable through config.

mreso (approved):
Besides the remaining question if we should consolidate runtimes LGTM

frontend/archive/src/main/java/org/pytorch/serve/archive/model/Manifest.java
namannandan: @agunapal here, `python` refers to the system default python binary and would not guarantee that it is aliased to a `python3` binary. That would depend on how the environment in which Torchserve is running has been set up. In spite of this, I think there may be use-cases where this could be helpful. Please see my comment below and let me know your thoughts.

@mreso yes, that makes sense to me.
Although, I was wondering if a use-case with the following environment would benefit from having support for both `python` and `python3` runtime options:
- `python` -> points to the system default python version (say python 3.8)
- `python3` -> points to a more recent release of python3 (say python 3.11)
This could give choice between the different python 3.x releases without having to change what `python` points to. Let me know your thoughts on if this is useful. If not, I can go ahead and remove the reference to `python3` as well.

namannandan:
> @namannandan please also add test log for gpu.

Updated summary to include test logs for GPU.

namannandan:
Will go ahead and merge this PR since it addresses the major concern about removing the `future` package. Further improvements to consolidating the runtime options could me made in a future PR.

frontend/archive/src/main/java/org/pytorch/serve/archive/model/Manifest.java
mreso: @namannandan yeah, I see your point, though it would look like a hacked together solution and not like a feature to me. Making runtimes configurable and extendable would probably be my preference but there is probably little gain for users out of the non-negligible effort.

PR #2129, opened 2023-02-14
[DO NOT MERGE] Benchmark cpu performance with IPEX launcher

min-jean-cho:
Compare nightly benchmark cpu performance with IPEX launcher enabled, expecting to see perf speedup from setting worker thread affinity.

### DIFF
--- a/.github/workflows/benchmark_launcher_cpu.yml
+++ b/.github/workflows/benchmark_launcher_cpu.yml
@@ -39,7 +39,7 @@ jobs:
             pip install intel_extension_for_pytorch
             export TS_CPU_LAUNCHER_ENABLE=true
       - name: Benchmark cpu nightly
-        run: python benchmarks/auto_benchmark.py --input benchmarks/benchmark_config_cpu.yaml --skip true
+        run: python benchmarks/auto_benchmark.py --input benchmarks/benchmark_config_cpu.yaml --skip false
       - name: Save benchmark artifacts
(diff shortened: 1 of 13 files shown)


min-jean-cho:
Hi @lxning @msaroufim, could we try running the nightly benchmark on aws instance to compare the perf with ipex launcher enabled?

msaroufim:
@agunapal do you have some instructions to run the benchmark suite locally

Otherwise was really hoping we could have some way of triggering them in CI by having for example a keyword in the description or a specific tag but that kept breaking the upload job @lxning

min-jean-cho:
I'm mostly interested in seeing the performance speedup on cloud instances (e.g., aws) that customers would use. Because my local machine cpu config is probably different from cloud instances that customers would use (e.g., most cloud instances tend to be non-NUMA, etc).

agunapal:
@msaroufim I suggested to change the benchmark github action to on push. Looks like she has done that, but I dont see the run triggering. Running locally is not helpful as @min-jean-cho wanted to mimic a realistic machine. Also, this would take really long to run on CPU.

min-jean-cho:
By the way, is there a way to cancel workflow? I forgot to remove `export omp_num_threads=1`

msaroufim:
@min-jean-cho just gave you write access, you should be able to cancel your own workflows now or retrigger them

msaroufim:
This is a benchmark only PR so gonna close

Issue #2143, opened 2023-02-17
Automate benchmarking output validation


agunapal:
### 🚀 The feature

Currently, TorchServe runs benchmark scripts on a nightly basis. These scripts generate a report which are uploaded to the cloud and a dashboard is populated.
The verification of the report is a manual process, i.e someone needs to eyeball the graphs and make sure nothing out of the ordinary pops up.

The goal is to address the above issue by running a validation script which will compare the current run output with the baseline output.

### Motivation, pitch

This feature will be addressed by the following tasks

 1. Automate gpu benchmark validation by identifying common set of metrics across CPU & GPU
 2. Automate cpu benchmark validation
 3.  Automate validation of other remaining metrics for both CPU & GPU

PR #2144, opened 2023-02-17
Automate benchmark validation 

agunapal:
## Description

This PR does the following
- Automatic validation of benchmark_nightly github action for the following models and metrics
    - Computes the average value for all the metrics for the past 7 runs based on data collected in benchmark artifacts
    - The default deviation from the average value for all metrics of each model is set to 30%. We can tune this depending on failures we see.
    - If all the metrics for a model from the current run is +-30% of the average value, the model passes benchmark validation
- Introduces `skip-upload` arg for auto_benchmark script
    - This is useful when running experimental benchmark runs. We don't want to be uploading to the dashboard in this case.

This PR validates the following metrics
### Metrics
- throughput
- total_latency_p50
- model_latency_p50
- total_latency_p90
- model_latency_p90
- total_latency_p99
- model_latency_p99
- memory_percentage_mean
- gpu_used_memory_mean
- cpu_percentage_mean
- gpu_percentage_mean

Fixes #([issue](https://github.com/pytorch/serve/issues/2143))

## Type of change

- [x] Error log when validation fails
```
Run python benchmarks/validate_report.py --input-artifacts-dir /tmp/ts_artifacts/inf1_benchmark_validation
Model scripted_mode_bert_neuron_batch_1_w4_b1 successfully validated
Error while validating memory_percentage_mean for model: scripted_mode_bert_neuron_batch_2_w4_b2, Expected value: 7.70,Observed value: 10.40
Error while validating cpu_percentage_mean for model: scripted_mode_bert_neuron_batch_2_w4_b2, Expected value: 51.56,Observed value: 100.00
  File "benchmarks/validate_report.py", line 103, in <module>
    main()
  File "benchmarks/validate_report.py", line 99, in main
    validate_reports(args.input_artifacts_dir, args.input_report_dir)
  File "benchmarks/validate_report.py", line 70, in validate_reports
    raise Exception("Failures in benchmark validation")
Exception: Failures in benchmark validation
```

- [x] Test for skip_upload
 ### Without flag
```
python benchmarks/auto_benchmark.py --input benchmarks/benchmark_config_gpu.yaml --skip true
...
finish benchmark eager_mode_mnist_w4_b1
report.md is generated
execute: aws s3 cp --recursive /tmp/ts_benchmark/ s3://torchserve-benchmark/nightly/2023-2-17/gpu/0.7.1
upload failed: ../../../tmp/ts_benchmark/eager_mode_mnist_w4_b1/ab_report.csv to s3://torchserve-benchmark/nightly/2023-2-17/gpu/0.7.1/eager_mode_mnist_w4_b1/ab_report.csv Unable to locate credentials
upload failed: ../../../tmp/ts_benchmark/eager_mode_mnist_w4_b1/benchmark.tar.gz to s3://torchserve-benchmark/nightly/2023-2-17/gpu/0.7.1/eager_mode_mnist_w4_b1/benchmark.tar.gz Unable to locate credentials
upload failed: ../../../tmp/ts_benchmark/eager_mode_mnist_w4_b1/logs.tar.gz to s3://torchserve-benchmark/nightly/2023-2-17/gpu/0.7.1/eager_mode_mnist_w4_b1/logs.tar.gz Unable to locate credentials
upload failed: ../../../tmp/ts_benchmark/report.md to s3://torchserve-benchmark/nightly/2023-2-17/gpu/0.7.1/report.md Unable to locate credentials
benchmark_serving.sh finished successfully.
```
### With skip_upload flag
```
python benchmarks/auto_benchmark.py --input benchmarks/benchmark_config_gpu.yaml --skip true --skip_upload true
...
finish benchmark eager_mode_mnist_w4_b1
report.md is generated
benchmark_serving.sh finished successfully.
```

## Checklist:

- [ ] Did you have fun?


benchmarks/models_baseline/bert_multi_gpu.yaml
```diff
@@ -0,0 +1,37 @@
+bert:
```
msaroufim, line 1: do we need to have the baseline stored as files in the repo? Github lets us store artifacts which we can easily read. I was hoping we could do something like look at the last report or average of last n reports. This seems like it'll be annoying to maintain

benchmarks/auto_benchmark.py
```diff
@@ -277,14 +278,23 @@ def main():
         action="store",
         help="true: skip torchserve installation. default: true",
     )
+    parser.add_argument(
```
msaroufim, line 281: should this default to true? The official benchmark reports seems like teh exception

msaroufim (changes requested):

benchmarks/models_baseline/bert_multi_gpu.yaml
agunapal: Ya I agree.. making these yaml files was painful... Let me look into github artifacts.

benchmarks/auto_benchmark.py
agunapal: I'm okay either way. The command doesn't error out if upload fails. So, false is fine.

benchmarks/validate_report.py
```diff
@@ -0,0 +1,101 @@
...
+        "--input-artifacts-dir",
+        nargs="?",
+        help="directory where benchmark artifacts have been saved",
+        type=str,
+        default=BENCHMARK_ARTIFACTS_PATH,
+    )
+
+    parser.add_argument(
+        "--input-report-dir",
+        nargs="?",
```
msaroufim, line 90: i dont think this is needed

benchmarks/utils/report.py
```diff
@@ -0,0 +1,65 @@
...
+        self.mode = mode
+
+    def read_csv(self, csv_file):
+        df = pd.read_csv(csv_file)
+        values = df.values.tolist()
+        self._populate_csv(values[0])
+        self._get_mode(csv_file)
+
+    def _populate_csv(self, values):
+        self.properties["throughput"] = values[9]
```
msaroufim, line 48: nit: cleaner to wrap this in a small `dataclass`

benchmarks/utils/report.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class Report:
+    def __init__(self):
+        self.properties = {}
+        self.mode = None
+        self.throughput = 0
+        self.batch_size = 0
+        self.workers = 0
+        self.deviation = ACCEPTABLE_METRIC_DEVIATION
```
msaroufim, line 33: pass the deviation instead to init with default set to 0.3

msaroufim (approved):

benchmarks/models_baseline/bert_multi_gpu.yaml
agunapal: This has been resolved with benchmark artifacts. We compute the average value based on last 7 reports

benchmarks/utils/report.py
agunapal: Great idea. Done

benchmarks/validate_report.py
agunapal: Done

benchmarks/utils/report.py
```diff
@@ -0,0 +1,65 @@
...
+        self.properties["model_latency_p99"] = values[17]
+        self.properties["memory_percentage_mean"] = values[23]
+        self.properties["gpu_used_memory_mean"] = values[26]
+        self.properties["cpu_percentage_mean"] = values[22]
+        self.properties["gpu_percentage_mean"] = values[24]
+
+    def update(self, report):
+        for property in self.properties:
+            self.properties[property] = (
+                self.properties[property] + report.properties[property]
```
namannandan, line 64: Does this method correctly compute the intended average across the available baseline report?
It looks like we take the average between previous average and a report value.
Should we not sum all baseline values for a metric and divide by the number of baseline reports we have?

agunapal: Great catch! You are correct. They are not mathematically equivalent. Thanks! Let me fix it

agunapal: done

benchmarks/utils/report.py
HamidShojanazeri: +1

benchmarks/validate_report.py
```diff
@@ -0,0 +1,106 @@
...
+
+from utils.report import ACCEPTABLE_METRIC_DEVIATION, METRICS_VALIDATED, Report
+from utils.update_artifacts import (
+    BENCHMARK_ARTIFACTS_PATH,
+    BENCHMARK_REPORT_FILE,
+    BENCHMARK_REPORT_PATH,
+)
+
+
+def metric_valid(key, obs_val, exp_val, threshold):
```
HamidShojanazeri, line 12: wondering if these util funcs could have lived under util.report

benchmarks/validate_report.py
```diff
@@ -0,0 +1,106 @@
...
+        lower = True
+    return check_if_within_threshold(exp_val, obs_val, threshold) or (
+        (obs_val < exp_val and lower)
+    )
+
+
+def check_if_within_threshold(value1, value2, threshold):
+    if float(value1) == 0.0:
+        return True
+    return abs((value1 - value2) / float(value1)) <= threshold
```
HamidShojanazeri, line 27: wondering what is our criteria for threshold?

agunapal: This is a good question. I don't have the correct answer for this. Currently, its at 30% for all metrics. Over time, depending on the failures and the reasons for failures ( could be network issues, or the machine itself being slow), we might have to configure a different threshold for different metrics. I believe this would also help us understand why we dont get consistent behavior across all metrics.

benchmarks/validate_report.py
agunapal: Thanks. Good catch. Moved them

benchmarks/utils/report.py
```diff
@@ -0,0 +1,86 @@
...
+        self.properties["model_latency_p90"] = values[16]
+        self.properties["model_latency_p99"] = values[17]
+        self.properties["memory_percentage_mean"] = values[23]
+        self.properties["gpu_used_memory_mean"] = values[26]
+        self.properties["cpu_percentage_mean"] = values[22]
+        self.properties["gpu_percentage_mean"] = values[24]
+
+    def update(self, report):
+        for property in self.properties:
+            # new average= old average * (n-1)/n + new value /n
```
namannandan, line 64: Nit: Although this method computes an average closer to the actual average, I believe it may not be fully accurate. For example, for the values say [1, 2, 3, 4], the average computed by this method would be 2.265625 whereas the actual average is 2.5. Would it be simpler to have two variables, one to track running sum and another to track count, so the accurate average can be computed and stored each time `update` is called?

namannandan (commented):
One minor follow up on average value computation, otherwise LGTM.

benchmarks/utils/report.py
agunapal: simplified the logic to use csv DictReader

benchmarks/utils/report.py
agunapal: Thanks. Simplified the logic to sum the values and then find the mean as an additional step.

namannandan (approved):

Issue #2139, opened 2023-02-16
Frontend Metric Interface and Class 


lxning:
### 🚀 The feature

Frontend defines IMetric as an interface which has two types of implementation.
- TSLogMetric (ie. existing Metric)
- PrometheusMetric: Counter, Gauge, Historgram, Summary

### Motivation, pitch

See details in #1492

PR #2164, opened 2023-02-28
Add frontend metrics to default metrics configuration

namannandan:
## Description

Add frontend metrics collected by [system_metrics.py](https://github.com/pytorch/serve/blob/master/ts/metrics/system_metrics.py) to [default metrics configuration file](https://github.com/pytorch/serve/blob/master/ts/configs/metrics.yaml)

Fixes #2133

PR #2165, opened 2023-03-01
[WIP] frontend IMetrics, LogMetrics and PrometheusMetrics

lxning:
## Description

Fixes #(issue)
#2139 

## Type of change

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

## Checklist:

- [ ] Did you have fun?

### DIFF
--- a/frontend/server/src/main/java/org/pytorch/serve/metrics/IMetrics.java
+++ b/frontend/server/src/main/java/org/pytorch/serve/metrics/IMetrics.java
@@ -1,0 +1,29 @@
+package org.pytorch.serve.metrics;
+
+import java.util.ArrayList;
+
+public abstract class IMetrics {
+    public IMetrics(
+            String metricsType,
+            String metricsName,
+            String unit,
+            ArrayList<String> dimensionNames) {
+        this.metricsType = metricsType;
+        this.metricsName = metricsName;
+        this.unit = unit;
+        this.dimensionNames = dimensionNames;
+    }
+
+    public abstract void emit(
+            ArrayList<String> dimensionValues,
+            double value);
+    public abstract void emit(
+            ArrayList<String> dimensionValues,
+            String requestIds,
+            double value);
+
+    private String metricsType;
+    private String metricsName;
+    private String unit;
+    private ArrayList<String> dimensionNames;
+}
--- a/frontend/server/src/main/java/org/pytorch/serve/metrics/LogMetrics.java
+++ b/frontend/server/src/main/java/org/pytorch/serve/metrics/LogMetrics.java
@@ -1,0 +1,19 @@
+package org.pytorch.serve.metrics;
+
+import java.util.ArrayList;
+
+public class LogMetrics extends  IMetrics{
+    public LogMetrics(String metricsType, String metricsName, String unit, ArrayList<String> dimensionNames) {
+        super(metricsType, metricsName, unit, dimensionNames);
+    }
+
+    @Override
+    public void emit(ArrayList<String> dimensionValues, double value) {
+
+    }
+
+    @Override
+    public void emit(ArrayList<String> dimensionValues, String requestIds, double value) {
+
+    }
+}
--- a/frontend/server/src/main/java/org/pytorch/serve/metrics/format/prometheous/PrometheusCounter.java
+++ b/frontend/server/src/main/java/org/pytorch/serve/metrics/format/prometheous/PrometheusCounter.java
@@ -1,0 +1,24 @@
+package org.pytorch.serve.metrics.format.prometheous;
+
+import io.prometheus.client.Counter;
+import org.pytorch.serve.metrics.IMetrics;
+
+import java.util.ArrayList;
+
+public class PrometheusCounter extends IMetrics {
+    public PrometheusCounter(String metricsType, String metricsName, String unit, ArrayList<String> dimensionNames) {
+        super(metricsType, metricsName, unit, dimensionNames);
+    }
+
+    @Override
+    public void emit(ArrayList<String> dimensionValues, double value) {
+
+    }
+
+    @Override
+    public void emit(ArrayList<String> dimensionValues, String requestIds, double value) {
+
+    }
+
+    private Counter counter;
+}
--- a/frontend/server/src/main/java/org/pytorch/serve/metrics/format/prometheous/PrometheusGauge.java
+++ b/frontend/server/src/main/java/org/pytorch/serve/metrics/format/prometheous/PrometheusGauge.java
@@ -1,0 +1,24 @@
+package org.pytorch.serve.metrics.format.prometheous;
+
+import io.prometheus.client.Gauge;
+import org.pytorch.serve.metrics.IMetrics;
+
+import java.util.ArrayList;
+
+public class PrometheusGauge extends IMetrics {
+    public PrometheusGauge(String metricsType, String metricsName, String unit, ArrayList<String> dimensionNames) {
+        super(metricsType, metricsName, unit, dimensionNames);
+    }
+
+    @Override
+    public void emit(ArrayList<String> dimensionValues, double value) {
+
+    }
+
+    @Override
+    public void emit(ArrayList<String> dimensionValues, String requestIds, double value) {
+
+    }
+
+    private Gauge gauge;
+}


frontend/server/src/main/java/org/pytorch/serve/metrics/format/prometheous/PrometheusMetric.java
```diff
@@ -0,0 +1,31 @@
...
+        }
+    }
+
+    @Override
+    public void addOrUpdate(
+            ArrayList<String> dimensionValues, String hostname, String requestIds, String timestamp, double value) {
+        ArrayList<String> modifiedDimensionValues = new ArrayList<String>(dimensionValues);
+        if (hostname != null && !hostname.isEmpty()) {
+            modifiedDimensionValues.add(hostname);
+        }
```
lxning, line 27: The order of the dimension's name and value should be 1:1 . Here will cause error if the hostname is appended to values for all cases.

It should be only applied to backend use case.

frontend/server/src/main/java/org/pytorch/serve/metrics/format/prometheous/PrometheusMetric.java
```diff
@@ -0,0 +1,37 @@
...
+            double value) {
+        ArrayList<String> modifiedDimensionValues = new ArrayList<String>(dimensionValues);
+        if (this.context == MetricBuilder.MetricContext.BACKEND
+                && hostname != null
+                && !hostname.isEmpty()) {
+            modifiedDimensionValues.add(hostname);
+        }
+
+        this.addOrUpdate(modifiedDimensionValues, value);
+    }
```
lxning, line 36: prometheus doesn't need implement this one

frontend/server/src/main/java/org/pytorch/serve/metrics/MetricBuilder.java
```diff
@@ -0,0 +1,51 @@
+package org.pytorch.serve.metrics;
+
+import java.util.ArrayList;
+import org.pytorch.serve.metrics.format.prometheous.PrometheusCounter;
+import org.pytorch.serve.metrics.format.prometheous.PrometheusGauge;
+import org.pytorch.serve.metrics.format.prometheous.PrometheusHistogram;
+
+public final class MetricBuilder {
```
lxning, line 8: Please make this as singleton

frontend/server/src/main/java/org/pytorch/serve/metrics/format/prometheous/PrometheusMetric.java
```diff
@@ -0,0 +1,37 @@
...
+            String hostname,
+            String requestIds,
+            String timestamp,
+            double value) {
+        ArrayList<String> modifiedDimensionValues = new ArrayList<String>(dimensionValues);
+        if (this.context == MetricBuilder.MetricContext.BACKEND
+                && hostname != null
+                && !hostname.isEmpty()) {
+            modifiedDimensionValues.add(hostname);
+        }
```
lxning, line 33: please remove if condition since only backend metric call this func.

frontend/server/src/main/java/org/pytorch/serve/metrics/MetricBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+                    return new PrometheusHistogram(context, type, name, unit, dimensionNames);
+                default:
+            }
+        } else {
+            return new LogMetric(context, type, name, unit, dimensionNames);
+        }
+        return null;
+    }
+
+    private MetricBuilder() {
```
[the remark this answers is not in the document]
namannandan: This is used to resolve the following warning:
```Utility classes should not have a public or default constructor. [HideUtilityClassConstructor]```

File: frontend/server/src/main/java/org/pytorch/serve/metrics/MetricBuilder.java (hunk 3 above)
Comment (namannandan) [reply frontend/server/src/main/java/org/pytorch/serve/metrics/MetricBuilder.java:8]: There is no use case to instantiate this class so I've skipped the singleton implementation.

File: frontend/server/src/main/java/org/pytorch/serve/metrics/format/prometheous/PrometheusMetric.java [hunk 6]
```diff
@@ -0,0 +1,31 @@
...
+            MetricBuilder.MetricType type,
+            String name,
+            String unit,
+            List<String> dimensionNames) {
+        super(context, type, name, unit, dimensionNames);
+        // Hostname dimension is automatically included for backend metrics
+        if (this.context == MetricBuilder.MetricContext.BACKEND) {
+            this.dimensionNames.add("Hostname");
+        }
+    }
```
lxning, line 20: it is not necessary to check for each backend metric. the metric parser can do it once. eg. if "model_metrics" then directly add hostname in dimensions

frontend/server/src/main/java/org/pytorch/serve/metrics/format/prometheous/PrometheusMetric.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Override
+    public void addOrUpdate(
+            List<String> dimensionValues, String hostname, String requestIds, double value) {
+        // Used for logging backend metrics
+        // The final entry in dimensionValues is expected to be Hostname
+        ArrayList<String> modifiedDimensionValues = new ArrayList<String>(dimensionValues);
+        modifiedDimensionValues.add(hostname);
```
lxning, line 28: The cost is high to create a array each time.

you already add hostname in each model metric dimensions. the parser part can directly add hostname values in dimension values.

frontend/server/src/main/java/org/pytorch/serve/metrics/LogMetric.java
```diff
@@ -0,0 +1,80 @@
...
+
+        boolean first = true;
+        for (int index = 0;
+                index < Math.min(this.dimensionNames.size(), dimensionValues.size());
+                index++) {
+            if (first) {
+                first = false;
+            } else {
+                sb.append(',');
+            }
```
lxning, line 62: check java Collectors.join function to make code elegant

frontend/server/src/main/java/org/pytorch/serve/metrics/LogMetric.java
```diff
@@ -0,0 +1,80 @@
...
+            sb.append(this.dimensionNames.get(index))
+                    .append(':')
+                    .append(dimensionValues.get(index));
+        }
+        if (hostname != null && !hostname.isEmpty()) {
+            sb.append("|#hostname:").append(hostname);
+        }
```
lxning, line 69: Doesn't hostname always exist?

frontend/server/src/main/java/org/pytorch/serve/metrics/MetricBuilder.java
namannandan: Since this is a utility class, we won't instantiate it, so no constructor is required. If no constructor is provided, the default `public` constructor will be added which leads to the following checkstyle error: `Utility classes should not have a public or default constructor. [HideUtilityClassConstructor]`. Therefore, the dummy private constructor is added. It is not intended to be called. Similar approach is used for example in [NettyUtils.java](https://github.com/pytorch/serve/blob/master/frontend/server/src/main/java/org/pytorch/serve/util/NettyUtils.java#L69)

frontend/server/src/main/java/org/pytorch/serve/metrics/LogMetric.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    public void addOrUpdate(List<String> dimensionValues, String requestIds, double value) {
+        // Used for logging backend metrics
+        String metricString = this.buildMetricString(dimensionValues, requestIds, value);
+        loggerModelMetrics.info(metricString);
+    }
+
+    private String buildMetricString(
```
lxning, line 40: can you have two buildMetricString so that if requestid condition can be removed?
- one have requestid => callled by modelMetrics.
- the other one doesn't have requestid => called by tsMetrics.

frontend/server/src/main/java/org/pytorch/serve/metrics/LogMetric.java
```diff
@@ -0,0 +1,76 @@
...
+        List<String> dimensions = new ArrayList<String>();
+        for (int index = 0; index < dimensionsCount; index++) {
+            dimensions.add(this.dimensionNames.get(index) + ":" + dimensionValues.get(index));
+        }
+        metricStringBuilder.append(dimensions.stream().collect(Collectors.joining(",")));
+
+        // The final dimension is expected to be Hostname
+        metricStringBuilder
+                .append("|#hostname:")
+                .append(dimensionValues.get(dimensionValues.size() - 1));
```
lxning, line 62: tsMetrics config already have hostname in dimensionnames. This part only applies to modelMetrics. That's why I'm asking for overload buildMetricString

frontend/server/src/main/java/org/pytorch/serve/metrics/LogMetric.java
```diff
@@ -0,0 +1,76 @@
...
+
+        if (requestIds != null && !requestIds.isEmpty()) {
+            metricStringBuilder.append(",requestID:").append(requestIds);
+        }
+
```
lxning, line 67: Remove this if condition via func overload

frontend/server/src/main/java/org/pytorch/serve/metrics/LogMetric.java
namannandan: I’ve updated the addOrUpdate API such that, one version is without request id and the other version is with request id [here](https://github.com/pytorch/serve/pull/2165/files#diff-560302d2d763e75ca22eb54b14259605d9c60115a4b5a5ccb9e8ef35be5aaa78R20-R22). So in all cases, the last dimension name and dimension value will be Hostname, so we won’t need different logic to handle tsMetrics and modelMetrics

namannandan:
Approving on @lxning behalf based on the above comment.

namannandan (approved):

agunapal (commented):
@lxning Could you please update the appropriate README files for the additional metrics we are collecting. 
Also, from. a user point of view, do they have to enable some config to get these metrics?

namannandan:
@agunapal Currently there aren't any new metrics that are going to be collected. The existing metrics are being migrated to a new implementation that supports two modes, i.e `log` and `prometheus`. A configuration file is used specify all the metrics that will be collected, which was updated in this PR: https://github.com/pytorch/serve/pull/2164
The final PR for the metrics frontend implementation will include necessary documentation for the `prometheus` mode: https://github.com/pytorch/serve/pull/2199

This PR implements the necessary helper classes for the frontend metrics implementation refactoring.

agunapal (approved):

PR #2166, opened 2023-03-01
Enable worker core pinning in CPU nightly benchmark

min-jean-cho:
This PR enables worker core pinning by default for better cpu performance. 

Below is a nightly cpu benchmark performance comparison:
torchserve_nightly_benchmark_throughput
torchserve_nightly_benchmark_latency

`default`: https://github.com/pytorch/serve/actions/runs/4238841923
`OMP_NUM_THREADS=1`: https://github.com/pytorch/serve/actions/runs/4255376669
`core pinning, --use_logical_cores`: https://github.com/pytorch/serve/actions/runs/4320222197


min-jean-cho:
cc. @ashokei @jgong5

msaroufim:
Just retriggered CI - @min-jean-cho I noticed CI failures last week, do you need any help resolving those?

min-jean-cho:
Thanks @msaroufim, let me convert this PR to WIP. The current status of this PR is to address the concerns brought up by @lxning  when deploying this in production. I'll update this when ready, thanks!

ts_scripts/install_dependencies.py
```diff
@@ -107,6 +110,10 @@ def install_libgit2(self):
         os.system(f"cd libgit2-1.3.0 && cmake . && make && sudo make install && cd ..")
         os.system(f"rm -rf libgit2-1.3.0 && rm libgit2-1.3.0.tar.gz")
 
+    def install_numactl(self):
+        if os.system("numactl --show") != 0 or args.force:
+            os.system(f"{self.sudo_cmd}apt-get install -y numactl")
+
```
lxning, line 116: do we need add numactl into docker container for cpu?

msaroufim (approved):
Same question as the one Li asked, otherwise LGTM. Approving because the this isn't breaking the docker CI

In our current dev docker images we do indeed run the install dependencies script https://github.com/pytorch/serve/blob/master/docker/Dockerfile.dev#L70 but since we're moving to using the regular docker images moving forward should also add numactl there - cc @agunapal

min-jean-cho:
We can add `numactl` in docker cpu as well. But I recall in docker, `numactl` required `privileged` mode -- `taskset` is available for in this case.

agunapal (commented):
@min-jean-cho Could you please tell me where numactl is being used here.
Trying to understand why docker CI is not complaining.

min-jean-cho:
> Could you please tell me where numactl is being used here.

@agunapal `numactl` is used internally by the launcher [here](https://github.com/pytorch/serve/pull/2166/files#diff-88e86db18d87ea2dd136cb1078dd4d7b31a94a78ea8033a5c132a4afc0b592b3R120-R139).

> Trying to understand why docker CI is not complaining.

I think because if `cpu_launcher_enable=true` and `numactl` is not available, then enabling launcher just skips. When we upgrade to the next torch 2.x release, if `cpu_launcher_enable=true` and `numactl` is not available, then then launcher automatically tries `taskset` [here](https://github.com/pytorch/pytorch/blob/master/torch/backends/xeon/run_cpu.py#L474-L485) instead of just skipping.

agunapal (approved):

lxning (approved):

PR #2182, opened 2023-03-16
feat: add PyTorch/XLA support  

morgandu:
## Description

This PR is to add PyTorch/XLA support in TorchServe backend base handler.

## Type of change

### DIFF
--- a/ts/torch_handler/base_handler.py
+++ b/ts/torch_handler/base_handler.py
@@ -113,7 +113,6 @@ class BaseHandler(abc.ABC):
             self.map_location = "cuda"
             self.device = torch.device(self.map_location + ":" + str(properties.get("gpu_id")))
         elif TORCHXLA_AVAILABLE:
-            self.map_location = "xla"
             self.device = xm.xla_device()
         else:
             self.map_location = "cpu"
@@ -253,7 +252,8 @@ class BaseHandler(abc.ABC):
         model_class = model_class_definitions[0]
         model = model_class()
         if model_pt_path:
-            state_dict = torch.load(model_pt_path, map_location=self.device)
+            map_location = None if (TORCHXLA_AVAILABLE and self.map_location is None) else self.device
+            state_dict = torch.load(model_pt_path, map_location=map_location)
             model.load_state_dict(state_dict)
         return model
 
--- a/ts/torch_handler/base_handler.py
+++ b/ts/torch_handler/base_handler.py
@@ -25,6 +25,13 @@ if packaging.version.parse(torch.__version__) >= packaging.version.parse("1.8.1"
 else:
     PROFILER_AVAILABLE = False
 
+try:
+    import torch_xla
+    import torch_xla.core.xla_model as xm
+    TORCHXLA_AVAILABLE = True
+except ImportError as error:
+    TORCHXLA_AVAILABLE = False
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


ts/torch_handler/base_handler.py
```diff
@@ -278,6 +303,9 @@ def inference(self, data, *args, **kwargs):
         with torch.no_grad():
             marshalled_data = data.to(self.device)
             results = self.model(marshalled_data, *args, **kwargs)
+            if torch_xla_enabled:
+                xm.mark_step()
```
msaroufim, line 307: not super familiar with xla internals but what does this line do?

ts/torch_handler/base_handler.py
```diff
@@ -59,6 +59,24 @@ def check_pt2_enabled():
         )
 
 
+def check_torch_xla_enabled() -> bool:
```
msaroufim, line 62: @lxning another good candidate for your new config change, it might be possible that a user has xla installed but doesnt want to necessarily comile the model with XLA

msaroufim (changes requested):
Very cool! Left some minor questions on the PR directly

One question I had was whether this is the right way to uses torch/xla nowadays or whether users are recommended to pass in an XLA backend to `torch.compile()`

Since most of machines are running on AWS in CI it's unlikely we'll get a TPU available to fuly test this but I'm assuming this should work just fine on GPU as well, in which case a quick test would also be super helpful

ts/torch_handler/base_handler.py
lxning: @msaroufim yes. the model yaml config can make this much easier

yyetim:
torch.compile() is a good point. I'm guessing we'll need both this version to support pytorch <2.0, and another change to support pytorch 2.0 models.

msaroufim:
So we do actually already support torch.compile https://github.com/pytorch/serve/pull/1960 and you can pass in a custom backend via a `compile.json`

I don't think supporting both workflows is a huge deal but curious which one would you prefer people use?

ts/torch_handler/base_handler.py
morgandu: Removed the `xm.mark_step()` cause this is essential for training, optional for inferencing. In short, the value /calculation is upon either a `xm.mark_step()` or when it gets retrieved. In our case it's the latter one.

morgandu:
> torch.compile() is a good point. I'm guessing we'll need both this version to support pytorch <2.0, and another change to support pytorch 2.0 models.

As discussed, we decided to prioritize pytorch/xla 2.0 and above.

morgandu:
> So we do actually already support torch.compile #1960 and you can pass in a custom backend via a `compile.json`
> 
> I don't think supporting both workflows is a huge deal but curious which one would you prefer people use assuming people have 2.0 installed

added torchxla backend

morgandu:
kokoro-run

msaroufim:
PR looks good but I was hoping we could have the test you're running checked in and only run it if a TPU is found

morgandu:
> PR looks good but I was hoping we could have the test you're running checked in and only run it if a TPU is found

Added test, PTAL

msaroufim (approved):

msaroufim:
LGTM thank you, as FYI we're killing the compile.json in the next release but I'll make the change and test out the kokoro CI directly

morgandu:
> 

Thanks for the heads up!

lxning (approved):

Issue #2180, opened 2023-03-16
TorchServe grpc server side streaming


lxning:
### 🚀 The feature

 TorchServe grpc server side streaming supports

- backend worker continuously send the intermediate prediction response to frontend
- frontend grpc endpoint continuously send the intermediate prediction response from backend to client.

### Motivation, pitch

Usually the predication latency is high (eg. 5sec) for a large model inference. Some models are able to generate intermediate prediction results (eg. generator AI). This feature will send the intermediate prediction results to user once the results are ready. The user will gradually get the entire response. For example, User may get first intermediate response within 1 sec, and gradually get the entire result until 5sec. This feature is to improve user prediction experience.

PR #2186, opened 2023-03-20
Feature/grpc streaming

lxning:
## Description

Fixes #(issue)
#2180 

## Type of change

- Regression test log
[reg.txt](https://github.com/pytorch/serve/files/11021281/reg.txt)

## Checklist:

- [ ] Did you have fun?

### DIFF
--- a/docs/grpc_api.md
+++ b/docs/grpc_api.md
@@ -71,3 +71,28 @@ python ts_scripts/torchserve_grpc_client.py infer densenet161 examples/image_cla
 ```bash
 python ts_scripts/torchserve_grpc_client.py unregister densenet161
 ```
+## GRPC Server Side Streaming
+TorchServe GRPC APS adds a server side streaming of the inference API "StreamPredictions" to allow a sequence of inference responses to be sent over the same GPRC stream. This new API is only recommended for the use case when the inference full response latency is high, and the inference intermediate results are sent to client. This new API automatically forces the batchSize to be one.
+
+```
+service InferenceAPIsService {
+    // Check health status of the TorchServe server.
+    rpc Ping(google.protobuf.Empty) returns (TorchServeHealthResponse) {}
+
+    // Predictions entry point to get inference using default model version.
+    rpc Predictions(PredictionsRequest) returns (PredictionResponse) {}
+
+    // Streaming response for an inference request.
+    rpc StreamPredictions(PredictionsRequest) returns (stream PredictionResponse) {}
+}
+```
+Backend handler calls "send_intermediate_predict_response" to send one intermediate result to frontend, and return the last result as the existing style. For example
+```
+from ts.protocol.otf_message_handler import send_intermediate_predict_response
+
+def handle(data, context):
+    if type(data) is list:
+        for i in range (3):
+            send_intermediate_predict_response(["hello"], context.request_ids, "Intermediate Prediction success", 200, context)
+        return ["hello world "]
+```
--- a/docs/grpc_api.md
+++ b/docs/grpc_api.md
@@ -9,6 +9,7 @@ TorchServe provides following gRPCs apis
 * [Inference API](https://github.com/pytorch/serve/blob/master/frontend/server/src/main/resources/proto/inference.proto)
   - **Ping** : Gets the health status of the running server
   - **Predictions** : Gets predictions from the served model
+  - **StreamPredictions** : Gets server side streaming predictions from the saved model
 
 * [Management API](https://github.com/pytorch/serve/blob/master/frontend/server/src/main/resources/proto/management.proto)
   - **RegisterModel** : Serve a model/model-version on TorchServe
--- a/frontend/server/src/main/java/org/pytorch/serve/wlm/BatchAggregator.java
+++ b/frontend/server/src/main/java/org/pytorch/serve/wlm/BatchAggregator.java
@@ -1,5 +1,7 @@
 package org.pytorch.serve.wlm;
 
+import static org.pytorch.serve.util.messages.RequestInput.TS_STREAM_NEXT;
+
 import java.util.LinkedHashMap;
 import java.util.Map;
 import org.pytorch.serve.job.Job;
@@ -9,6 +11,7 @@ import org.pytorch.serve.util.messages.ModelLoadModelRequest;
 import org.pytorch.serve.util.messages.ModelWorkerResponse;
 import org.pytorch.serve.util.messages.Predictions;
 import org.pytorch.serve.util.messages.RequestInput;
+import org.pytorch.serve.util.messages.WorkerCommands;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -48,6 +51,9 @@ public class BatchAggregator {
                 }
                 return new ModelLoadModelRequest(model, gpuId);
             } else {
+                if (j.getCmd() == WorkerCommands.STREAMPREDICT) {
+                    req.setCommand(WorkerCommands.STREAMPREDICT);
+                }
                 j.setScheduled();
                 req.addRequest(j.getPayload());
             }
@@ -56,8 +62,8 @@ public class BatchAggregator {
     }
 
     public void sendResponse(ModelWorkerResponse message) {
+        boolean jobDone = true;
         // TODO: Handle prediction level code
-
         if (message.getCode() == 200) {
             if (jobs.isEmpty()) {
                 // this is from initial load.
@@ -70,6 +76,10 @@ public class BatchAggregator {
                 if (job == null) {
                     throw new IllegalStateException(
                             "Unexpected job in sendResponse() with 200 status code: " + jobId);
+                }
+                if (job.getCmd() == WorkerCommands.STREAMPREDICT
+                        && prediction.getHeaders().get(TS_STREAM_NEXT).equals("true")) {
+                    jobDone = false;
                 }
                 job.response(
                         prediction.getResp(),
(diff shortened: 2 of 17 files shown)


docs/grpc_api.md
```diff
@@ -70,3 +71,28 @@ python ts_scripts/torchserve_grpc_client.py infer densenet161 examples/image_cla
...
+    rpc Ping(google.protobuf.Empty) returns (TorchServeHealthResponse) {}
+
+    // Predictions entry point to get inference using default model version.
+    rpc Predictions(PredictionsRequest) returns (PredictionResponse) {}
+
+    // Streaming response for an inference request.
+    rpc StreamPredictions(PredictionsRequest) returns (stream PredictionResponse) {}
+}
+```
+Backend handler calls "send_intermediate_predict_response" to send one intermediate result to frontend, and return the last result as the existing style. For example
```
HamidShojanazeri, line 89: thanks @lxning does backend handler meaning or referring to a custom handler at the moment?

How about it we change, `send_intermediate_predict_response(["hello"]` to `send_intermediate_predict_response(["intermediate_response"]`

docs/grpc_api.md
```diff
@@ -70,3 +71,28 @@ python ts_scripts/torchserve_grpc_client.py infer densenet161 examples/image_cla
 ```bash
 python ts_scripts/torchserve_grpc_client.py unregister densenet161
 ```
+## GRPC Server Side Streaming
+TorchServe GRPC APIs adds a server side streaming of the inference API "StreamPredictions" to allow a sequence of inference responses to be sent over the same GRPC stream. This new API is only recommended for the use case when the inference full response latency is high, and the inference intermediate results are sent to client. This new API automatically forces the batchSize to be one.
```
HamidShojanazeri, line 75: maybe we can add one line here, 
`This new API is only recommended for use cases such as generative AI and mostly LLMs (Large Language models) where the latency for the full response is high, and the intermediate results are available and can be sent to client as they get generated. This new API automatically forces the batchSize to be one.
`

msaroufim, line 75: n00b q: what does intermediate response mean? I initially understood this feature as send partial batches back so what's the scenario in which it'd be useful to use this feature? Or is this ineternal only to the large model work?

docs/grpc_api.md
```diff
@@ -70,3 +71,28 @@ python ts_scripts/torchserve_grpc_client.py infer densenet161 examples/image_cla
...
+```
+Backend handler calls "send_intermediate_predict_response" to send one intermediate result to frontend, and return the last result as the existing style. For example
+```
+from ts.protocol.otf_message_handler import send_intermediate_predict_response
+
+def handle(data, context):
+    if type(data) is list:
+        for i in range (3):
+            send_intermediate_predict_response(["hello"], context.request_ids, "Intermediate Prediction success", 200, context)
+        return ["hello world "]
```
msaroufim, line 97: should this be an async request

frontend/server/src/main/java/org/pytorch/serve/wlm/BatchAggregator.java
```diff
@@ -56,8 +62,8 @@ public BaseModelRequest getRequest(String threadName, WorkerState state)
     }
 
     public void sendResponse(ModelWorkerResponse message) {
+        boolean jobDone = true;
```
msaroufim, line 65: this variable name is a bit confusing, outside of the context of streaming - the job is not done yet, maybe streamcomplete or something of the sort would be cleaerer

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
```diff
@@ -201,8 +200,9 @@ public void pollBatch(String threadId, long waitTime, Map<String, Job> jobsRepo)
             logger.trace("get first job: {}", Objects.requireNonNull(j).getJobId());
 
             jobsRepo.put(j.getJobId(), j);
-            // describe request job batch size always is 1
-            if (j.getCmd() == WorkerCommands.DESCRIBE) {
+            // batch size always is 1 for describe request job and stream prediction request job
```
msaroufim, line 203: Not sure I understood this limitation why batch size 1

test/pytest/test_gRPC_inference_api.py
```diff
@@ -25,66 +26,138 @@ def __get_change(current, previous):
...
+        response = managment_stub.RegisterModel(
+            management_pb2.RegisterModelRequest(
+                url=path,
+                initial_workers=item["worker"],
+                synchronous=bool(item["synchronous"]),
+                model_name=item["model_name"],
+            )
+        )
+
+        print(response.msg)
```
msaroufim, line 145: are these prints necessary? slightly worries that messages will fill out our CI logs which are already long to make search frustrating

ts_scripts/torchserve_grpc_client.py
```diff
@@ -35,6 +35,23 @@ def infer(stub, model_name, model_input):
...
+    input_data = {"data": data}
+    responses = stub.StreamPredictions(
+        inference_pb2.PredictionsRequest(model_name=model_name, input=input_data)
+    )
+
+    try:
+        for resp in responses:
+            prediction = resp.prediction.decode("utf-8")
+            print(prediction)
+    except grpc.RpcError as e:
```
msaroufim, line 51: Should we also catch the `UnicodeDecodeError`?

ts/context.py
```diff
@@ -37,6 +37,7 @@ def __init__(
         self._metrics = None
         self._limit_max_image_pixels = True
         self.metrics = metrics
+        self.cl_socket = None
```
msaroufim, line 40: could we add a brief comment around how this is referring to a client socket? Had to spend a minute to figure out what this is refering to

test/pytest/test_gRPC_inference_api.py
```diff
@@ -25,66 +26,138 @@ def __get_change(current, previous):
...
+    input_data = {"data": data}
+    responses = stub.StreamPredictions(
+        inference_pb2.PredictionsRequest(model_name=model_name, input=input_data)
+    )
+
+    prediction = []
+    for resp in responses:
+        prediction.append(resp.prediction.decode("utf-8"))
+
+    return " ".join(prediction)
```
msaroufim, line 120: will this be a list of partial predictions so a single prediction or a list of multiplee predictions with batch size 1

msaroufim (changes requested):
Left a bunch of minor feedback but not sure I have enough context on what this PR is trying to do to give system-level feedback

frontend/server/src/main/java/org/pytorch/serve/wlm/BatchAggregator.java
lxning: The most use cases are non-streaming. They only  require one single message retrieving for a batch of jobs. Var "jobDone" is used to reflect if the message retrieving is completed for a batch of jobs.

test/pytest/test_gRPC_inference_api.py
lxning: prediction is a list of partial prediction responses. Here returns joining all of the partial response together to make the later comparing with expected values much easier.

test/pytest/test_gRPC_inference_api.py
lxning: it is helpful for debugging the regression test failure point (ie. which model registration fails).

docs/grpc_api.md
lxning: yes, it is backend customized handler

msaroufim (approved):

docs/grpc_api.md
HamidShojanazeri:
```suggestion
TorchServe GRPC APIs adds a server side streaming of the inference API "StreamPredictions" to allow a sequence of inference responses to be sent over the same GRPC stream. This new API is only recommended for the use case when the inference latency of the full response is high and the inference intermediate results are sent to client. An example could be LLMs for generative applications, where generating "n" number of tokens can have high latency, in this case user can receive each generated token once ready until the full response completes. This new API automatically forces the batchSize to be one.
```

HamidShojanazeri (approved):

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
excray: Not really true. We might want a batch of streams as well.It is upto the client.

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerThread.java
```diff
@@ -184,28 +186,55 @@ public void run() {
...
+                        int val = model.incrFailedInfReqs();
+                        logger.error("Number or consecutive unsuccessful inference {}", val);
+                        throw new WorkerInitializationException(
+                                "Backend worker did not respond in given time");
+                    }
+                } else {
+                    ModelInferenceRequest inferReq = (ModelInferenceRequest) req;
+                    boolean streamNext = true;
+                    while (streamNext) {
+                        reply = replies.poll(responseTimeout, TimeUnit.SECONDS);
```
excray, line 215: Looks like responseTimeout is same for streaming and non-streaming as well. clients might want different timeouts for streaming and non streaming api right?

lxning: responseTimeout is planned to move to model level config.

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
lxning: there are two issues when batch size is set >1.

- it breaks current the protocol b/w frontend and backend.
- the latency most likely will be even higher if batch size is larger than 1.

PR #2189, opened 2023-03-22
[RFC] Composable handler

### LINKED ISSUE #2181: Refactor our handlers and examples for more code reuse
After chatting with folks like @hamelsmu and a lot of folks on the core team it's become clear that we have a reusability problem with the way our examples and handlers are setup so this is an attempt to remedy that

So the goal of this RFC is to reduce the time it takes for someone to write their own custom handler and not provide custom handlers they can just out of the box

Some symptoms of the deeper problem are 
1. Our image classifier example can only resize images up to a specific size https://github.com/pytorch/serve/blob/master/ts/torch_handler/image_classifier.py#L21
2. Most of our large model model capabilities are available for a specific huggingface example https://github.com/pytorch/serve/tree/master/examples/Huggingface_Largemodels
3. Our compilation specific code is interspersed into the based handler as opposed to being standalone utils https://github.com/pytorch/serve/blob/master/ts/torch_handler/base_handler.py so much so that it's almost impossible to understand how torchserve works without reading the code for base handler

So the hot take from this RFC is to
1. Delete all handlers except for the base handler 
2. Refactor all non critical functionality from the base handler into standalone utils that can be easily composed into a new handler script
3. For critical features like large model inference or scriptable tokenizers we refactor them into their own utils

I don't want people to read an example and then copy it over, I want them to just type in `from ts.handler.utils import torchcompile, largemodelinference etc...` - that way they have a reusable set of primitives they can leverage in a more scalable way

mreso:
## Description

This PR is meant as an RFC to gather feedback on composable handler.
The motivation for this PR can be found in issue #2181 .

This PR converts all functionality provided by the handler classes into functional form and moves it into ts.handler_utils. A custom handler is then created by composing the handler using the functional interface. Each component can be individually configured:

```python
from ts.torch_handler.base_handler import BaseHandler
from ts.handler_utils import VisionPreprocessing, VisionPostProcessing
class CustomHandler(BaseHandler):
   def __init__(self):
      self.preprocess = VisionPreprocess(image_size=128, crop_size=64)
      self.postprocessing = VisionPostprocess(topk=5)
```

The solution is backward compatible and most handler written in the old style should work as before with one exception. It is now necessary to call super().__init__() at the beginning of the custom handler if it inherits from any of our handlers (BaseHandler, ImageClassifier etc.). Just like nn.Module does not let you create submodules until you executed super().__init__() we check if the BaseHandler constructor ran. If not we throw an error. There are 2 possible scenarios how we can incentivize users to call the super constructor:

a) Just announcing the upcoming change and throw an error in the new release.
b) Introduce composable handler whilst duplicating all the code and throw a warning that using handlers without calling the constructor is deprecated in upcoming release n. Then convert warning into an error in release n+1 or later while deleting duplicate code.

Happy about any other possible strategy.

Please note that the PR is a PoC and is meant to collect feedback. The handler unit tests run successful but there are at least a couple of imports missing in some paths which are not covered by the tests. (Great opportunity to extend the tests! :-) )

Looking forward for your feedback!

Fixes #2181 

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

## Checklist:

- [X] Did you have fun?


ts/torch_handler/image_classifier.py
```diff
@@ -15,26 +14,18 @@ class ImageClassifier(VisionHandler):
...
-        transforms.Normalize(mean=[0.485, 0.456, 0.406],
-                             std=[0.229, 0.224, 0.225])
-    ])
-
-    def set_max_result_classes(self, topk):
-        self.topk = topk
-
-    def get_max_result_classes(self):
-        return self.topk
+    image_processing = transforms.Compose(
```
msaroufim, line 19: I think we could be more aggressive, this is fairly domain specific logic and should be provided by users, we can keep it so tests dont break but then i'd prefer if we move this handler to the test utils

HamidShojanazeri: +1

mreso: Agreed, all handler in torch_handler will only stay for bc reasons. The only  remaining important handler will be the base handler as it will integrate compose logic and provide functionality that users to not want to provide.

msaroufim: my suspicion is very few people use them if at all, can't remember last time someone opened an issue with the builtin handlers - I might update the current RFC and just ask folks what they think - it should be fine for us to iron out a BC guarantee API by the time we reach 1.0

ts/handler_utils/caller.py
```diff
@@ -0,0 +1,16 @@
...
+
+class InitCaller(ABC):
+    def __call__(self, *args):
+        if self._prev:
+            self._prev(*args)
+
+        return self._method(*args)
+
+
+class PipeCaller(ABC):
```
msaroufim, line 12: wdyt about this just being a python iterable instead?

ts/handler_utils/base.py
```diff
@@ -0,0 +1,411 @@
...
+
+    model_class = model_class_definitions[0]
+    model = model_class()
+    if model_pt_path:
+        state_dict = torch.load(model_pt_path, map_location=obj.device)
+        model.load_state_dict(state_dict)
+    return model
+
+
+def base_initialize(obj, context):
```
msaroufim, line 81: you were probably already thinking this but worth splitting this out even more the IPEX, TensorRT, PT 2 logic shouldnt be in the base handler proper but in its own file maybe compiler_optimizations.py or something - regardless i was planning on taking a crack at this part so we can worry about it later

ts/torch_handler/base_handler.py
```diff
@@ -59,17 +40,26 @@ def check_pt2_enabled():
...
 class BaseHandler(abc.ABC):
     """
     Base default handler to load torchscript or eager mode [state_dict] models
     Also, provides handle method per torch serve custom model specification
     """
 
     def __init__(self):
+        self.initialized = False
+        super().__setattr__("__assigned__", set())
+        self.initialize = BaseInitialize()
```
msaroufim, line 55: nice and I do think it makes sense to error if super isnt set

ts/torch_handler/unit_tests/test_envelopes.py
```diff
@@ -51,7 +51,6 @@ def test_json_double_batch(handle_fn, base_model_context):
 
     envelope = JSONEnvelope(handle_fn)
     results = envelope.handle(test_data, base_model_context)
-    print(results)
```
msaroufim, line 54: lol thank you <3

ts/torch_handler/vision_handler.py
```diff
@@ -19,44 +17,12 @@ class VisionHandler(BaseHandler, ABC):
...
-                image = torch.FloatTensor(image)
-
-            images.append(image)
-
-        return torch.stack(images).to(self.device)
+    image_processing = transforms.Compose([transforms.ToTensor()])
+
+    def __init__(self):
+        super().__init__()
+        self.initialize = VisionInitialize(BaseInitialize())
```
msaroufim, line 24: mmm I don't quite like this as much as the rest, to me it feels like i should call one initialize or the other

Which reminds me it's kinda funky that we have both init and initialize

msaroufim (changes requested):

msaroufim:
I think all the limitations you've listed are OK, I also think we might be able to get away with breaking BC for this change

ts/handler_utils/caller.py
```diff
@@ -0,0 +1,16 @@
+from abc import ABC
+
+
+class InitCaller(ABC):
```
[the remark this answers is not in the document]
mreso: PipeCaller pipes the return value through the calls so you can use it like a torch.data DataPipe the other just calls all consecutively with the same parameter where you make changes to the context object given as parameter.
You can potentially use one or the other to achieve both but this way meant I did not have to touch the initialize and preprocess/postprocess functions of the handler at all. I use InitCaller for initialize and PipeCaller for the others without changing a line of code in the functionalized former member methods.

ts/handler_utils/caller.py
mreso: Okay, I think now I see what the confusion is. Will probably redefine this interface to not use interitance and put the constructor in the Caller. The intention is this:

```python
def foo(a):
   return a+1
def bar(b):
  return b+2
pp = PipeCaller(foo)
pp = PipeCaller(pp, bar)
c=pp(1) # c==4
```

ts/handler_utils/vision/vision.py
```diff
@@ -0,0 +1,62 @@
...
+import io
+
+import torch
+from captum.attr import IntegratedGradients
+from PIL import Image
+
+from ts.handler_utils.caller import InitCaller, PipeCaller
+
+
+def vision_preprocess(obj, data):
```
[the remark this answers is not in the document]
mreso: Yes, I did not touch the preprocessing methods so far. They should be split up like this:
```python
class CustomHandler(BaseHandler):
   def __init__(self):
      pp = Decode(256)
      pp = Resize(pp,256)
      pp = CenterCrop(pp, 224)
      pp = ToTensor(pp)
      self.preprocess = pp
```

ts/torch_handler/vision_handler.py
mreso: Agreed, for initialization its kind of weird. This reflects the previous behavior. We are calling [super().initialize() in VisionHandler](https://github.com/pytorch/serve/blob/6daaa428c855c6eff6831807bba46f109b623d79/ts/torch_handler/vision_handler.py#L23). I think if you break up the handler method like the pre- and postprocessing it could make sense.
```python
class CustomHandler(BaseHandler):
   def __init__(self):
      init = LoadModel()
      init = LoadWeights(init)
      init = Quantize(init, np.int8, layer_names)
      init = Compile(init, method='max-autotune')
      self.initialize = init
```

PR #2190, opened 2023-03-22
Frontend metrics configuration handling

### LINKED ISSUE #2134: Frontend metrics yaml parser
### 🚀 The feature

Parse metrics config yaml to get metrics name, dimension. 

### Motivation, pitch

See details in #1492

### Alternatives

_No response_

### Additional context

_No response_

namannandan:
## Description
Implementation of frontend metrics configuration parsing and validation.
Design: https://github.com/pytorch/serve/issues/1492

Fixes #2134 

## Type of change

- Unit tests included with PR run as part of sanity test

## Checklist:

- [x] Did you have fun?


frontend/server/src/main/java/org/pytorch/serve/metrics/configuration/YAMLMetricConfigurationHandler.java
```diff
@@ -0,0 +1,25 @@
...
+            throws FileNotFoundException, ComposerException, RuntimeException {
+        Constructor constructor = new Constructor(MetricConfiguration.class);
+        Yaml yaml = new Yaml(constructor);
+        FileInputStream inputStream = new FileInputStream(new File(configFilePath));
+        MetricConfiguration config = yaml.load(inputStream);
+        config.validate();
+
+        return config;
+    }
+}
```
lxning, line 25: Only a simple function as following is needed.

public static <T> T readYamlFile(File file, Class<T> type)
            throws InvalidModelException, IOException {
        Yaml yaml = new Yaml(new Constructor(type));
        try (Reader r =
                new InputStreamReader(
                        Files.newInputStream(file.toPath()), StandardCharsets.UTF_8)) {

            return yaml.load(r);
        } catch (YAMLException e) {
            throw new InvalidModelException("Failed to parse model config yaml file.", e);
        }
    }

lxning (commented):
TsMetrics and ModelMetrics are exactly same. it is not necessary to have them.

Only need:
- MetricConfiguraiton.java
- MetricSpec.java

frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
```diff
@@ -387,6 +388,10 @@ public String getTorchRunLogDir() {
         return torchrunLogDir;
     }
 
+    public String getMetricsMode() {
+        return getProperty(TS_METRICS_MODE, null);
```
lxning, line 392: default should be log, not null. please also add unit test.

namannandan: It is useful to return `null` in the case where `TS_METRICS_MODE` is not configured and the `mode` configuration is set in the configuration file. Here, we would not want to override the value set in the metrics configuration file with the default value returned by `getMetricsMode`: https://github.com/pytorch/serve/pull/2190/files#diff-c587b142c0b47f58987ec3279409e835e802bb8e27d2202cc504588e21e6f348R76-R79

The case where `mode` is not specified in both the metrics configuration file and in `TS_METRICS_MODE` is handled [here](https://github.com/pytorch/serve/pull/2190/files#diff-c587b142c0b47f58987ec3279409e835e802bb8e27d2202cc504588e21e6f348R81-R83).

So, as per above implementation, the priority order for mode configuration is as follows:
`TS_METRICS_MODE` > `metrics configuration file` > `default mode ("log")`

The default mode value, i.e `"log"` is verified [here](https://github.com/pytorch/serve/pull/2190/files#diff-2f95379a19d5fcc641c241febeb61ec5dcb0207b0224f702df489844d2efa75fR84).

frontend/server/src/main/java/org/pytorch/serve/metrics/configuration/MetricConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+        config.validate();
+
+        String metricsMode = ConfigManager.getInstance().getMetricsMode();
+        if (metricsMode != null) {
+            config.setMode(metricsMode.toLowerCase());
+        }
+
+        if (config.getMode() == null) {
+            config.setMode("log");
+        }
```
lxning, line 83: Let's make mode config simple via ConfigManager as a single place to control the mode.

frontend/server/src/main/java/org/pytorch/serve/metrics/configuration/MetricTypes.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public List<MetricSpecification> getHistogram() {
+        return this.histogram;
+    }
+
+    public void validate() {
+        if (this.counter != null) {
+            for (MetricSpecification spec : this.counter) {
+                spec.validate();
+            }
```
lxning, line 38: shouldn't return at here?

frontend/server/src/main/java/org/pytorch/serve/metrics/configuration/MetricTypes.java
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        if (this.gauge != null) {
+            for (MetricSpecification spec : this.gauge) {
+                spec.validate();
+            }
```
lxning, line 44: ditto

frontend/server/src/main/java/org/pytorch/serve/metrics/configuration/MetricTypes.java
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        if (this.histogram != null) {
+            for (MetricSpecification spec : this.histogram) {
+                spec.validate();
+            }
```
lxning, line 50: ditto

frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
lxning: To make it simple, let's use ConfigManager as a single place to control the mode. The default value in ConfigManager is log.

the default value can be set in config.property or "TS_xxx" env.

frontend/server/src/main/java/org/pytorch/serve/metrics/configuration/MetricTypes.java
namannandan: Here, we're validating all the metric configurations that have been loaded, `this.counter` is a list of metric specifications. If there are no counter metrics defined, we skip it and check for the next metric type. So we won't return here after validating only `this.counter`.

frontend/server/src/main/java/org/pytorch/serve/metrics/configuration/MetricConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    public static MetricConfiguration loadConfiguration(String configFilePath)
+            throws FileNotFoundException, ComposerException, RuntimeException {
+        Constructor constructor = new Constructor(MetricConfiguration.class);
+        Yaml yaml = new Yaml(constructor);
+        FileInputStream inputStream = new FileInputStream(new File(configFilePath));
+        MetricConfiguration config = yaml.load(inputStream);
+        config.setMode(ConfigManager.getInstance().getMetricsMode());
```
lxning, line 74: let's make it simple. metric mode is only configured via ConfigManager  such as TS_Metric_Mode, not from metric yaml config file. please also update default metric config yaml (remove metric mode)

lxning (approved):

agunapal (approved):

PR #2222, opened 2023-04-06
Refactor PT2 code changes

msaroufim:
## Description

Note: the link check lint keeps timing out so i might ignore it for now

Now that PT 2.0 was official released there's a few updates we needed to make
1. Gate behavior behind checks for PT 2.0 version before using compile
2. Refactor a lot of our optimization utils: ipex, onnx, pt2 into standalone functions - I can't quite put them in seperate files since imports need to happen in the same file in python
3. Using the new config.yaml and deprecated the compile.json - I updated a test util to also leverage it
4. Added a few tests - something I'm not too happy with is there seems to be some code duplication in our pytest folder with many ways of running a test, a mock handler can't take in extra files and I don't have bandwidth right now to fix all these issues
5. The yaml config object is not guaranteed to have the config but should check for it in model initialization
7. I check in logs if a compilation actually occurs
8. Added a few misc warnings and info calls that are useful
9. I checked that ONNX tests still pass, if we're merging all the distributed PRs this week I'd like to just add the ONNX dependency as well

## Type of change

- [x] Tested in CI

## Checklist:

- [x] Did you have fun?


msaroufim:
@morgandu FYI this is how we're gonna do the configs moving forward

ts/torch_handler/base_handler.py
```diff
@@ -25,46 +21,74 @@
...
-        )
-        pt2_enabled = False
-    return pt2_enabled
+if packaging.version.parse(torch.__version__) >= packaging.version.parse("2.0.0"):
+    PT2_AVAILABLE = True
+    if torch.cuda.is_available():
+        # If Ampere enable tensor cores which will give better performance
+        # Ideally get yourself an A10G or A100 for optimal performance
+        if torch.cuda.get_device_capability() >= (8, 0):
+            torch.backends.cuda.matmul.allow_tf32 = True
```
HamidShojanazeri, line 42: @msaroufim wondering if `torch.set_float32_matmul_precision('high')` will be set automatically, in my benchmark before I found in for example HF models it was not set automatically and setting it change the perf substantially.

ts/torch_handler/base_handler.py
```diff
@@ -25,46 +21,74 @@
...
+    logger.info("ONNX enabled")
+    ONNX_AVAILABLE = True
+except ImportError as error:
+    logger.warning("proceeding without onnxruntime")
+    ONNX_AVAILABLE = False
+
+
+def setup_ort_session(model_pt_path):
+    providers = (
+        ["CUDAExecutionProvider", "CPUExecutionProvider"]
```
HamidShojanazeri, line 78: wonder if it should be  "CUDAExecutionProvider" if `self.map_location == "cuda"`?

ts/torch_handler/base_handler.py
```diff
@@ -160,49 +163,39 @@ def initialize(self, context):
...
-        if check_pt2_enabled() and backend:
+        if PT2_AVAILABLE and valid_backend:
             # Compilation will delay your model initialization
             try:
                 self.model = torch.compile(
                     self.model,
-                    backend=backend,
-                    mode="default" if TORCHXLA_AVAILABLE else "reduce-overhead",
+                    backend=pt2_backend,
+                    mode="default" if XLA_AVAILABLE else "reduce-overhead",
```
HamidShojanazeri, line 186: @msaroufim does "reduce-overhead" force the cuda graphs? in that case I am thinking if we could default to something more flexible as it may break more often.

HamidShojanazeri: alternatively adding a comment/to suggest this mode/ warn about it.

ts/torch_handler/base_handler.py
msaroufim: Yeah those are the same thing but I agree that warnings are suggesting matmul precision so will change

ts/torch_handler/base_handler.py
msaroufim: So for online inference with small batch size I feel like if we can't do reduce overhead then we shouldn't use torch.compile at all. Granted maybe we should also pass in the mode as a field in the yaml file

ts/torch_handler/base_handler.py
msaroufim: Oh yeah good point, I'm not too happy with the state of our ONNX tests at this point so maybe this is a good opportunity to revisit

msaroufim: ACtually just checked this is not needed, ORT will read in the list of providers from left to right and pick the first available one, so if cuda is not available cpu will be picked

test/pytest/test_torch_compile.py
```diff
@@ -0,0 +1,106 @@
...
+YAML_CONFIG = os.path.join(TEST_DATA_DIR, "pt2.yaml")
+
+
+SERIALIZED_FILE = os.path.join(TEST_DATA_DIR, "model.pt")
+MODEL_STORE_DIR = os.path.join(TEST_DATA_DIR, "model_store")
+MODEL_NAME = "half_plus_two"
+
+
+@pytest.mark.skipif(PT_2_AVAILABLE == False, reason="torch version is < 2.0.0")
+class TestTorchCompile:
```
lxning, line 34: the functions such as test_start_torchserve, test_server_status, test_registered_model, and test_serve_inference can be generalized to be shared with the other test cases. This work can be done later.

examples/pt2/README.md
```diff
@@ -1,30 +1,30 @@
...
 pip install torchserve torch-model-archiver
 ```
 
 ## Package your model
 
-PyTorch 2.0 supports several compiler backends and you pick which one you want by passing in an optional file `compile.json` during your model packaging
+PyTorch 2.0 supports several compiler backends and you pick which one you want by passing in an optional file `model_config.yaml` during your model packaging
 
-`{"pt2" : "inductor"}`
+`pt2: "inductor"`
```
lxning, line 20: does this pr cover the feature enable/disable torch.compile flag?

ts/torch_handler/base_handler.py
HamidShojanazeri: agree with making it a config as I guess it increase the time/mem pressure for slightly larger models --> run into OOM easily.

ts/torch_handler/base_handler.py
HamidShojanazeri: I see,  it seems setting to HIGH will allow tf32 https://github.com/pytorch/pytorch/blob/d8d479c854c622a6ef21190b7e62c9a76f0ea2a7/aten/src/ATen/Context.cpp#L178-L184,

test/pytest/test_data/torch_compile/compile_handler.py
```diff
@@ -0,0 +1,20 @@
...
+        super().initialize(context)
+
+    def preprocess(self, data):
+        instances = data[0]["body"]["instances"]
+        input_tensor = torch.as_tensor(instances, dtype=torch.float32)
+        return input_tensor
+
+    def postprocess(self, data):
+        # Convert the output tensor to a list and return
+        return data.tolist()[2]
```
agunapal, line 20: Going back to the idea of making features composable, trying to understand if this handler is specific to 2.0 or needed for this example? For someone to use torch.compile, do they need a custom handler. I think it would be good to clarify that in the README.

agunapal (approved):
LGTM otherwise

HamidShojanazeri (approved):

test/pytest/test_data/torch_compile/compile_handler.py
msaroufim: So this is just a test handler, you don't need to create your own custom handler for torch.compile, I just put it since it's one that's easy to understand. Testing the base handler directly is kind of annoying and something I'd like us to fix eventually

msaroufim:
GitHub is acting up in CI. Whoever sees this tomorrow please merge

msaroufim:
@agunapal does the docker 3.10 CI failure look legit to you here?

agunapal:
@msaroufim the 3.8 failure seems like a network issue. I am not sure about the 3.10 one. I guess we need to add a retry action for this . This is the first time i have seen it fail

Issue #2232, opened 2023-04-16
http1.1 stream response via chunked encoding


lxning:
### 🚀 The feature

TorchServe supports streaming response for both HTTP and GRPC endpoint. 
- GRPC server side streaming see #2180 
- HTTP chunked encoding will be implemented in this ticket

### Motivation, pitch

Usually the predication latency is high (eg. 5sec) for a large model inference. Some models are able to generate intermediate prediction results (eg. generator AI). This feature will send the intermediate prediction results to user once the results are ready. The user will gradually get the entire response. For example, User may get first intermediate response within 1 sec, and gradually get the entire result until 5sec. This feature is to improve user prediction experience.

PR #2233, opened 2023-04-16
http stream response via http 1.1 chunked encoding

lxning:
## Description

Support http streaming inference response via http 1.1 chunked encoding

Fixes #(issue)
#2232 

## Type of change

- [ ] Test A
[reg.txt](https://github.com/pytorch/serve/files/11256622/reg.txt)

## Checklist:

- [ ] Did you have fun?


msaroufim:
Just 2 basic questions before I start reviewing
1. What's a usecase for this PR? What's a good example for an intermediate response that a user might actually want?
2. How does this interact with batching? Do we provide partial responses for all elements in a batch?

docs/inference_api.md
```diff
@@ -95,6 +95,34 @@ The result is JSON that tells you that the image is most likely a tabby cat. The
...
+def handle(data, context):
+    if type(data) is list:
+        for i in range (3):
+            send_intermediate_predict_response(["intermediate_response"], context.request_ids, "Intermediate Prediction success", 200, context)
+        return ["hello world "]
+```
+Client side receives the chunked data.
+```
+def test_echo_stream_inference():
+    test_utils.start_torchserve(no_config_snapshots=True, gen_mar=False)
```
msaroufim, line 111: can we just promote these test_utils functions to the core library they're very useful

frontend/server/src/main/java/org/pytorch/serve/wlm/BatchAggregator.java
```diff
@@ -77,9 +75,16 @@ public void sendResponse(ModelWorkerResponse message) {
                     throw new IllegalStateException(
                             "Unexpected job in sendResponse() with 200 status code: " + jobId);
                 }
-                if (job.getCmd() == WorkerCommands.STREAMPREDICT
-                        && prediction.getHeaders().get(TS_STREAM_NEXT).equals("true")) {
-                    jobDone = false;
+                if (jobDone) {
```
msaroufim, line 78: do all the java side changes need unit tests?

lxning: There is a tech debt in frontend which is lack of unit tests, only includes [integration test ](https://github.com/pytorch/serve/blob/master/frontend/server/src/test/java/org/pytorch/serve/ModelServerTest.java). This PR uses regression to test the e2e.

docs/inference_api.md
lxning: which core library? who is the user of the core library?

msaroufim: I mean the `ts` namespace

msaroufim (approved):

frontend/server/src/main/java/org/pytorch/serve/wlm/BatchAggregator.java
```diff
@@ -61,13 +59,13 @@ public BaseModelRequest getRequest(String threadName, WorkerState state)
         return req;
     }
 
-    public void sendResponse(ModelWorkerResponse message) {
+    public boolean sendResponse(ModelWorkerResponse message) {
```
sanjams2, line 62: nit: A java doc here might be helpful, specifically highlighting what the return value denotes.

ts/protocol/otf_message_handler.py
```diff
@@ -79,11 +79,11 @@ def create_predict_response(
         if context is None:
             msg += struct.pack("!i", 0)  # content_type
         else:
-            if "true" == context.get_request_header(idx, "ts_stream_next"):
-                if ts_stream_next is False:
+            if ts_stream_next is True:
+                context.set_response_header(idx, "ts_stream_next", "true")
+            else:
+                if "true" == context.get_response_headers(idx).get("ts_stream_next"):
```
sanjams2, line 85: Curious: why do we need to check this? When would this be false?

frontend/server/src/main/java/org/pytorch/serve/job/RestJob.java
```diff
@@ -136,29 +154,42 @@ private void responseInference(
...
-            NettyUtils.sendHttpResponse(ctx, resp, true);
+            if (numStreams == 0) { // non-stream response
+                MetricAggregator.handleInferenceMetric(
+                        getModelName(), getModelVersion(), getScheduled() - getBegin(), inferTime);
+                NettyUtils.sendHttpResponse(ctx, resp, true);
+            } else if (numStreams == -1) { // the last response in a stream
+                MetricAggregator.handleInferenceMetric(
+                        getModelName(), getModelVersion(), getScheduled() - getBegin(), inferTime);
+                ctx.writeAndFlush(new DefaultHttpContent(Unpooled.wrappedBuffer(body)));
+                ctx.writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT);
```
sanjams2, line 165: Minor/Curious: Instead of writing twice, can you just write once and tell netty that this is the last chunk. Also assuming you cant do that, do you need to `writeAndFlush` here or can you just write knowing that you are immediately `writeAndFlush`'ing the `LastHttpConent` chunk after?

sanjams2 (commented):
Mostly minor comments except one about the logging. I would also like to see the enum used to denote the current state of a request instead of the int (especially with the current choice for which int denotes which state), but this is more of an opinion so therefore not blocking.

frontend/server/src/main/java/org/pytorch/serve/job/RestJob.java
```diff
@@ -38,6 +45,7 @@ public class RestJob extends Job {
 
     private ChannelHandlerContext ctx;
     private CompletableFuture<byte[]> responsePromise;
+    private int numStreams; // -1: stream end; 0+: number of streams received; 0: non stream
```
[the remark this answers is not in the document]
lxning: the following 4 cases need take different actions.
-1: stream end.
0: non-stream (default).
1: 1st stream.
>1: 2nd+ stream.

frontend/server/src/main/java/org/pytorch/serve/job/RestJob.java
lxning: it needs call writeAndFlush twice, kind of similar as grpc (onComplete).

ts/protocol/otf_message_handler.py
lxning: there are 2 scenarios when ts_stream_next is false.
1. non-stream response.
2. last stream response. Line 85 is to handle this case.

context.get_response_headers(idx).get("ts_stream_next") is a string, not bool. it's value can be:
"true": has next stream
"false": last stream
"": non-stream response.

sanjams2 (approved):

HamidShojanazeri (approved):

PR #2256, opened 2023-04-21
Enable `torch.compile` support for torch 2.0.0 pre release

morgandu:
We are currently locked on an alpha version of torch 2.0.0. TorchServe (HEAD) was working for us before the PT2 refactor when the `base_handler` was deciding PT2, as below:
```
    try:
        import torch._dynamo

        pt2_enabled = True
```
Creating a minor PR for enable torch2.0.0 alpha pre release to use torch.compile.

### DIFF
--- a/ts/torch_handler/base_handler.py
+++ b/ts/torch_handler/base_handler.py
@@ -37,7 +37,7 @@ except ImportError as error:
     XLA_AVAILABLE = False
 
 
-if packaging.version.parse(torch.__version__) >= packaging.version.parse("2.0.0"):
+if packaging.version.parse(torch.__version__) >= packaging.version.parse("2.0.0a"):
     PT2_AVAILABLE = True
     if torch.cuda.is_available():
         # If Ampere enable tensor cores which will give better performance


msaroufim (approved):

agunapal (commented):
@msaroufim I wonder if the fix should be to change the logic to just check for >= 2 in 2.x.x

agunapal:
@morgandu Could you please share whats the context of this alpha release. Also, the proposed fix would mean torch.compile wouldn't work on 2.0.0

msaroufim:
@agunapal for context this PR is needed for the new TPU support, I believe this PR is correct and will return True in the case of torch__version__ = 2.0.0 since that's considered greater than an alpha release

morgandu:
> @morgandu Could you please share whats the context of this alpha release. Also, the proposed fix would mean torch.compile wouldn't work on 2.0.0?

@agunapal we are currently fixed on using a particular version of torch 2.0.0, which is an alpha pre release version. According to the packaging.version comparison. '2.0.0a0' < '2.0.0'. However, I believe dynamo is enabled in '2.0.0a0'. 

So this change should enable the torch.compile support for users on any pre release of 2.0.0, but not lower, and it will not affect users on 2.0.0 official release / post release and other future versions above 2.0.0.

agunapal (approved):

morgandu:
All green but I don't have write access so can't merge myself. Please help merge this PR. Thanks!

PR #2275, opened 2023-04-25
Regression fix

HamidShojanazeri:
## Description

Fixing the dist inference test as it needs multi-gpu machine. Changing the regression runner to g5 instance being used for benchmarks.


msaroufim (approved):

test/pytest/test_distributed_inference_handler.py
```diff
@@ -26,7 +26,7 @@
 
 
 @pytest.mark.skipif(
```
msaroufim, line 28: You can break apart these lines to make it easier to read

```
pytest.mark.skipIf(condition_a)
pytest.mark.skipIf(condition_b)
test_function()
```

I had to triple read the not condition and where the parenthesis was to make sure there wasn't some bug that would run this on CPU

Would also be great to make sure this test gets skipped in the CPU regression tests

msaroufim (changes requested):
c comment

test/pytest/test_distributed_inference_handler.py
HamidShojanazeri: Thanks Mark, sure, just testing now if solve the issue the would fix regression_gpu.yaml back to run nightlies.

.github/workflows/regression_tests_gpu.yml
```diff
@@ -1,14 +1,14 @@
...
-  schedule:
-    - cron:  '15 9 * * *'
+  # schedule:
+  #   - cron:  '15 9 * * *'
 
 jobs:
   regression-gpu:
     # creates workflows for CUDA 11.6 & CUDA 11.7 on ubuntu
-    runs-on: [self-hosted, regression-test-gpu]
+    runs-on: [self-hosted, benchmark-gpu-new]
```
lxning, line 11: let's do not change the runner at this moment. we can skip the large model regression if # gpu < 4

PR #2310, opened 2023-05-03
Pippy deferred init

HamidShojanazeri:
## Description
This PR adds deferred init to pippy for large model inference.

- [ X] New feature (non-breaking change which adds functionality)
- [ X] This change requires a documentation update


examples/large_models/Huggingface_pippy/Readme.md
```diff
@@ -6,23 +6,16 @@ PiPPy provides pipeline parallelism for serving large models that would not fit
 
 ## How to serve your large HuggingFace models with PiPPy in Torchserve?
 
-We use a Torchserve custom handler that inherits from base_pippy_handler to load the model and define our logic for preprocess, inference and post processing. This is basically very similar to your evaluation process.
+We use a Torchserve custom handler that inherits from base_pippy_handler to load the model and define our logic for preprocess, inference and post processing. This is basically very similar to your evaluation process. Following settings has been tested on g5.12xlarge EC2 instance which has 4xA10 GPUs.
 
 ### Step 1: Download model
```
chauhang, line 11: What about the pippy install steps? Will it be bundled as part of the base dependencies?

examples/large_models/Huggingface_pippy/Readme.md
```diff
@@ -46,37 +39,42 @@ pippy:
...
+    dtype: fp16 # data type to load your model checkpoint, supported fp32, fp16, bf16
 ```
 
 ### Step 3: Generate Tar/ MAR file
 
 Navigate up to `Huggingface_Largemodels` directory.
 
 ```bash
-torch-model-archiver --model-name bloom --version 1.0 --handler pippy_handler.py --extra-files model/models--facebook--opt-iml-max-1.3b/snapshots/d60fa58f50def19751da2075791da359ca19d273  -r requirements.txt --config-file model-config.yaml --archive-format tgz
+torch-model-archiver --model-name opt --version 1.0 --handler pippy_handler.py  -r requirements.txt --config-file model-config.yaml --archive-format tgz
```
chauhang, line 57: Will be helpful to also add steps for --no-archive mode so that user can point to the directory that has the downloaded model  and use that without creating archive file

examples/large_models/Huggingface_pippy/Readme.md
```diff
@@ -46,37 +39,42 @@ pippy:
     input_names: ['input_ids'] # input arg names to the model, this is required for FX tracing
     model_type: "HF" # set the model type to HF if you are using Huggingface model other wise leave it blank or any other model you use.
     rpc_timeout: 1800
+    num_worker_threads: 512 #number of threads for rpc worker usually 512 is a good number
```
chauhang, line 42: Do we need to such large value for pippy to work? Will be good to verify, if there is any special requirement override the default value of 16 for RPC num_worker_threads

examples/large_models/Huggingface_pippy/model-config.yaml
```diff
@@ -17,5 +17,8 @@ pippy:
     num_worker_threads: 512
 
 handler:
+    model_path: "/home/ubuntu/serve/examples/large_models/Huggingface_pippy/model/models--facebook--opt-30b/snapshots/ceea0a90ac0f6fae7c2c34bcb40477438c152546"
```
chauhang, line 20: Add comment asking users to change the path to actual value of the model file

examples/large_models/Huggingface_pippy/Readme.md
chauhang, line 57: Will be good to clarify that in this case since the model file is large we are using the model-config file to point to location of the file so that .gz does not copy over entire model again and simplifies the developer testing. For production settings, the large models can be put in some shared location and used from there in the model-config. We can create a separate doc page describing the details for "large models" and point to that.

examples/large_models/Huggingface_pippy/Readme.md
```diff
@@ -46,37 +39,42 @@ pippy:
...
 ### Step 5: Start torchserve
 
 Update config.properties and start torchserve
 
 ```bash
-torchserve --ncs --start --model-store model_store --models bloom.mar
+torchserve --ncs --start --model-store model_store --models opt.tar.gz
 ```
 
 ### Step 6: Run inference
```
chauhang, line 76: Will be helpful to have steps to verify model has been loaded, since for large models more warm-up time may be needed for loading

chauhang (changes requested):
Thanks Hamid. I have left few review comments based on the testing

examples/large_models/Huggingface_pippy/Readme.md
HamidShojanazeri: Good catch, its in TS dev dependencies, will add to the example requirements.txt and call it out.

examples/large_models/Huggingface_pippy/Readme.md
HamidShojanazeri: we have logs in place to show the model loading time when its finished.

ts/handler_utils/distributed/pt_pippy.py
```diff
@@ -54,7 +54,12 @@ def get_pipeline_driver(model, world_size, ctx):
     chunks = ctx.model_yaml_config["pippy"]["chunks"]
     input_names = ctx.model_yaml_config["pippy"]["input_names"]
     model_type = ctx.model_yaml_config["pippy"]["model_type"]
+    model_path = ctx.model_yaml_config["handler"]["model_path"]
+    index_filename = ctx.model_yaml_config["handler"]["index_filename"]
```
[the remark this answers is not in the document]
HamidShojanazeri: Great point, will add it.

examples/large_models/Huggingface_pippy/Readme.md
HamidShojanazeri: right, this number has been set mostly for training to avoid deadlock caused by RPC thread pool drain. For inference it seems it can be lowered like 4x. There is no clear number guidance here checked with Ke.

examples/large_models/Huggingface_pippy/Readme.md
```diff
@@ -46,37 +48,42 @@ pippy:
     input_names: ['input_ids'] # input arg names to the model, this is required for FX tracing
     model_type: "HF" # set the model type to HF if you are using Huggingface model other wise leave it blank or any other model you use.
     rpc_timeout: 1800
+    num_worker_threads: 512 #number of threads for rpc worker usually 512 is a good number
 
 handler:
     max_length: 80 # max length of tokens for tokenizer in the handler
+    model_name: "/home/ubuntu/serve/examples/large_models/Huggingface_pippy/model/models--facebook--opt-30b/snapshots/ceea0a90ac0f6fae7c2c34bcb40477438c152546" #the path to the checkpoints, in this example downloaded file. Please change to your model path.
```
lxning, line 55: Here is a hard coded path for model. Can we update it with HF cache dir to make it more generic for user to follow? We can set env
HUGGINGFACE_HUB_CACHE and TRANSFORMERS_CACHE

HamidShojanazeri: This is specific to this example so if someone use the download script here, then they will end up having the path. Overall we are expecting user to pass the checkpoint path.

examples/large_models/Huggingface_pippy/Readme.md
```diff
@@ -6,23 +6,25 @@ PiPPy provides pipeline parallelism for serving large models that would not fit
...
+pip install torchpippy
+
 ```
-paste the token generated from huggingface hub.
+
+### Step 1: Download model
 
 ```bash
-python Download_model.py --model_name facebook/opt-6.7b
+python ../Download_model.py --model_name facebook/opt-30b
```
lxning, line 23: the path is "../utils/Download_models.py"

lxning (approved):

ts/handler_utils/distributed/pt_pippy.py
```diff
@@ -51,10 +51,46 @@ def get_pipeline_driver(model, world_size, ctx):
...
     """
     # Extract configuration parameters from the context
+
+    # Check that the "pippy" and "handler" keys are present in the YAML config
+    assert "pippy" in ctx.model_yaml_config, "Missing 'pippy' key in YAML config"
+    assert "handler" in ctx.model_yaml_config, "Missing 'handler' key in YAML config"
+
+    # Check that the required keys are present in the "pippy" section
+    assert (
+        "chunks" in ctx.model_yaml_config["pippy"]
```
chauhang, line 61: Can we use default value of 1 if not specified for chunks?

chauhang (commented):
Left few comments

chauhang (approved):

agunapal (approved):
LGTM

PR #2320, opened 2023-05-04
add omp in env for torchrun and update doc

lxning:
## Description

Fixes #(issue)

## Type of change

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

## Checklist:

- [ ] Did you have fun?


chauhang (commented):
Thanks Li for the updates. Few comments: 

1. Will be good to move the Download_models.py one level up as a common file after https://github.com/pytorch/serve/pull/2310 is merged. 

2. For the Large model best practices, we should expand and provide more details for model packaging, threads and other configs and troubleshooting tips. This can be done as a separate PR

docs/large_model_inference.md
```diff
@@ -2,11 +2,22 @@
 
 This document explain how Torchserve supports large model serving, here large model refers to the models that are not able to fit into one gpu so they need be split in multiple partitions over multiple gpus.
 
+## Internal
```
HamidShojanazeri, line 5:
```suggestion
## How it works?
```

examples/large_models/deepspeed/opt/requirements.txt
```diff
@@ -1,2 +1,2 @@
-transformers
-deepspeed
+transformers==4.28.1
```
chauhang, line 1: Is there a reason for pinning the version numbers? Can we do >= instead?

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerLifeCycle.java
```diff
@@ -187,6 +187,7 @@ public void startWorker(int port, String deviceIds)
     private void attachRunner(
             ArrayList<String> argl, List<String> envp, int port, String deviceIds) {
         envp.add("LOGLEVEL=INFO");
+        envp.add("OMP_NUM_THREADS=1");
```
chauhang, line 190: Why is the value hardcoded? OMP_NUM_THREADS = 1 by default. It will be good to allow user to change this value,  my tests showed speedup on multi-gpu instances for higher value when the CPU core are available

chauhang (commented):
Left few comments please check

examples/large_models/deepspeed/opt/requirements.txt
lxning: The version combination of DS and transformer is not stable. Some combination does not work. So far this combination works for this example.

frontend/archive/src/main/java/org/pytorch/serve/archive/model/ModelConfig.java
```diff
@@ -402,6 +409,13 @@ public static TorchRun build(Map<?, ?> torchRunMap) {
                                     logger.warn("Invalid torchrun.node-rank:{}, reset to 0", v);
                                 }
                                 break;
+                            case "OMP_NUMBER_THREADS":
```
chauhang, line 412: Is the value getting picked from the model-config.yaml? Please add steps in the documentation as well for how user's can set this value

chauhang (commented):
For OMP_NUM_THREADS, please add documentation for how user can set this value

frontend/archive/src/main/java/org/pytorch/serve/archive/model/ModelConfig.java
lxning: The default value is 1, which is auto set in the code. The value can be modified in model-config.yaml.

chauhang (approved):

PR #2324, opened 2023-05-04
Basic CI for k8s

agunapal:
## Description

Added a basic testcase to validate mnist inference with kubernetes using minikube

Currently, this is going to run once a day, pulling the latest nightly docker cpu image

Fixes #(issue)

## Type of change

[k8s CI](https://github.com/pytorch/serve/files/11402356/1_test-model-prediction-on-k8s.txt)

## Checklist:

- [ ] Did you have fun?


.github/workflows/k8s-ci.yaml
```diff
@@ -0,0 +1,37 @@
+name: k8s CI
```
msaroufim, line 1: So does this running telling us anything about whether what's here is safe to run https://github.com/pytorch/serve/tree/master/kubernetes - if not then I'm not sure this should be its own CI job

agunapal: The goal is to have a CI run with a k8s executing multiple test cases. Right now this has a single testcase similar to docker

If things work with minikube on one instance, then we can assume it will work with https://github.com/pytorch/serve/tree/master/kubernetes ? 

if not with minikube, we will have to implement something with multiple machines, aws etc. That might be validating EKS more than torchserve

msaroufim: Makes sense - I think then It'b be good to check whether the kserve docker image works https://github.com/pytorch/serve/tree/master/kubernetes/kserve

agunapal:
Will send a new PR when I can figure out how to test kserve images

Issue #2292, opened 2023-04-29
Request Identification by Header Value


mpoemsl:
### 🚀 The feature

Parse an external request id from a custom header value. This request id is used as a prefix to the `TorchServe`-internal request id which occurs in logs. The uniqueness of internal ids is still guaranteed, since the suffix is the `uuid` as generated by the current method. 

The feature is optional and backwards compatible in the sense that for requests for which the header is not set, the current request id behavior is retained.

### Motivation, pitch

If `TorchServe` is used as part of a microservice architecture, this feature makes the tracing of individual requests across the whole architecture easier.

### Additional context

I implemented this feature as part of my work at @textshuttle in a manner customized to their products' needs. If there is interest in this feature, I could create a more general PR for this feature. Note that users that do not need this feature can choose to simply not set the header value and retain the current behavior.

aniongithub:
This is a great idea - I ended up here because I was looking for this feature myself.  Most real-world architectures that use Torchserve do so as part of a larger service architecture so this makes complete sense.

PR #2347, opened 2023-05-15
Set request ID prefix via request header

mpoemsl:
## Description

Implement #2292. If a request contains a value for header `x-request-id-prefix`, this string is used as a prefix to the internal request id, which appears in logs and is returned in the response header `x-request-id`. This can be useful for tracing individual requests in a multi-service architecture.

## Type of change

I have implemented a test `testPredictionsRequestIdPrefixHeader` in `ModelServerTest.java`, which makes a request with a header value `x-request-id-prefix` and checks if it is the prefix of response header value `x-request-id`.

The test can be reproduced by running `./gradlew clean test` in `frontend/server`. I have also ran existing tests as well as `spellcheck.sh` to ensure no new typos are introduced.

## Checklist:

- [x] Did you have fun?

## Note

This is my first feature PR. I plan to make more PRs in the future, so I would welcome any constructive feedback on the form of this PR to ensure that my future PRs will be useful.


mpoemsl:
I have added `fcc` and `ee` - which are parts of the example request ids used in the docs - to the spellcheck word list to make the test pass. However, I'm not sure if that's the best solution and would value any input on this.

lxning:
@mpoemsl Thank you for the contribution. it looks good to me for the http part. Could you also add this feature on grpc side?

mpoemsl:
> @mpoemsl Thank you for the contribution. it looks good to me for the http part. Could you also add this feature on grpc side?

Thanks! I'm not that familiar with gRPC, but I'll try to also implement the feature on that side.

msaroufim:
@lxning should we still merge this without gRPC support to unblock OP?

mpoemsl:
> @lxning should we still merge this without gRPC support to unblock OP?

That would also be okay for me! In terms of gRPC support, it's unclear to me to how to extract a HTTP header field with the existing libraries and if this makes sense at all. I haven't given up on it, but it's unlikely that I will be able to implement it soon.

Issue #2333, opened 2023-05-10
return 503 for inference request if job queue is full


lxning:
### 🚀 The feature

When a job queue of a model is full, the new inference request is blocked to add into the job queue. For high SLA requirements use case, user expects an immediate return 503 for an inference request if jobQ is full.

The feature requires to provide an option for users to enable an immediate return 503 for an inference request if jobQ is full.

### Motivation, pitch

When a job queue of a model is full, the new inference request is blocked to add into the job queue.  For high SLA requirements use case, user expects an immediate return 503 for an inference request if jobQ is full.

Issue #2322, opened 2023-05-04
Allow different queue sizes for different models


lopuhin:
### 🚀 The feature

I'd like to allow having different queue sizes (job_queue_size) for different models. As I understand, each model has it's own queue, but their size is the same.

### Motivation, pitch

We'd like to have a minimal queue for one of the models ("main one"), but it can also call other models from the same instance, where a larger queue size would be helpful. The reason for separation would be that we want more CPU workers than GPU workers.

### Alternatives

An alternative would be to run multiple torch serve instances, but this is slightly less convenient operationally.

agunapal:
@lopuhin Currently we don't have a way to monitor the queue size. So, I am curious about what issue you are facing by having the same queue size. Ex: with queue size 100, what is the problem in the main queue?

agunapal:
@lopuhin Thanks for the context. I'll bring it up with the team

PR #2350, opened 2023-05-16
support jobQueueSize and job ticket per model in model config YAML

lxning:
## Description

Fixes #(issue)
#2322 
#2333 

## Type of change

- Test jobQueueSize
Add model_config.yaml in frontend/archive/src/test/resources/models/mnist_scripted.mar for unit test.
jobQueueSize: 1

- Test jobTicket
```
cat model_store/mnist/model-config.yaml
minWorkers: 2
maxWorkers: 2
jobQueueSize: 2
useJobTicket: true
```
 seq 3 | xargs -n1 -P 3 curl http://localhost:8080/predictions/mnist_scripted_v2 -T test_data/0.png
{
  "code": 503,
  "type": "ServiceUnavailableException",
  "message": "Model \"mnist_scripted_v2\" has no worker to serve inference request. Please use scale workers API to add workers."
}

## Checklist:

- [ ] Did you have fun?


frontend/archive/src/main/java/org/pytorch/serve/archive/model/ModelConfig.java
```diff
@@ -44,6 +44,16 @@ public class ModelConfig {
      * timeout. default: 0 which means no timeout (ie. clientExpireTS default value Long.MAX_VALUE.
      */
     private long clientTimeoutInMills;
+    /**
+     * the job queue size of a model. By default, job_queue_size is set as 100 in config.property
+     * for all models. Here, jobQueueSize: -1 means no customized setting for the model.
+     */
+    private int jobQueueSize;
```
andrewjguo, line 51: For my information, where do we default it to 100?

frontend/server/src/main/java/org/pytorch/serve/wlm/BatchAggregator.java
```diff
@@ -34,6 +34,11 @@ public BaseModelRequest getRequest(String threadName, WorkerState state)
         model.pollBatch(
                 threadName, (state == WorkerState.WORKER_MODEL_LOADED) ? 0 : Long.MAX_VALUE, jobs);
 
+        if (model.isUseJobTicket() && jobs.isEmpty()) {
+            model.decNumJobTickets();
```
andrewjguo, line 38: Will the `NumJobTickets` be negative if simply keep reducing the number

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
```diff
@@ -94,6 +97,11 @@ public Model(ModelArchive modelArchive, int queueSize) {
             }
             maxRetryTimeoutInMill = modelArchive.getModelConfig().getMaxRetryTimeoutInSec() * 1000;
             clientTimeoutInMills = modelArchive.getModelConfig().getClientTimeoutInMills();
+            if (modelArchive.getModelConfig().getJobQueueSize() > 0) {
+                // overwrite the queueSize defined on config.property
+                queueSize = modelArchive.getModelConfig().getJobQueueSize();
```
andrewjguo, line 102: Do we need to worry about NPE here

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
```diff
@@ -365,4 +381,25 @@ public long getClientTimeoutInMills() {
...
+    public int decNumJobTickets() {
+        return this.numJobTickets.decrementAndGet();
+    }
+
+    public synchronized boolean hasJobTickets() {
+        if (this.numJobTickets.get() == 0) {
+            return false;
+        }
+
+        this.numJobTickets.decrementAndGet();
```
andrewjguo, line 402: Why are we decreasing here when just checking if `hasJobTickets`

frontend/archive/src/main/java/org/pytorch/serve/archive/model/ModelConfig.java
lxning: the default queueSize  100 is passed into [Model](https://github.com/pytorch/serve/blob/master/frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java#L66). This customized config value is used to overwrite the queueSize in this PR.

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
```diff
@@ -225,6 +234,10 @@ public void removeJobQueue(String threadId) {
     }
 
     public boolean addJob(Job job) {
+        if (isUseJobTicket() && !hasJobTickets()) {
```
namannandan, line 237: Will a similar check for available job tickets be required in the `public void addJob(String threadId, Job job)` method as well?

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
lxning: multi-threads can query func hasJobTickets. Decreasing in this func is used to guarantee the ticket is available for this request.

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
```diff
@@ -253,6 +266,9 @@ public void pollBatch(String threadId, long waitTime, Map<String, Job> jobsRepo)
         }
 
         try {
+            if (isUseJobTicket()) {
+                incNumJobTickets();
```
namannandan, line 270: Will a similar update to job ticket count be required in the code segment above:
```
        LinkedBlockingDeque<Job> jobsQueue = jobsDb.get(threadId);
        if (jobsQueue != null && !jobsQueue.isEmpty()) {
            Job j = jobsQueue.poll(waitTime, TimeUnit.MILLISECONDS);
            if (j != null) {
                jobsRepo.put(j.getJobId(), j);
                return;
            }
        }
```

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
namannandan: Could there potentially be a race condition here with two threads calling `hasJobTickets` simultaneously which could cause `numJobTickets` to become negative since at the time of checking both threads may not have completed the update just yet?

frontend/server/src/main/java/org/pytorch/serve/wlm/BatchAggregator.java
lxning: The ticket is increased when there is an idle worker coming to [pollbatch](https://github.com/pytorch/serve/blob/5ec2395c35a623bcae4198e5fcf1a668030ce905/frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java#L270). 

the behind logic is to increase a ticket at pollbatch; decrease a ticket at [addjob](https://github.com/pytorch/serve/blob/5ec2395c35a623bcae4198e5fcf1a668030ce905/frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java#L237). 

The corner case is there are no incoming requests. idle workers keep trying to pollbatch and get empty request. Here is used to make sure the number of tickets are always balanced (ie. +1/-1)

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
lxning: this is a synchronized func.

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
lxning: again hasJobTickets is synchronized func

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
lxning: why?

namannandan (approved):

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
namannandan: Looks like we are only tracking requests on the `DEFAULT_DATA_QUEUE`, so it won't be required for the above case, makes sense.

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
andrewjguo: I see, then a function name like `getJobTicket` or `takeJobTicket` would make more sense, as it's actually trying to reduce the ticket count.

andrewjguo (approved):

HamidShojanazeri (commented):
Thanks @lxning LGTM and I think its a great feature. It would be also great if you could add an example/ extending the LLM example for custom job queue size as well.

docs/large_model_inference.md
```diff
@@ -194,3 +194,12 @@ torchrun:
                       # gpus you wish to split your model
     OMP_NUMBER_THREADS: 2
 ```
+#### Feature Job ticket is recommended for the use case of inference latency Sensitive
+When the job ticket feature is enabled, it verifies the availability of a model's active worker for processing a client's request. If an active worker is available, the request is accepted; otherwise, a 503 response is sent back to client. Here is an example of enabling job ticket.
```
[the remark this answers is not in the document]
lxning: added more information in this doc

docs/large_model_inference.md
```diff
@@ -194,3 +194,12 @@ torchrun:
                       # gpus you wish to split your model
     OMP_NUMBER_THREADS: 2
 ```
+#### Feature Job ticket is recommended for the use case of inference latency sensitive
+When the job ticket feature is enabled, TorchServe verifies the availability of a model's active worker for processing a client's request. If an active worker is available, the request is accepted and processed immediately without waiting time incurred from job queue or dynamic batching; otherwise, a 503 response is sent back to client. This feature can help the use case which is sensitive to inference latency, such as interactive code generator, chatGPT. These applications can take effective actions,for example, routing the rejected request to a different server, or scaling up model server capacity, based on the business requirements. Here is an example of enabling job ticket.
```
HamidShojanazeri, line 198:
```suggestion
When the job ticket feature is enabled, TorchServe verifies the availability of a model's active worker for processing a client's request. If an active worker is available, the request is accepted and processed immediately without waiting time incurred from job queue or dynamic batching; otherwise, a 503 response is sent back to client.

This feature help with use-cases where inference latency can be high, such as generative models, auto regressive decoder models like chatGPT. This feature help such applications to take effective actions, for example, routing the rejected request to a different server, or scaling up model server capacity, based on the business requirements. Here is an example of enabling job ticket.
```

HamidShojanazeri (approved):

PR #2356, opened 2023-05-21
Update Telemetry env variable.

chen3933:
## Description

Fixes #(issue)

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

## Checklist:

- [ ] Did you have fun?


namannandan (approved):

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerThread.java
```diff
@@ -178,7 +178,9 @@ public void run() {
         currentThread.set(thread);
         BaseModelRequest req = null;
         int status = HttpURLConnection.HTTP_INTERNAL_ERROR;
-
+        if (java.lang.System.getenv("SM_TELEMETRY_LOG_REV_2022_12") != null) {
+            loggerTelemetryMetrics.info("Telemetry enabled.");
+        }
```
lxning, line 183: Why this log is needed at here?

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerThread.java
```diff
@@ -283,15 +285,15 @@ public void run() {
         } catch (OutOfMemoryError oom) {
             logger.error("Out of memory error when creating workers", oom);
             status = HttpURLConnection.HTTP_ENTITY_TOO_LARGE;
-            if (java.lang.System.getenv("SM_TELEMETRY_LOG") != null) {
+            if (java.lang.System.getenv("SM_TELEMETRY_LOG_REV_2022_12") != null) {
```
lxning, line 288: Why need get and check this env var everywhere? It only needs once and simply flag can be applied.

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerThread.java
```diff
@@ -283,15 +285,15 @@ public void run() {
...
                 loggerTelemetryMetrics.info(
                         "ModelServerError.Count:1|#TorchServe:{},{}:-1",
                         ConfigManager.getInstance().getVersion(),
                         oom.getClass().getCanonicalName());
             }
         } catch (Throwable t) {
             logger.warn("Backend worker thread exception.", t);
-            if (java.lang.System.getenv("SM_TELEMETRY_LOG") != null) {
+            if (java.lang.System.getenv("SM_TELEMETRY_LOG_REV_2022_12") != null) {
```
lxning, line 296: ditto

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerThread.java
chen3933: Fixed

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerThread.java
chen3933: This logging message will help us to get endpoint name using TS for query total invoke account (in the case of no error log being generated.)

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerThread.java
```diff
@@ -80,6 +80,7 @@ public class WorkerThread implements Runnable {
     private WorkerLifeCycle lifeCycle;
     private int responseTimeout;
     private long recoveryStartTS; // 0: default value. no recovery needed, in healthy mode
+    private String telemetryPath = System.getenv("SM_TELEMETRY_LOG_REV_2022_12");
```
maaquib, line 83: Do we have this env variable documented somewhere? Or is this SM internal only? We are not parsing the value here and just assuming non-null value `== true`. What if a customer sets this to `False` explicitly?

chen3933: This is a SM internal only. We use `SM_TELEMETRY_LOG_REV_2022_12 ` to avoid collision with customer defined env variable. In the case of collision or customer set this environment variable by accident, Remote appender will fail (https://github.com/aws/sagemaker-pytorch-inference-toolkit/pull/146) and we are not going to see Telemetry log on our end.

Issue #2360, opened 2023-05-24
Fail to load ONNX model from MAR file


chenkai036:
### 🐛 Describe the bug

Since this [commit](https://github.com/pytorch/serve/commit/cf7544bf2150e6c6d42a8b75f52ea31508e9382b#diff-d30e1f5ef9fd05e2ab9f652c116461632586cc132dde5d23ac4bc5cd19c799ccR83) the backend worker process dies consistently during initialising the ONNX session from a given model archive.  The refactoring introduced the bug where `self` is still used in `setup_ort_session`.

### Error logs

2023-05-24T14:33:51,640 [INFO ] W-9001-mnist_8-stdout MODEL_LOG - ONNX enabled
2023-05-24T14:33:51,642 [INFO ] W-9001-mnist_8-stdout MODEL_LOG - Backend worker process died.
2023-05-24T14:33:51,642 [INFO ] W-9001-mnist_8-stdout MODEL_LOG - Traceback (most recent call last):
2023-05-24T14:33:51,643 [INFO ] W-9001-mnist_8-stdout MODEL_LOG -   File "/home/venv/lib/python3.9/site-packages/ts/model_service_worker.py", line 253, in <module>
2023-05-24T14:33:51,643 [INFO ] W-9001-mnist_8-stdout MODEL_LOG -     worker.run_server()
2023-05-24T14:33:51,645 [INFO ] W-9001-mnist_8-stdout MODEL_LOG -   File "/home/venv/lib/python3.9/site-packages/ts/model_service_worker.py", line 221, in run_server
2023-05-24T14:33:51,646 [INFO ] W-9001-mnist_8-stdout MODEL_LOG -     self.handle_connection(cl_socket)
2023-05-24T14:33:51,650 [INFO ] W-9001-mnist_8-stdout MODEL_LOG -   File "/home/venv/lib/python3.9/site-packages/ts/model_service_worker.py", line 184, in handle_connection
2023-05-24T14:33:51,659 [INFO ] W-9001-mnist_8-stdout MODEL_LOG -     service, result, code = self.load_model(msg)
2023-05-24T14:33:51,660 [INFO ] W-9001-mnist_8-stdout MODEL_LOG -   File "/home/venv/lib/python3.9/site-packages/ts/model_service_worker.py", line 131, in load_model
2023-05-24T14:33:51,661 [INFO ] W-9001-mnist_8-stdout MODEL_LOG -     service = model_loader.load(
[...]

PR #2363, opened 2023-05-25
Enable ONNX test in CI

msaroufim:
ONNX is one of our most popular model optimization runtimes, we shouldn't disable this test in CI by default anymore, it's hiding basic failures

Our CI takes 40min and our regression tests take 1h and for context the PyTorch CI takes around 4h. Would just suggest we run regression tests on CI all the time otherwise we get dumb bugs

Fixes https://github.com/pytorch/serve/issues/2360


requirements/developer.txt
```diff
@@ -16,3 +16,6 @@ twine
 mypy
 torchpippy
 intel_extension_for_pytorch; sys_platform != 'win32' and sys_platform != 'darwin'
+onnx-runtime
+psutil
```
msaroufim, line 20: adding psutil and pyyaml here since otherwise they're not being installed, not sure why yet

agunapal (approved):

PR #2378, opened 2023-06-01
Fix INF2 example handler

namannandan:
## Description
Fix handler for Inferentia2 example to handle partial batches and calls to `transformers-neuronx` library api

## Type of change

- [ ] Successful manual test for `125m` parameter variant of the `opt` model
```
$ cat sample_text.txt 
Today the weather is really nice and I am planning on
$ curl http://127.0.0.1:8080/predictions/opt-125m -T sample_text.txt 
Today the weather is really nice and I am planning on
going through some camping next weekend. I thought if I didn’t
wear rain gear on Saturday maybe I would have to wear a coat but
I am using my wet shoes
```
- [ ] Test for the `6.7b` parameter variant of `opt` model [in progress]
```
$ cat sample_text.txt 
Today the weather is really nice and I am planning on
$ curl http://127.0.0.1:8080/predictions/opt-6.7b -T sample_text.txt 
Today the weather is really nice and I am planning on
spending the day in the park and riding my bike to the
store for ice cream. Then I will come home and study
more Spanish. It is raining and sunny here,
```

### DIFF
--- a/examples/large_models/inferentia2/inf2_handler.py
+++ b/examples/large_models/inferentia2/inf2_handler.py
@@ -115,7 +115,7 @@ class LLMHandler(BaseHandler, ABC):
         # insert padding if a partial batch was received
         num_inferences = len(input_ids_batch)
         logger.info("Number of inference requests in batch: %s", num_inferences)
-        logger.info("Batch size: %s", self.batch_size)
+        logger.info("Model batch size: %s", self.batch_size)
         padding = self.batch_size - num_inferences
         if padding > 0:
             logger.info("Padding input batch with %s padding inputs", padding)
--- a/examples/large_models/inferentia2/inf2_handler.py
+++ b/examples/large_models/inferentia2/inf2_handler.py
@@ -36,7 +36,6 @@ class LLMHandler(BaseHandler, ABC):
 
         # settings for model compiliation and loading
         seed = ctx.model_yaml_config["handler"]["manual_seed"]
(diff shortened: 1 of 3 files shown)


examples/large_models/inferentia2/inf2_handler.py
```diff
@@ -112,17 +111,27 @@ def inference(self, input_batch):
...
         """
         input_ids_batch, attention_mask_batch = input_batch
-        input_ids_batch = input_ids_batch
+
+        # insert padding if a partial batch was received
+        num_inferences = len(input_ids_batch)
+        logger.info("Input ids batch: %s", input_ids_batch)
+        logger.info("Num inferences: %s", num_inferences)
+        padding = self.batch_size - num_inferences
+        if padding > 0:
```
HamidShojanazeri, line 120: Thansk @namannandan, I am wondering if it would be better to ask user to make sure register model with batch size consistent with what they have compiled the model with (adding this in README) instead of appending dummy inputs to the batch.

namannandan: @HamidShojanazeri Yes, that should work too. Although adding support for padding would make the model compatible with the batch aggregator in torchserve.

HamidShojanazeri: I see, sure, lets then add some loggings to make user know when this is happening. I already updated the README as well.

HamidShojanazeri (approved):
LGTM thanks @namannandan

PR #2383, opened 2023-06-01
Speed up unit tests by removing sleep in start/stop torchserve

mreso:
## Description

This PR removes sleep(10) in our unit test utils and replaces it with a function which parses the log to wait the actual time it takes to start torchserve. This speeds up single unit tests by up to 4x.

## Type of change

- [X] Run  `pytest test/pytest/test_example_micro_batching.py` before and after change

Before:
Screenshot 2023-06-01 at 1 39 44 PM

After:
Screenshot 2023-06-01 at 1 40 29 PM

## Checklist:

- [X] Did you have fun?

### DIFF
--- a/ts_scripts/tsutils.py
+++ b/ts_scripts/tsutils.py
@@ -1,52 +1,88 @@
 import os
 import platform
 import sys
-import time
+import threading
+from subprocess import PIPE, STDOUT, Popen
+
 import requests
+
 from ts_scripts import marsgen as mg
 
 torchserve_command = {
     "Windows": "torchserve.exe",
     "Darwin": "torchserve",
-    "Linux": "torchserve"
+    "Linux": "torchserve",
 }
 
 torch_model_archiver_command = {
-        "Windows": "torch-model-archiver.exe",
-        "Darwin": "torch-model-archiver",
-        "Linux": "torch-model-archiver"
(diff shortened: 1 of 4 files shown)


ts_scripts/tsutils.py
```diff
@@ -1,66 +1,105 @@
...
-        "Darwin": "torch-workflow-archiver",
-        "Linux": "torch-workflow-archiver"
-    }
+    "Windows": "torch-workflow-archiver.exe",
+    "Darwin": "torch-workflow-archiver",
+    "Linux": "torch-workflow-archiver",
+}
+
+
+class LogPipeTillTheEnd(threading.Thread):
```
msaroufim, line 31: Is the assumption that each new run of torchserve will overwite its logs?

mreso: The tests still run sequentially and nothing is overwritten. This class only prints the logs even after TS initialization is completed. This replicates the same behavior as we had before where we ran TS in a background process and waited for 10 seconds. Now we parse the stdout pipe for the initialization completed signal and then just let this class print the rest of the log. This way folks can still see and debug issues as before.

msaroufim (approved):

agunapal (approved):

PR #2403, opened 2023-06-09
Docker Regression Tests Github action

agunapal:
## Description

This PR creates a github action to build TorchServe Docker Images and run regression tests inside a docker container

To make this happen,
  - Create a new target for building docker ci image
  - Some of the tests don't run as is inside the container. These need to be fixed. Skipping them for now
  - Run docker regression tests nightly as they take longer to run. Each run takes around 1.5 hours
  
 Fix these Pytests

- [ ] https://github.com/pytorch/serve/issues/2460
- [ ] https://github.com/pytorch/serve/issues/2461 
- [ ] https://github.com/pytorch/serve/issues/2462

Fixes #(issue)

## Type of change

[2_docker-regression (self-hosted, regression-test-gpu).txt](https://github.com/pytorch/serve/files/11941058/2_docker-regression.self-hosted.regression-test-gpu.txt)
[1_docker-regression (ubuntu-20.04).txt](https://github.com/pytorch/serve/files/11941059/1_docker-regression.ubuntu-20.04.txt)

## Checklist:

- [ ] Did you have fun?


test/pytest/test_example_intel_extension_for_pytorch.py
```diff
@@ -88,6 +88,10 @@ def scale_workers_with_core_pinning(scaled_num_workers):
     or ((torch.cuda.device_count() > 0) and torch.cuda.is_available()),
     reason="Make sure intel-extension-for-pytorch is installed and torch.backends.xeon.run_cpu is available",
 )
+@pytest.mark.skipif(
+    os.environ.get("AM_I_IN_A_DOCKER_CONTAINER", False),
```
msaroufim, line 92: can we give this a shorter name? :)

test/pytest/test_torch_compile.py
```diff
@@ -64,6 +64,10 @@ def test_start_torchserve(self):
         assert len(glob.glob("logs/model_log.log")) == 1
         assert len(glob.glob("logs/ts_log.log")) == 1
 
+    @pytest.mark.skipif(
```
msaroufim, line 67: Does this not work? What's the error you get?

Same comment for all of the failing tests curious what's the problem

.github/workflows/regression_tests_docker.yml
```diff
@@ -0,0 +1,52 @@
...
+concurrency:
+  group: ci-cpu-${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.run_number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  docker-regression:
+    strategy:
+      fail-fast: false
+      matrix:
+        hardware: [ubuntu-20.04, [self-hosted, regression-test-gpu]]
```
msaroufim, line 17: how does the doubly nested list work here?

docker/Dockerfile.base
```diff
@@ -0,0 +1,114 @@
+# syntax = docker/dockerfile:experimental
```
msaroufim, line 1: Docs needd to be updated if we have a new base dockerfile, usually convention is Dockerfile means base

docker/build_image.sh
```diff
@@ -130,17 +131,31 @@ then
...
+# Build docker base image
+DOCKER_BUILDKIT=1 docker build --file Dockerfile.base --build-arg BASE_IMAGE="${BASE_IMAGE}" --build-arg CUDA_VERSION="${CUDA_VERSION}"  --build-arg PYTHON_VERSION="${PYTHON_VERSION}" -t $DOCKER_BASE_IMAGE .
+
 if [ "${BUILD_TYPE}" == "production" ]
 then
   DOCKER_BUILDKIT=1 docker build --file Dockerfile --build-arg BASE_IMAGE="${BASE_IMAGE}" --build-arg CUDA_VERSION="${CUDA_VERSION}"  --build-arg PYTHON_VERSION="${PYTHON_VERSION}" -t "${DOCKER_TAG}" .
 elif [ "${BUILD_TYPE}" == "benchmark" ]
 then
   DOCKER_BUILDKIT=1 docker build --pull --no-cache --file Dockerfile.benchmark --build-arg USE_LOCAL_SERVE_FOLDER=$USE_LOCAL_SERVE_FOLDER --build-arg BASE_IMAGE="${BASE_IMAGE}" --build-arg BRANCH_NAME="${BRANCH_NAME}" --build-arg CUDA_VERSION="${CUDA_VERSION}" --build-arg MACHINE_TYPE="${MACHINE}" --build-arg PYTHON_VERSION="${PYTHON_VERSION}" -t "${DOCKER_TAG}" .
+elif [ "${BUILD_TYPE}" == "ci" ]
```
msaroufim, line 153: How is ci different from dev? Can't we just have those be the same?

docker/Dockerfile.ci
```diff
@@ -0,0 +1,41 @@
...
+    zip \
+    unzip \
+    && npm install -g newman newman-reporter-htmlextra markdown-link-check \
+    && rm -rf /var/lib/apt/lists/* \
+    && cd /tmp
+
+
+ENV PATH="/home/venv/bin:$PATH"
+
+RUN python -m pip install --no-cache-dir -r https://raw.githubusercontent.com/pytorch/serve/master/requirements/developer.txt
```
msaroufim, line 35: We might have been doing this previously but this is incorrect, it should be cloning the repo and pulling in a specific branch

Otherwise testing this locally is not meaningful

msaroufim (changes requested):

test/pytest/test_torch_compile.py
agunapal: The curl commands don't work. All the tests which are using curl commands need to be refactored using requests

docker/build_image.sh
agunapal: dev will have torchserve starting when container is run. 
ci is running the regression tests script on start

docker/Dockerfile.base
agunapal: Dockerfile is still the main file for building official TorchServe Images. Its not referenced anywhere in the docs . But, let me add a comment that the build script is first building a base image and then building the actual image.

agunapal: Also, this split is necessary for the docker optimization PR. With that PR, we will also have an additional split to differentiate cpu vs gpu

docker/Dockerfile.base
```diff
@@ -0,0 +1,114 @@
...
+WORKDIR "serve"
+
+RUN \
+    if echo "$BASE_IMAGE" | grep -q "cuda:"; then \
+        # Install CUDA version specific binary when CUDA version is specified as a build arg
+        if [ "$CUDA_VERSION" ]; then \
+            python ./ts_scripts/install_dependencies.py --cuda $CUDA_VERSION; \
+        # Install the binary with the latest CPU image on a CUDA base image
+        else \
+            python ./ts_scripts/install_dependencies.py; \
```
mreso, line 77: I still think there is no real use case for cpu Pytorch in a gpu docker image and we should default to current most supported version.

docker/Dockerfile.base
```diff
@@ -0,0 +1,114 @@
...
+# Re-state ARG PYTHON_VERSION to make it active in this build-stage (uses default define at the top)
+ARG PYTHON_VERSION
+ENV PYTHONUNBUFFERED TRUE
+
+RUN --mount=type=cache,target=/var/cache/apt \
+    apt-get update && \
+    apt-get upgrade -y && \
+    apt-get install software-properties-common -y && \
+    add-apt-repository -y ppa:deadsnakes/ppa && \
+    apt remove python-pip  python3-pip && \
```
mreso, line 99: What is the intention of this line? Do we remove pip here because the version is creating problems down the line?

docker/build_image.sh
mreso: hmm, can we at least have the root of both dev and regression images be the same? I.e. having a common dev base and then just add the expose ports and start torchserve separatelty?
Speaking of this. What is the intention of the docker regression test? For me it is to test the functionality of TS inside the docker files we deploy? In that case we should stay as close to the actual images we deploy as possible. Ideally using the actually deployed images as  base and adding onmly stuff needed for the regression test on top to avoid diversion over time.

mreso: For me it should loo like
```
Dockerfile -> Docker.run_time
         L -> Docker.dev -> Docker.regression (of use run_time as base for this)
```

test/pytest/test_example_scriptable_tokenzier.py
```diff
@@ -318,7 +318,7 @@ def test_inference_with_pretrained_model(model_store, test_file, torchserve):
     assert "Positive" in result_entries
 
     assert float(result_entries["Negative"]) == pytest.approx(
-        0.0001851904089562595, 1e-3
+        0.0001851904089562595, abs=1e-6
```
mreso, line 321: Why not apply this uniformly to 323 as well?

test/pytest/test_torch_compile.py
mreso: Is this because curl is not installed or not working?

test/regression_tests.py
```diff
@@ -1,28 +1,23 @@
...
-from ts_scripts import marsgen as mg
-
-import datetime
 
 now = datetime.datetime.now()
 print("Current date and time : " + now.strftime("%Y-%m-%d %H:%M:%S"))
 
 check_python_version()
 
-git_branch = Repository('.').head.shorthand
```
mreso, line 23: Whats the intention behind this change?

mreso (changes requested):

test/pytest/test_example_intel_extension_for_pytorch.py
agunapal: Done

test/pytest/test_torch_compile.py
agunapal: curl is installed. curl to localhost doesn't work. Tried various things like --network=host etc. Not sure. We can address this in a different PR

docker/Dockerfile.base
agunapal: Done

docker/build_image.sh
agunapal: We can. I didn't do it yet since we plan to release next week. If we can push the release by a few weeks, I can test various combinations.

docker/Dockerfile.ci
agunapal: Done

docker/Dockerfile.base
```diff
@@ -0,0 +1,114 @@
...
+# NOTE: To build this you will need a docker version > 18.06 with
+#       experimental enabled and DOCKER_BUILDKIT=1
+#
+#       If you do not use buildkit you are not going to have a good time
+#
+#       For reference:
+#           https://docs.docker.com/develop/develop-images/build_enhancements/
+
+
+ARG BASE_IMAGE=ubuntu:rolling
```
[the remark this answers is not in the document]
agunapal: This is going to change in my next PR to slim

docker/Dockerfile.base
agunapal: Yes, will address this in the PR to reduce CPU/GPU docker images

docker/Dockerfile.base
agunapal: Yes, this was done by another user who refactored our Dockerfile and made sure things work.

test/pytest/test_example_scriptable_tokenzier.py
agunapal: Done. Somehow the pytest showed a failure only here

test/regression_tests.py
agunapal: This is now working inside the docker container. Also, I'm not sure if this is adding any value. We should remove it.

docker/Dockerfile.base
```diff
@@ -0,0 +1,114 @@
...
+# Define here the default python version to be used in all later build-stages as default.
+# ARG and ENV variables do not persist across stages (they're build-stage scoped).
+# That is crucial for ARG PYTHON_VERSION, which otherwise becomes "" leading to nasty bugs,
+# that don't let the build fail, but break current version handling logic and result
+# in images with wrong python version. To fix that, we will restate the ARG PYTHON_VERSION
+# on each build-stage.
+ARG PYTHON_VERSION=3.9
+
+FROM ${BASE_IMAGE} AS compile-image
+ARG BASE_IMAGE=ubuntu:rolling
```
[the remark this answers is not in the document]
agunapal: Will be resolved in the CPU/GPU optimization PR

.github/workflows/regression_tests_docker.yml
```diff
@@ -0,0 +1,56 @@
+name: Run Regression Tests on Docker
+
+on:
+  push
+    # run every day at 5:15am
```
msaroufim, line 5: nit: remove comments

.github/workflows/regression_tests_docker.yml
```diff
@@ -0,0 +1,56 @@
...
+concurrency:
+  group: ci-cpu-${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.run_number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  docker-regression:
+    strategy:
+      fail-fast: false
+      matrix:
+        hardware: [ubuntu-20.04, [self-hosted, regression-test-gpu]]
```
msaroufim, line 18: why are there 2 github actions? This matrix seems to indicate that you're running both GPU and CPU tests here?

.github/workflows/regression_tests_docker.yml
```diff
@@ -0,0 +1,56 @@
...
+          sudo rm -rf ./* || true
+          sudo rm -rf ./.??* || true
+          ls -la ./
+          docker system prune -f
+      - name: Checkout TorchServe
+        uses: actions/checkout@v3
+      - name: Branch name
+        run: |
+          echo $GITHUB_REF_NAME
+      - name: Build CPU Docker Image
```
msaroufim, line 35: Is there now enough memory on non custom runners to build both images? My main worry is that subsequent steps should fail if this step fails

docker/Dockerfile
```diff
@@ -85,7 +85,7 @@ RUN \
 RUN python -m pip install --no-cache-dir torchserve torch-model-archiver torch-workflow-archiver
 
 # Final image for production
-FROM ${BASE_IMAGE} AS runtime-image
+FROM ${BASE_IMAGE} AS production-image
```
msaroufim, line 88: do you mean nvidia production image? if so could you add a comment to make this clearer

docker/Dockerfile
```diff
@@ -130,3 +130,48 @@ WORKDIR /home/model-server
...
+    && npm install -g newman newman-reporter-htmlextra markdown-link-check \
+    && rm -rf /var/lib/apt/lists/* \
+    && cd /tmp
+
+
+COPY --from=compile-image /home/venv /home/venv
+
+ENV PATH="/home/venv/bin:$PATH"
+
+RUN python -m pip install --no-cache-dir -r https://raw.githubusercontent.com/pytorch/serve/$BRANCH_NAME/requirements/developer.txt
```
msaroufim, line 171: this only works if the branch has been pushed to our repo, IMO this should just use the requirements that are present locally. We want to make sure to support people that want to build our images

docker/Dockerfile
```diff
@@ -130,3 +130,48 @@ WORKDIR /home/model-server
...
+
+RUN mkdir /home/serve
+ENV RUN_IN_DOCKER True
```
msaroufim, line 174: can you give this a longer name to make sure risk of name conflicts with user code is very low?

docker/Dockerfile
```diff
@@ -130,3 +130,48 @@ WORKDIR /home/model-server
...
+
+WORKDIR /home/serve
+CMD ["python", "test/regression_tests.py"]
```
msaroufim, line 177: do we need to do this? This will break BC for people that used to use our dockerfiles. I was thinking we could just docker run in interactive mode for our tests

examples/dcgan_fashiongen/create_mar.sh
```diff
@@ -16,7 +16,6 @@ function cleanup {
 trap cleanup EXIT
 
 # Download and Extract model's source code
-sudo apt-get install zip unzip -y
```
msaroufim, line 19: is this some other PR?

test/pytest/test_example_intel_extension_for_pytorch.py
```diff
@@ -88,6 +88,10 @@ def scale_workers_with_core_pinning(scaled_num_workers):
     or ((torch.cuda.device_count() > 0) and torch.cuda.is_available()),
     reason="Make sure intel-extension-for-pytorch is installed and torch.backends.xeon.run_cpu is available",
 )
+@pytest.mark.skipif(
```
msaroufim, line 91: Could you please create an issue tracker with the skipped tests, these tests are not great but they're all important dependencies ONNX, worker thread affinity and compile

test/regression_tests.py
```diff
@@ -1,28 +1,23 @@
...
-build_hdr_printer.main(git_branch)
```
msaroufim, line 24: are changes to this file from another PR?

msaroufim (changes requested):

.github/workflows/regression_tests_docker.yml
agunapal: Yes, this is temporary. Will update this once we have the PR approved. We don't have a good way to run this

.github/workflows/regression_tests_docker.yml
agunapal: These will be executed on 2 different runners. One for CPU and one for GPU

docker/Dockerfile
agunapal: Done.Changed to `TS_RUN_IN_DOCKER`

.github/workflows/regression_tests_docker.yml
```diff
@@ -0,0 +1,56 @@
...
+        if: contains(matrix.hardware, 'ubuntu')
+        run: |
+          cd docker
+          ./build_image.sh -bt ci -b $GITHUB_REF_NAME -t pytorch/torchserve:ci
+      - name: Build GPU Docker Image
+        if: false == contains(matrix.hardware, 'ubuntu')
+        run: |
+          cd docker
+          ./build_image.sh -g -cv cu117 -bt ci -b $GITHUB_REF_NAME -t pytorch/torchserve:ci
```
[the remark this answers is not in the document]
agunapal: They are on different machines and they are temporary. It should be fine

docker/Dockerfile
agunapal: No, its TorchServe's production image. We are making use of docker tags to build different images vs having multiple Dockerfile. There is a comment already on L87

docker/Dockerfile
agunapal: It won't break because there is a target specified here https://github.com/pytorch/serve/blob/issues/fix_regression_tests/docker/build_image.sh#L140

examples/dcgan_fashiongen/create_mar.sh
agunapal: Initially, I ended up fixing one of the failing tests to make it work in docker. Thought I'll keep it. Should I create a separate PR for this?

test/pytest/test_example_intel_extension_for_pytorch.py
agunapal: Sure, will create and update the PR with issues. But, we cannot test ONNX because of NVIDIA base image

agunapal: I intend to address this when I update Dockerfile.dev, where we can possibly build with nvidia runtime image

test/regression_tests.py
agunapal: Because we are passing TS repo as a shared volume to Docker container, it was erroring on L23. I can keep it and check for the ENV variable. Should I do this?

docker/Dockerfile
agunapal: I'm ok either way. This would mean going back to using git clone. Should I make the change?

test/pytest/test_example_intel_extension_for_pytorch.py
agunapal: Just noticed that ONNX test case is passing in docker :D (It shouldn't)

test/regression_tests.py
agunapal: or we can use this opportunity to get rid of pygit2?

examples/dcgan_fashiongen/create_mar.sh
msaroufim: Na it's fine

test/regression_tests.py
msaroufim: No iz fine

docker/Dockerfile
msaroufim: Yeah git clone is fine, you can use `--depth` to control size

msaroufim (approved):
Approved assuming CI is green on those images - for the failing test tracker please create a seperate github issue

.github/workflows/regression_tests_docker.yml
```diff
@@ -0,0 +1,55 @@
...
+      matrix:
+        hardware: [ubuntu-20.04, [self-hosted, regression-test-gpu]]
+    runs-on:
+      - ${{ matrix.hardware }}
+    steps:
+      - name: Clean up previous run
+        run: |
+          echo "Cleaning up previous run"
+          ls -la ./
+          sudo rm -rf ./* || true
```
namannandan, line 25: Nit: Why do we need the `|| true` part? Does this command fail in certain conditions and we want to ignore it?

.github/workflows/regression_tests_docker.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: |
+          cd docker
+          ./build_image.sh -bt ci -b $GITHUB_REF_NAME -t pytorch/torchserve:ci
+      - name: Build GPU Docker Image
+        if: false == contains(matrix.hardware, 'ubuntu')
+        run: |
+          cd docker
+          ./build_image.sh -g -cv cu117 -bt ci -b $GITHUB_REF_NAME -t pytorch/torchserve:ci
+      - name: Torchserve GPU Regression Tests
+        if: false == contains(matrix.hardware, 'ubuntu')
```
namannandan, line 45: Nit: Would it be better check for the specific GPU runner attributes/labels instead of absence of `ubuntu`?

namannandan (approved):

.github/workflows/regression_tests_docker.yml
agunapal: Yes, there are weird permission issues because of pycache geenrated by docker run. This logic worked for me. Will have to monitor the runs for the next few days

.github/workflows/regression_tests_docker.yml
agunapal: I tried, the problem is that we need to check for the entire thing: [self.hosted, ...] . I couldnt get it to work , hence went for the negation logic.

PR #2445, opened 2023-07-05
Update PyPiDescription.rst

bryanwweber:
Fix the link to the README on PyPI. Use python -m pip for installation.

## Checklist:

- [X] Did you have fun?


msaroufim (approved):

bryanwweber:
I'm sorry, I didn't provide any context for these changes! My mistake, I submitted the PR and then forgot about it.

I was browsing PyPI and noticed that the content formatting is not correct: 
image
I believe at some point, PyPI defaulted to assuming that the `long_description` is Markdown formatted, so the `setup.py` changes fix the text format type so that the README will be displayed properly. In addition, the `url` link is missing the `blob/master` portion, so it returns 404 from GitHub. This only affects the URL displayed in the PyPI user interface, it it is correct for the root-level `setup.py`. Finally, the same problem occurs in the `PypiDescription.rst` as far as the link is concerned.

Thanks for reviewing!

agunapal (approved):

PR #2488, opened 2023-07-21
Ensure that self.manifest is defined before call profiler

### LINKED ISSUE #2480: Read manifest inside _infer_with_profiler method
### 🚀 The feature

Add a line to read the manifest file inside the [method ](https://github.com/pytorch/serve/blob/754c2f9a550ff57790e586c199dc8b7a842af159/ts/torch_handler/base_handler.py#L354-L401) `_infer_with_profiler` if `self.manifest` is None (this can happen in cases the user overwrites the `initialize` method).

If you think this makes sense, Im more than happy to contribute with it :)

### Motivation, pitch

I'm working in some projects where we are overwriting the `initialize` method from base handler and we weren't creating the `self.manifest` attribute on it. When I tried to run the torch profiler, the request broke because `self.manifest` was None. IMO, as this attribute is used for other methods we should guarantee that we will have that or if the attribute is None we can read the file inside the `_infer_with_profiler` method.

This issue will happen to some examples too, like [this one](https://github.com/pytorch/serve/blob/7e5857fee8eb502559da97ff808041749282e868/examples/text_to_speech_synthesizer/waveglow_handler.py#L55C1-L55C1).

### Alternatives

- If `self.manifest` is None, read the file inside `_infer_with_profiler`;
- Update documentation to be explicit that you need to have the `self.manifest` defined in your `initialize` method.

### Additional context

_No response_

pedrogengo:
## Description

Fixes #2480

## Type of change

- Unit test added for base handler, where I set the `self.manifest` as None, imagining a scenario where someone created a custom `initialize` method and didn't defined `self.manifest`, and validate if we can run `_infer_with_profiler` even in this case.

  - Before the change:
  image

  - After the change
  image

## Checklist:

- [x] Did you have fun?

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Refer to [torchserve docker](docker/README.md) for details.
 
 
 ## 🏆 Highlighted Examples
-* [🤗 HuggingFace Transformers](examples/Huggingface_Transformers) with a [Better Transformer Integration](examples/Huggingface_Transformers#Speed-up-inference-with-Better-Transformer)
+* [🤗 HuggingFace Transformers](examples/Huggingface_Transformers) with a [Better Transformer Integration/ Flash Attention & Xformer Memory Efficient ](examples/Huggingface_Transformers#Speed-up-inference-with-Better-Transformer)
 * [Model parallel inference](examples/Huggingface_Transformers#model-parallelism)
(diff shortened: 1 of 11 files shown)


agunapal (approved):
Thanks for your contribution! LGTM

pedrogengo:
Tks! I will just fix the lint issue and commit again

msaroufim (approved):

PR #2516, opened 2023-08-04
Updated example for metrics and backwards compatibility warnings

namannandan:
## Description
Provide an updated example to show usage of custom metrics APIs and the `prometheus` metrics mode.
Also update metrics documentation to include warnings about backward compatibility with custom metrics APIs in prior versions.

## Type of change

- [ ] Results of manually running the example have been included in the example documentation


examples/custom_metrics/torchserve_custom.mtail
```diff
@@ -1,24 +0,0 @@
-counter request_count
```
msaroufim, line 1: why was this deleted?

docs/metrics.md
```diff
@@ -622,3 +628,23 @@ class CustomHandlerExample:
         # except this time with gauge metric type object
         metrics.add_size("GaugeModelMetricNameExample", 42.5)
 ```
+
+## Backwards compatibility warnings
+1. Starting [v0.6.1](https://github.com/pytorch/serve/releases/tag/v0.6.1), the `add_metric` API signature changed\
+   from [add_metric(name, value, unit, idx=None, dimensions=None)](https://github.com/pytorch/serve/blob/61f1c4182e6e864c9ef1af99439854af3409d325/ts/metrics/metrics_store.py#L184)\
```
msaroufim, line 634: so the old metrics were always counters? If that's the case then keeping BC automatically shouldn't be too hard

examples/custom_metrics/README.md
```diff
@@ -1,93 +1,164 @@
...
+- Step 3: Register the model to torchserve using the above model archive file.
 
   ```bash
   mkdir model_store
   mv mnist.mar model_store/
-  torchserve --start --model-store model_store --models mnist=mnist.mar
+  torchserve --ncs --start --model-store model_store --ts-config examples/custom_metrics/config.properties --models mnist=mnist.mar
   ```
 
-- Step 4: Install [mtail](https://github.com/google/mtail/releases)
```
msaroufim, line 38: ah i see, maybe keep a note as well in bc change that mtail is no longer needed

msaroufim (changes requested):
The doc looks fine but I'm not convinced this is the right BC story

Here are some other options I see
1. Keep BC guarantee: for example if old metric API is always a counter then it shouldn't be too hard to make the old function call the new function
2. If it's not possible to get BC guarantee (I'd like to understand why this is the case a bit more): then also make sure in code when the old metrics API is called to print a warning once 

Regardless of if we go for 1 or 2 it's helpful to explain to users why they should migrate and what are the benefits to them of doing so, one example might be more metric types but worth clarifying this more

docs/metrics.md
namannandan: The [prior metrics implementation](https://github.com/pytorch/serve/blob/96450b9d0ab2a7290221f0e07aea5fda8a83efaf/ts/metrics/metric.py#L16-L29) did not have `types` associated with them. The [new metrics implementation](https://github.com/pytorch/serve/blob/96450b9d0ab2a7290221f0e07aea5fda8a83efaf/ts/metrics/caching_metric.py#L24) does add support for [MetricTypes](https://github.com/pytorch/serve/blob/96450b9d0ab2a7290221f0e07aea5fda8a83efaf/ts/metrics/metric_type_enum.py#L7).

In addition, while the prior metrics implementation did not have a way to specify metrics and their specifications (name, unit, dimension names and type) in a central configuration file, the new metrics implementation introduced this, as a result which, the semantics of `add_metric` method was changed
**from**: [Create a metric object and store in a list to emit](https://github.com/pytorch/serve/blob/96450b9d0ab2a7290221f0e07aea5fda8a83efaf/ts/metrics/metrics_store.py#L184)
**to**: [Add a metric object consisting only of its specifications (name, unit, dimension names and type) to a metics cache](https://github.com/pytorch/serve/blob/96450b9d0ab2a7290221f0e07aea5fda8a83efaf/ts/metrics/metric_cache_yaml_impl.py#L97-L103). The dimension values are provided at the time of updating a metric using the [add_or_update](https://github.com/pytorch/serve/blob/96450b9d0ab2a7290221f0e07aea5fda8a83efaf/ts/metrics/caching_metric.py#L107-L112) method.

namannandan: Couple of options to ensure backwards compatibility are as follows:
1) Introduce a new method in [metric_cache_abstract.py](https://github.com/pytorch/serve/blob/96450b9d0ab2a7290221f0e07aea5fda8a83efaf/ts/metrics/metric_cache_abstract.py), say `add_metric_bc` which has the same signature as that of the old [add_metric](https://github.com/pytorch/serve/blob/61f1c4182e6e864c9ef1af99439854af3409d325/ts/metrics/metrics_store.py#L184) API. This method can internally call `add_metric` and then `add_or_update` on the metric object. The default metric type in this case would be `counter`.
2) Change the name of the new `add_metric` method to `add_metric_to_cache` and reimplement the `add_metric` method to have the same signature as the old implementation. Then, the `add_metric` API can internally call `add_metric_to_cache` method and the `add_or_update` on the metric object.

Please share your thoughts on these approaches.

msaroufim: I like 2, `add_metric_to_cache()` seems more like an internal detail whereas what a user wants to do is `add_metric()`.

namannandan: Draft PR to implement backwards compatibility for `add_metric` API: https://github.com/pytorch/serve/pull/2525

duk0011:
From user perspective, I like option 2 (**change the name of the new add_metric method to add_metric_to_cache and reimplement the add_metric method to have the same signature as the old implementation**). This will help with version upgrade with backward compatibility.

msaroufim (approved):

namannandan:
Thanks for the feedback @msaroufim and @duk0011 
PR: https://github.com/pytorch/serve/pull/2525 to make `add_metric` backwards compatible has been merged.
This PR includes the following:
- Custom metrics API documentation updates
- Backwards compatibility warnings
- Upgrade path documentation
- Updated example

@duk0011 please note the following in addition to the backwards compatibility implemented in https://github.com/pytorch/serve/pull/2525:
- `add_metric` is now backwards compatible but the default metric type is inferred to be `COUNTER`. If the metric is of a different type, it will need to be specified in the call to `add_metric` as follows:
`metrics.add_metric(name='GenericMetric', value=10, unit='count', dimensions=[...], metric_type=MetricTypes.GAUGE)`
- All custom metrics updated in the custom handler will need to be included in the metrics configuration file for them to be emitted by Torchserve. This is shown [here](https://github.com/pytorch/serve/blob/58eb2d2c79cf1cf711abf9ffea5678420a5ff65a/docs/metrics.md#central-metrics-yaml-file-definition).

msaroufim:
SGTM let's explicitly mention your note when we do the patch release next week

lxning (approved):

PR #2565, opened 2023-09-03
Add support for using Zstandard encoded payload for Inference.

abergmeier:
With LSTMs we found that payload gets quite large - larger than the hosting platform accepts.
An easy workaround for that is to simply compress the payload.

Currently this implements only zstd support.

Adds dependency on `aircompressor_version`.

## Description

> Please read our [CONTRIBUTING.md](https://github.com/pytorch/serve/blob/master/CONTRIBUTING.md) prior to creating your first pull request.

## Type of change

> Please describe the Unit or Integration tests that you ran to verify your changes and relevant result summary. Provide instructions so it can be reproduced.

- [x] testPredictionsModifyResponseHeaderZstd
- [x] testZstdSimple
- [x] testZstdLarge 

## Checklist:

- [ ] Did you have fun?


abergmeier:
@fabridamicelli wdyt?

abergmeier:
Could use a hand with `testPredictionsModifyResponseHeaderZstd`. Cannot seem to get local test execution to work.

abergmeier:
Have a version of decompressing in Java. Maybe better than to have to modify Java -> Python codec....

mreso:
Hi @abergmeier , thanks for your contribution! Could you elaborate a bit on why the compression needs to be integrated into FE or otf protocol? Would it not be more flexible to implement it in the handler and perform decompression during preprocessing?

abergmeier:
> Hi @abergmeier , thanks for your contribution! Could you elaborate a bit on why the compression needs to be integrated into FE or otf protocol? Would it not be more flexible to implement it in the handler and perform decompression during preprocessing?

The goal with this implementation was/is to be able to use standards over having to implement code.
So in our particular example, we have an (envoy) proxy, which intercepts the data sent to FE before it leaves the sending computer and compresses it via zstd (also setting the necessary HTTP headers).
No code needs to be written for that compression - it is all just a matter of configuration.

Supporting decompression in the FE enables any user to send larger payloads and save bandwidth costs while at the same time also being able to use any software that produces standardized HTTP data using compression/mime-types.

abergmeier:
So seems like `master` does not work yet again in CI - how to handle this?

msaroufim:
Master has been healthy last few commits https://github.com/pytorch/serve/commits/master

Main flaky job is a workflow test in postman in regression cpu that @agunapal was looking at disabling

abergmeier:
We migrated away from torchserve - so no longer necessary.

PR #2586, opened 2023-09-11
Update docker nightly workflow to use GPU runner since these have more space

agunapal:
## Description

Update docker nightly build to use CI GPU runner
- Cleanup built docker images

Fixes #([issue](https://github.com/pytorch/serve/issues/2583))

## Type of change

Passing Run
https://github.com/pytorch/serve/actions/runs/6152373521/job/16694390169?pr=2586

## Checklist:

- [ ] Did you have fun?


agunapal:
@msaroufim To test this, I can do make a temporary edit to run this workflow. But, this would also push the builds. Do we want to do this?

docker/docker_nightly.py
```diff
@@ -55,3 +60,11 @@
     # Push images with latest tag
     try_and_handle(f"docker push {organization}/{project}:latest-cpu", dry_run)
     try_and_handle(f"docker push {organization}/{project}:latest-gpu", dry_run)
+
+    # Cleanup built images
+    if args.cleanup:
```
msaroufim, line 65: is this needed? 

I think something like `docker system prune --all --volumes` should work instead, otherwise now people might miss a few chars here and there

There are some well known docker nuke commands online, maybe we should bring something in

msaroufim (changes requested):

docker/docker_nightly.py
agunapal: Tried it. It works! Updated the code.

msaroufim (approved):

PR #2592, opened 2023-09-13
A guide for loading models in TorchServe

agunapal:
## Description

This is a guide for loading models in TorchServe

Main motivation was from the note by Hamel Husain https://hamel.dev/notes/serving/torchserve/
Fixes #(issue)

## Type of change

N/A

## Checklist:

- [ ] Did you have fun?

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -7,6 +7,7 @@ TorchServe is a performant, flexible and easy to use tool for serving PyTorch ea
 * [Serving Quick Start](https://github.com/pytorch/serve/blob/master/README.md#serve-a-model) - Basic server usage tutorial
 * [Model Archive Quick Start](https://github.com/pytorch/serve/tree/master/model-archiver#creating-a-model-archive) - Tutorial that shows you how to package a model archive file.
 * [Installation](https://github.com/pytorch/serve/blob/master/README.md#install-torchserve) - Installation procedures
+* [Model loading](model_loading.md) - How to load a model in TorchServe?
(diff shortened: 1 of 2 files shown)


msaroufim (changes requested):
On whether handler has an initialize method i think it's more nuanced, someone could also be calling `super()`
The model type node makes a lot of sense
in the self contained package node i didn't really understand what you were trying to say because serialized file node intersects with torchscript and onnx. The righthand bottom node should be is your model large and do you care about model initialization time

agunapal:
@msaroufim Thanks. Agree with both. Changed both

msaroufim:
top part looks much clearer

bottom still unclear you ask two questions in the node so if i say yes its unclear which part i should go to


Screenshot 2023-09-13 at 4 18 32 PM

Also in the bottom row theres rows going both ways with no labels

also there's too much text in the large model thing

agunapal:
> top part looks much clearer
> 
> The first node in the photo i think deciision should be do you want a self contained artifact or do you want fast model loads
> 
> Screenshot 2023-09-13 at 4 18 32 PM
> Also in the bottom row theres arrows going both ways with no labels
> 
> also there's too much text in the large model thing

Got it. Let me think about how to address it

agunapal:
@msaroufim Changed the bottom part a bit. If the model is large, we want users to not package the model at all. So, I think that should be the starting point

msaroufim (approved):

Issue #2223, opened 2023-04-06
Debugging extension for VSCode


rbavery:
### 🚀 The feature

Be able to set breakpoints and step through custom handlers after sending requests to Torchserve. something like VSCode Live Server or the Javascript Debugger for Torchserve that works with VSCode Remote Containers.

### Motivation, pitch

Debugging containers with prints is tedious: https://github.com/pytorch/serve/issues/711

### Alternatives

debugging with prints: https://github.com/pytorch/serve/issues/711

msaroufim:
That's an awesome idea TBH, i don't think anyone on the team has bandwidth for it but if you're interested in picking this up I'd be happy to actively advise

msaroufim:
Ok didn't manage to finish this today but here's at a high level how this might work

The idea is you
1. Package up your mar file locally
2. Create a docker container with all the dependencies you need
3. Set breakpoints in your handler
4. Send curl requests to torchserve and get the breakpoints triggered

And here's step by step how one might set this up

1. `mkdir project`
5. Setup a dockerfile with torchserve


Dockerfile
```docker
FROM pytorch/torchserve:latest
USER root

ENV JAVA_HOME /usr/lib/jvm/java-17-openjdk-amd64
RUN pip install ptvsd

WORKDIR /workspace
COPY . /workspace
```

6. `touch .devcontainer/.devcontainer.json`

```json
{
"name": "Torchserve Debugging",
"dockerFile": "../Dockerfile",
"settings": {
  "terminal.integrated.shell.linux": "/bin/bash"
},
"extensions": ["ms-python.python"],
"forwardPorts": [8080, 8081, 6789],
"runArgs": ["--user", "root"]

}


```

7. In your `handler.py`, add at the very top

```python
import ptvsd

ptvsd.enable_attach(address=('0.0.0.0'con89), redirect_output=True)
ptvsd.wait_for_attach()
```

8. Install the VS code Dev container extension
9. Install the VS code python extension
10. Setup a `launch.json`

```json
{
  "version": "0.2.0",
  "configurations": [
    

      {
          "name": "Python: Attach",
          "type": "python",
          "request": "attach",
          "connect": {
              "host": "localhost",
              "port": 6789
          },
          "pathMappings": [
              {
                  "localRoot": "${workspaceFolder}",
                  "remoteRoot": "/workspace"
              }
          ]
      }
  ]
}
```
11. Launch a new Dev container
[...]

PR #2605, opened 2023-09-21
Example showing how to run TorchServe's backend code only

agunapal:
## Description

Example showing how to run TorchServe's backend code only
- This can be used with a debugger to find any issues in your handler code.

Fixes #(issue)

## Type of change

```
python debugging_backend/test_handler.py --batch_size 2
Torch TensorRT not enabled
DEBUG:ts.torch_handler.base_handler:Model file /home/ubuntu/serve/examples/image_classifier/resnet_18/resnet-18.pt loaded successfully
INFO:__main__:Result is [{'tabby': 0.4096629023551941, 'tiger_cat': 0.34670525789260864, 'Egyptian_cat': 0.13002872467041016, 'lynx': 0.02391958236694336, 'bucket': 0.011532173492014408}, {'tabby': 0.4096629023551941, 'tiger_cat': 0.34670525789260864, 'Egyptian_cat': 0.13002872467041016, 'lynx': 0.02391958236694336, 'bucket': 0.011532173492014408}]

```

## Checklist:

- [ ] Did you have fun?


examples/image_classifier/resnet_18/debugging_backend/test_handler.py
```diff
@@ -0,0 +1,66 @@
...
+import torch
+
+from ts.torch_handler.image_classifier import ImageClassifier
+from ts.torch_handler.unit_tests.test_utils.mock_context import MockContext
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.DEBUG)
+
+CURR_FILE_PATH = Path(__file__).parent.absolute()
+REPO_ROOT_DIR = CURR_FILE_PATH.parent.parent.parent.parent
```
mreso, line 15: nit: You can also just use CURR_FILE_PATH.parents[3] here

ts/torch_handler/unit_tests/test_utils/mock_context.py
```diff
@@ -2,11 +2,13 @@
 Mocks for adding model context without loading all of Torchserve
 """
 
+import os
 import uuid
 
 import torch
 
 from ts.metrics.metrics_store import MetricsStore
+from ts.utils.util import get_yaml_config
```
mreso, line 11: Are these new imports necessary?

examples/image_classifier/resnet_18/debugging_backend/test_handler.py
```diff
@@ -0,0 +1,66 @@
...
+    read_data = f.read()
+    data = []
+    for i in range(batch_size):
+        tmp = {}
+        tmp["data"] = read_data
+        data.append(tmp)
+    return data
+
+
+def main(batch_size):
```
mreso, line 35: Can you convert this into a unit test runnable with pytest (we can still keep the main if people come along and run it separately) and encourage people in the doc to write a unit test for their handler?

mreso (approved):
Would love if we could convert this into a unit test example for handler (see comments). Otherwise LGTM

ts/torch_handler/unit_tests/test_utils/mock_context.py
agunapal: If user wants to pass a yaml file, it is.. Its a bug

examples/image_classifier/resnet_18/debugging_backend/test_handler.py
agunapal: How about if I make an additional test in the  pytest folder and link as that an example to encourage people to write tests. I want to keep this separate for debugging. As someone who is not the most comfortable with all pytest syntax, having this example in pytest fodler, it would probably make me less likely to use it to debug.

Actually, I guess I can create a separate PR for the pytest example. The goal of that PR would be encourage everyone to write a test when they push an example. This test would also need creation of mar file, validation of output etc. What do you think?

mreso: @agunapal Actually, I did not mean to move it into the unit test folder. What I mean is that we rename main into test_resnet_something so that we can also run it with pytest __as well__. Then you can show users: "Here try to run this with pytest test_handler.py and there goes your unit test." I would argue that unit tests are a great way to debug your handler. Think about an assert in pytest. When you assert in python you only get: "This test failed". In pytest you get: "This test failed because the objects differ in this specific way" which is very valuable I think. And because we have a single test in a not unit test folder the syntax is minimal.

examples/image_classifier/resnet_18/debugging_backend/test_handler.py
agunapal: Done

examples/image_classifier/resnet_18/debugging_backend/test_handler.py
agunapal: @mreso I get your point and I will make a separate PR for pytest. I tried running this as a pytest, it expects me to a fixture for batch_size.  
I am specifically trying to address something like this https://github.com/pytorch/serve/issues/2223

I wanted this example to be a low barrier for entry kind of example. I also included a screenshot of the debugger. I am hoping TorchServe customers use this with a debugger

mreso: @agunapal just give batch_size a default value:
```
def test_main(batch_size=1):
```
Then is works in debugger and pytest without any scary syntax.

agunapal: Thanks! That works. Added pytest info in the README as well

PR #2632, opened 2023-09-29
Use process build for GPU command

msaroufim:
Solves https://github.com/pytorch/serve/security/code-scanning/4

### DIFF
--- a/frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerThread.java
+++ b/frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerThread.java
@@ -124,7 +124,7 @@ public class WorkerThread implements Runnable {
                         new ProcessBuilder(
                                 "nvidia-smi",
                                 "-i",
-                                String.valueOf(validatedGpuId),
+                                String.valueOf(gpuId),
                                 "--query-gpu=utilization.gpu,utilization.memory,memory.used",
                                 "--format=csv");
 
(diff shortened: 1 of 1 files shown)


frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerThread.java
```diff
@@ -120,12 +120,25 @@ public String getGpuUsage() {
...
-                                        "nvidia-smi -i "
-                                                + gpuId
-                                                + " --query-gpu=utilization.gpu,utilization.memory,memory.used --format=csv");
+                // Ensure gpuId is a valid number
+                int validatedGpuId;
+                try {
+                    validatedGpuId = Integer.parseInt(gpuId);
+                } catch (NumberFormatException e) {
+                    throw new IllegalArgumentException("Invalid GPU ID");
+                }
```
lxning, line 129: GPU id is already validated during assignment stage.
- https://github.com/pytorch/serve/blob/master/frontend/archive/src/main/java/org/pytorch/serve/archive/model/ModelConfig.java#L236
- https://github.com/pytorch/serve/blob/master/frontend/server/src/main/java/org/pytorch/serve/wlm/WorkLoadManager.java#L210

msaroufim: reverted gpu id validation but kep the process builder for gpu

namannandan (approved):

PR #2676, opened 2023-10-04
Cleaning up previous runs from GPU runner

agunapal:
## Description

This PR cleans up the workspace directory fully from the CI GPU runners.

Fixes #(issue)

## Type of change

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

## Checklist:

- [ ] Did you have fun?

### DIFF
--- a/.github/workflows/ci_gpu.yml
+++ b/.github/workflows/ci_gpu.yml
@@ -20,7 +20,6 @@ jobs:
       - name: Clean up previous run
         run: |
           echo "Cleaning up previous run"
-          cd $RUNNER_WORKSPACE
           ls -la ./
           sudo rm -rf ./* || true
           sudo rm -rf ./.??* || true
--- a/.github/workflows/ci_gpu.yml
+++ b/.github/workflows/ci_gpu.yml
@@ -20,6 +20,7 @@ jobs:
       - name: Clean up previous run
         run: |
           echo "Cleaning up previous run"
+          cd $RUNNER_WORKSPACE
           ls -la ./
           sudo rm -rf ./* || true
           sudo rm -rf ./.??* || true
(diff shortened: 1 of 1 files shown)


msaroufim:
Let's wait on ci, might also need to skip the streaming handler tests to actually merge this

I would have expected `cd $RUNNER_WORKSPACE` to stick around

.github/workflows/ci_gpu.yml
```diff
@@ -20,11 +20,10 @@ jobs:
...
           echo "Cleaning up previous run"
-          cd $RUNNER_WORKSPACE
-          pwd
-          cd ..
-          pwd
-          rm -rf _tool
+          ls -la ./
+          sudo rm -rf ./* || true
+          sudo rm -rf ./.??* || true
+          ls -la ./
```
lxning, line 26: we need specify the path and then delete

agunapal: This doesn't work https://github.com/pytorch/serve/actions/runs/6409297939

Also, we have been doing the same in Regression GPU runners and it works     

https://github.com/pytorch/serve/blob/master/.github/workflows/regression_tests_gpu.yml#L21

msaroufim: Could you maybe add a comment to more explicitly say what this is trying to achieve?

The benefit of not having `*` in remove commands is we can be more explicit about what's being deleted

msaroufim:
Just so we can move on 

```bash
# deletes all files in the current directory
sudo rm -rf ./* || true
# deletes all hidden files in a directory
sudo rm -rf ./.??* || true
``

msaroufim (approved):

PR #2701, opened 2023-10-11
Split official release workflow for binaries and docker

agunapal:
## Description

This PR splits the official release workflow into 2 parts
  - The original workflow will only publish binaries
  - Added a second workflow for publishing docker images
    - This is using the CI_GPU runner as self hosted runners have more disk space
    -  This will also ensure that, we can publish docker images once we confirm that PyPI indices are updated with the latest version of TorchServe

Also changed the `docker push` commands to use `os.system` calls as we want this to execute first and be blocking.

Fixes #(issue)

## Type of change

[Test run of Docker workflow](https://github.com/pytorch/serve/actions/runs/6486104508/job/17613580328)

## Checklist:

- [ ] Did you have fun?

### DIFF
--- a/.github/workflows/official_release_docker.yml
+++ b/.github/workflows/official_release_docker.yml
@@ -1,5 +1,5 @@
 #name: Make an official Release of Docker Images
-name: Testing the official Release of Docker Images
+name: Testing official Release of Docker Images
 
 #on: workflow_dispatch
 on: push
--- a/.github/workflows/official_release_docker.yml
+++ b/.github/workflows/official_release_docker.yml
@@ -1,5 +1,5 @@
 #name: Make an official Release of Docker Images
-name: Testing official Release of Docker Images
+name: Testing the official Release of Docker Images
 
 #on: workflow_dispatch
 on: push
--- a/.github/workflows/official_release_docker.yml
+++ b/.github/workflows/official_release_docker.yml
@@ -1,6 +1,16 @@
 name: Make an official Release of Docker Images
 
-on: workflow_dispatch
+on:
+  workflow_dispatch:
+    inputs:
+          upload_docker:
+            description: 'Upload TorchServe Docker Images?'
+            required: true
+            default: 'no'
+          upload_kfs:
+            description: 'Upload TorchServe-kfs Docker Images?'
+            required: true
+            default: 'no'
 
 jobs:
   official-release-docker:
@@ -25,10 +35,12 @@ jobs:
           DOCKER_PASSWORD: ${{secrets.DOCKER_PASSWORD}}
         run: docker login --username pytorchbot --password "$DOCKER_PASSWORD"
       - name: Build & Upload pytorch/torchserve Docker images
+        if: github.event.inputs.upload_docker == 'yes'
         run: |
           cd docker
           python build_upload_release.py --cleanup
       - name: Build & Upload pytorch/torchserve-kfs Docker images
+        if: github.event.inputs.upload_kfs == 'yes'
         run: |
           cd kubernetes/kserve
           python build_upload_release.py --cleanup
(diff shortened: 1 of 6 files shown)


kubernetes/kserve/docker_nightly.py
```diff
@@ -58,9 +58,9 @@
     )
 
     # Push images with latest tag
-    try_and_handle(f"docker push {organization}/{project}:latest-cpu", dry_run)
-    try_and_handle(f"docker push {organization}/{project}:latest-gpu", dry_run)
+    os.system(f"docker push {organization}/{project}:latest-cpu")
```
msaroufim, line 61: os.system won't error handle so if this line fails it will just keep going

msaroufim (changes requested):

kubernetes/kserve/docker_nightly.py
agunapal: Reading more about subprocess, it will execute in parallel to the main process.
Also, if you check the logs here , you will see that it executed the commands after delete was done https://github.com/pytorch/serve/actions/runs/6486104508/job/17613580328

So, not sure whats the right way to handle this.

.github/workflows/official_release_docker.yml
```diff
@@ -0,0 +1,34 @@
...
+        uses: actions/checkout@v3
+      - name: Login to Docker
+        env:
+          DOCKER_PASSWORD: ${{secrets.DOCKER_PASSWORD}}
+        run: docker login --username pytorchbot --password "$DOCKER_PASSWORD"
+      - name: Build & Upload pytorch/torchserve Docker images
+        run: |
+          cd docker
+          python build_upload_release.py --cleanup
+      - name: Build & Upload pytorch/torchserve-kfs Docker images
```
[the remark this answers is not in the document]
agunapal: I didn't follow. What are we gating on? The process would be for us to trigger the binary workflow as before, which runs pending approval. Once its complete, do manual tests. Once thats complete, trigger the docker workflow, which again runs after approval.

kubernetes/kserve/docker_nightly.py
msaroufim: subprocess.call and subprocess.run should be synchronous only subprocess.Popen is async and you make it sync by calling an await

agunapal: Unless, we actually push, I dont have a way to prove that this doesn't work.

But this is what i see in the logs.

Screenshot 2023-10-11 at 11 36 47 AM

agunapal: I can revert the nightly to use try_and_handle as before and see what happens. But for the official release, we can use the system command. Is that fine?

.github/workflows/official_release_docker.yml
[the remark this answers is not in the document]
msaroufim: I'm suggesting not to upload kserve images by default

kubernetes/kserve/docker_nightly.py
msaroufim: is this the dry run?

.github/workflows/official_release_docker.yml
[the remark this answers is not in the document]
agunapal: Oh.I get it. We would need another input to skip uploading docker images when we want to actually push kserve images. I will add it, but not sure if I will be able to test it before the release.

kubernetes/kserve/docker_nightly.py
agunapal: Yes

agunapal: Reverted the nightly as before. We can see if it works tonight. So far, images were not getting deleted because the prune command didnt have the `-f` option

msaroufim (approved):
lgtm

agunapal:
@msaroufim `subprocess.run` worked fine in last night's run. It was blocking. So, I guess only the print statements are probably not executed in sequence. Will update the official workflow to do the same

PR #2734, opened 2023-10-23
KServe CPU Nightly Test

agunapal:
## Description

This PR introduces CI for TorchServe-kfs nightly images
- Tests V1 and V2 Inference protocol
- Works with Mnist
- CPU only

Fixes #(issue)

## Type of change

[Passing run](https://github.com/pytorch/serve/actions/runs/6643934661/job/18052010275)

## Checklist:

- [ ] Did you have fun?


kubernetes/kserve/tests/scripts/test_mnist.sh
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env bash
```
msaroufim, line 1: a bit conflicted about this, ideally all our tests should run from regression or sanity tests that way people have a clear way of reproducing failures both locally and remotely 

Do we need the extra json and yaml files? Can't we leverage more of the existing kserve tests?

kubernetes/kserve/tests/scripts/test_mnist.sh
```diff
@@ -0,0 +1,119 @@
...
+echo "Waiting 5s for kserve pod to come up ..."
+sleep 5
+
+echo "Deploy the cluster"
+
+cd $GITHUB_WORKSPACE
+kubectl apply -f kubernetes/kserve/tests/configs/mnist_v2_cpu.yaml
+
+echo "Waiting 120s for pod to come up..."
+sleep 120
```
msaroufim, line 27: I don't like these sleep staetments, can you poll the pod status with kubectl or minikube instead?

kubernetes/kserve/tests/scripts/test_mnist.sh
```diff
@@ -0,0 +1,119 @@
...
+
+echo "Make inference request"
+
+PREDICTION=$(curl -H "Content-Type: application/json" -H "Host: ${SERVICE_HOSTNAME}" http://${INGRESS_HOST}:${INGRESS_PORT}/v2/models/${MODEL_NAME}/infer -d @./kubernetes/kserve/tests/data/mnist_v2.json)
+
+EXPECTED='{"model_name":"mnist","model_version":null,"id":"d3b15cad-50a2-4eaf-80ce-8b0a428bd298","parameters":null,"outputs":[{"name":"input-0","shape":[1],"datatype":"INT64","parameters":null,"data":[1]}]}'
+echo "Creating a Service"
+
+if [ "${PREDICTION}" = "${EXPECTED}" ]; then
+     echo "✓ Prediction: ${PREDICTION} (Expected ${EXPECTED})"
```
msaroufim, line 51: these predictions can get pretty long I presume? also best to be explict and say success so its easier to grep in logs later

kubernetes/kserve/tests/scripts/test_mnist.sh
```diff
@@ -0,0 +1,119 @@
...
+echo "Wait for ports to be in forwarding"
+sleep 10
+
+echo "Make inference request"
+
+PREDICTION=$(curl -H "Content-Type: application/json" -H "Host: ${SERVICE_HOSTNAME}" http://${INGRESS_HOST}:${INGRESS_PORT}/v1/models/${MODEL_NAME}:predict -d @./kubernetes/kserve/tests/data/mnist_v1.json)
+
+EXPECTED='{"predictions":[2]}'
+echo "Creating a Service"
+
```
msaroufim, line 107: this is duplicated right? can you encapsulate this into functions? If writing bash is too annoying would reocmmend rewriting this as a python script

msaroufim (changes requested):

.github/workflows/kserve_ci_cpu.yml
```diff
@@ -0,0 +1,40 @@
+name: KServe CI CPU
+
+on:
+  workflow_dispatch:
+  # runs everyday  at 5:15am
+  schedule:
+    - cron:  '15 5 * * *'
```
msaroufim, line 7: instead of running this nightly you can run it only when changes to kubernetes folder are made

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths

kubernetes/kserve/tests/scripts/test_mnist.sh
agunapal: Re-used the json files we have. We don't have the yaml files.

kubernetes/kserve/tests/scripts/test_mnist.sh
agunapal: Done. Refactored the code to be modular

kubernetes/kserve/tests/scripts/test_mnist.sh
agunapal: Done

kubernetes/kserve/tests/scripts/test_mnist.sh
agunapal: Done. Refactored the code

.github/workflows/kserve_ci_cpu.yml
agunapal: Since we push nightly images, we should be running this nightly. It takes < 5 minutes to run

msaroufim (approved):

PR #2779, opened 2023-11-09
deepspeed mii fastgen example

lxning:
## Description

Fixes #(issue)
#2780 

## Type of change

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

## Checklist:

- [ ] Did you have fun?


examples/large_models/deepspeed_mii/LLM/model-config.yaml
```diff
@@ -0,0 +1,20 @@
+# TorchServe frontend parameters
+minWorkers: 1
+maxWorkers: 1
+maxBatchDelay: 100
+responseTimeout: 1200
+parallelType: "tp"
+deviceType: "gpu"
+# example of user specified GPU deviceIds
+deviceIds: [0,1,2,3] # seting CUDA_VISIBLE_DEVICES
```
mreso, line 9: Is this working if we have 8 gpus in a host?

ts/protocol/otf_message_handler.py
```diff
@@ -64,6 +64,9 @@ def create_predict_response(
     :param code:
     :return:
     """
+    if str(os.getenv("LOCAL_RANK", 0)) != "0":
```
mreso, line 67: This low level method should not be concerned about checking the LOCAL_RANK env environment. We should just not call it when its not "0".

mreso (changes requested):
Overall is LGTM but I would like to move the check on LOCAL_RANK up the chain.

examples/large_models/deepspeed_mii/LLM/model-config.yaml
lxning: yes, TorchServe support customer specified gpu allocation.

ts/protocol/otf_message_handler.py
mreso: I see the intention here. I still think this is not the preferred place to put this check. The function creating a response from some arguments should not need to know about the concept of ranks. In fact we already have a central point to check for the rank in this PR which is [here](https://github.com/pytorch/serve/blob/e7dbf4562f9aac8d6cc07aa6d36e4c017e8c919c/ts/model_service_worker.py#L183).
The only point where this is missing is [here](https://github.com/pytorch/serve/blob/e7dbf4562f9aac8d6cc07aa6d36e4c017e8c919c/ts/protocol/otf_message_handler.py#L362). Looking at this I think we should move send_intermediate_predict_response out of ts.protocol.otf_message_handler (Users should not need deal with the modules containing our comms between fe/be. What if we drop otf protocol?). My first instinct is to move it under ts.handler_utils. What do you think?

mreso: AFAIK we do not guarantee bc as of now. Its probably time to plan out our bc strategy for future release as we're running out or 0.X version numbers. Anyways, if your concern is very high you can leave a wrapper in otf_message_handler that fires a deprecation warning and then calls the one in handler_utils until next release. cx will only get worse if we let the code rot.

mreso (changes requested):
Could you check the deprecation decorator? I've added some checks to your test that should pass but every time foo is called the warning is emitted again.

lxning:
> Could you check the deprecation decorator? I've added some checks to your test that should pass but every time foo is called the warning is emitted again.

A deprecation warning message will be triggered once the deprecated function is called. For example, Two warning messages will be emitted if func foo is called twice in the test. This is what I expected. I updated the assertion:

Replace
```
assert len(record) == 0
```
with
```
assert len(record) == 2
```

mreso:
> > Could you check the deprecation decorator? I've added some checks to your test that should pass but every time foo is called the warning is emitted again.
> 
> A deprecation warning message will be triggered once the deprecated function is called. For example, Two warning messages will be emitted if func foo is called twice in the test. This is what I expected. I updated the assertion:
> 
> Replace
> 
> ```
> assert len(record) == 0
> ```
> 
> with
> 
> ```
> assert len(record) == 2
> ```

@lxning If that is the expected outcome, can we remove this?
https://github.com/pytorch/serve/blob/839782a1aad1dd16380c75a8b4d911167c161088/ts/utils/util.py#L14

lxning:
> > > Could you check the deprecation decorator? I've added some checks to your test that should pass but every time foo is called the warning is emitted again.
> > 
> > 
> > A deprecation warning message will be triggered once the deprecated function is called. For example, Two warning messages will be emitted if func foo is called twice in the test. This is what I expected. I updated the assertion:
> > Replace
> > ```
> > assert len(record) == 0
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > with
> > ```
> > assert len(record) == 2
> > ```
> 
> @lxning If that is the expected outcome, can we remove this?
> 
> https://github.com/pytorch/serve/blob/839782a1aad1dd16380c75a8b4d911167c161088/ts/utils/util.py#L14

Good catch. Remove this unused func. Thanks!

mreso (approved):
LGTM

PR #2786, opened 2023-11-12
Clarify that "Not Accepted" is a valid classification

### LINKED ISSUE #1848: Confusing Sequence Classification Example for HuggingFace
### 📚 The doc issue

In the sequence classification example in examples/Huggingface_Transformers, the model results in a binary classification 'Accepted' or 'Not Accepted' as defined in `index_to_name.json`. The example query in `sample_text_captum_input.txt` will results in a 'Not Accepted' classification.. If I am new to torchserve and working through the example for the first time, this looks like a error - as in "the input that you gave was not accepted by the torchserve API" when really it is the correct classification from the model

### Suggest a potential alternative/fix

I would suggest either 

1. changing `index_to_name.json` to result in a different binary classification besides 'Accepted'/'Not Accepted' 
2. changing the documentation to make it explicitly clear that the binary classification will result in an 'Accepted'/'Not Accepted'  classification, and that 'Not accepted' is not some internal torchserve error

nathanweeks:
## Description

In the examples/Huggingface_Transformers Sequence Classification example, clarify that "Not Accepted" is a valid classification response, rather than a TorchServe error.

Fixes #1848 

## Type of change

### DIFF
--- a/examples/Huggingface_Transformers/README.md
+++ b/examples/Huggingface_Transformers/README.md
@@ -106,6 +106,9 @@ torchserve --start --model-store model_store --models my_tc=BERTSeqClassificatio
 ### Run an inference
 
 To run an inference: `curl -X POST http://127.0.0.1:8080/predictions/my_tc -T Seq_classification_artifacts/sample_text_captum_input.txt`
+
+The response should be a "Not Accepted" classification.
+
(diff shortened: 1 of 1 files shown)


agunapal (approved):
LGTM

PR #2804, opened 2023-11-22
Use `inference_mode` instead of `no_grad`

bryant1410:
Use `inference_mode` instead of `no_grad` for `BaseHandler`. It should bring a (small?) performance boost.

Though, I have 2 questions:
* What's the min supported PyTorch version? Because `inference_mode` is supported since PT 1.9.
* Should other handlers (including example ones) be changed as well?

### DIFF
--- a/ts/torch_handler/base_handler.py
+++ b/ts/torch_handler/base_handler.py
@@ -313,7 +313,7 @@ class BaseHandler(abc.ABC):
         Returns:
             Torch Tensor : The Predicted Torch Tensor is returned in this function.
         """
-        with torch.no_grad():
+        with torch.inference_mode():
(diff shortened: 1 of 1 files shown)


msaroufim:
A bit conflicted about this one since inference mode has a bad interaction with torch.compile

bryant1410:
> A bit conflicted about this one since inference mode has a bad interaction with torch.compile

Yeah, I've seen that. I think it's only for 2.0 but not 2.1: https://github.com/pytorch/pytorch/issues/103132. It works fine for me for 2.1 (but not with 2.0).

We could only use it for PT > 2.0, else use `no_grad`. What do you think?

lxning (approved):

Issue #1532, opened 2022-03-25
permission error while trying to remove previous artifact files in `/tmp/`


replon:
https://github.com/pytorch/serve/blob/6a191f822f2e759c2aad07b8f9eacbedcb039827/model-archiver/model_archiver/model_packaging_utils.py#L137

In a multi-user system, this line can raise Permission Error because of the same file name in the shared `/tmp/` directory which maybe created by another user. Maybe we can give more unique name when we store it using timestamp or something.

mahadev-k:
Hey, I would like to work on this issue. Would you like to assign it to me ?.
I believe the model_path to be corrected with the addition of date and time to make it unique. Let me know your thoughts. Thanks!!

agunapal:
@mahadev-k Please feel free to try it out and send a PR

mahadev-k:
Thanks !!. I have done some changes. I would like to test it out. How to build this application? I tried couple of things but didn't work. Also, currently I am relying on unit test to test this out. If I want to test it in a real way, how can I do that.

PR #2868, opened 2023-12-23
Create temporary directory using uuid

lamttic:
## Description

I suggest creating a temporary directory using a unique hash(UUID) for robustness.
There was an issue where the build didn't complete successfully when storing artifacts in the `/tmp/{model_name}` path and building the archive.
So, upon reviewing the code, I found that the folder was being deleted by the model_packaging_utils. To address this issue, I'm looking to uniquely designate the temporary folder path for the archive.
Furthermore, As mentioned previously in [this issue](https://github.com/pytorch/serve/issues/1532), accessing the temporary path simultaneously by multiple accounts can lead to permission issues. I also hope that it will be resolved along with this PR.

### DIFF
--- a/model-archiver/model_archiver/model_packaging_utils.py
+++ b/model-archiver/model_archiver/model_packaging_utils.py
@@ -9,6 +9,7 @@ import re
 import shutil
 import tarfile
 import tempfile
+import uuid
 import zipfile
 from io import BytesIO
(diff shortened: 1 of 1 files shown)


agunapal:
@lamttic Could you please attach the logs of an example directory that is created.

PR #2887, opened 2024-01-03
Fix Inf2 package dependency version conflicts

namannandan:
## Description
Recent Inf2 benchmark runs have been failing due to conflicting package versions as follows:
```
ERROR: Cannot install -r requirements/torch_neuronx_linux.txt (line 4), -r requirements/torch_neuronx_linux.txt (line 5), -r requirements/torch_neuronx_linux.txt (line 6), -r requirements/torch_neuronx_linux.txt (line 7) and torch==1.13.1+cpu because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested torch==1.13.1+cpu
    torchvision 0.14.1+cpu depends on torch==1.13.1
    torchtext 0.14.1 depends on torch==1.13.1
    torchaudio 0.13.1+cpu depends on torch==1.13.1
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
    torchdata 0.7.0 depends on torch==2.1.0
```
```
ERROR: Cannot install -r requirements/neuronx.txt (line 4) and protobuf==4.25.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested protobuf==4.25.1
    grpcio-tools 1.48.2 depends on protobuf<4.0dev and >=3.12.0
```

## Type of change

- [ ] Passing Inf2 benchmark run
[Inf2 benchmark](https://github.com/pytorch/serve/actions/runs/7391694705/job/20110116669)

### DIFF
--- a/requirements/neuronx.txt
+++ b/requirements/neuronx.txt
@@ -1,7 +1,5 @@
 --extra-index-url https://pip.repos.neuron.amazonaws.com
-numpy==1.22.4
-protobuf==4.25.1
-grpcio-tools==1.48.2
 neuronx-cc
 torch-neuronx
 transformers-neuronx
+grpcio-tools
--- a/requirements/neuronx.txt
+++ b/requirements/neuronx.txt
@@ -2,4 +2,3 @@
 neuronx-cc
 torch-neuronx
 transformers-neuronx
-grpcio-tools
--- a/.github/workflows/benchmark_nightly.yml
+++ b/.github/workflows/benchmark_nightly.yml
@@ -1,14 +1,16 @@
 name: Benchmark torchserve nightly
 
 on:
-  push
+  # run every day at 2:15am
+  schedule:
+    - cron:  '15 02 * * *'
 
 jobs:
   nightly:
     strategy:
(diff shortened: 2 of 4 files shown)


requirements/neuronx.txt
sachanub, line 1: What is the motivation to remove the dependencies altogether rather than handle the version dependency conflict? Also, the conflict seems to exist for `protobuf` and `grpcio-tools`. Why are we removing `numpy` as well?

namannandan: We can remove `numpy` and `protobuf` because `neuronx-cc` will setup the required supported version of `numpy` and `torch-neuronx` will do the same for `protobuf`. This way, we don't have to hardcode individual package versions. Also, `grpcio-tools` is not a dependency for `neuronx-cc`, `torch-neuronx` or `transformers-neuronx` so I've removed that as well.

sachanub: LGTM.

sachanub (approved):

PR #2928, opened 2024-02-05
Fix set model state if runtime is null

mreso:
## Description
This PR fixes an issue when a model has no runtime set in the snapshot and we load with a newer torchserve version which expects the field.

Fixes #(issue)

## Type of change

- 

## Checklist:

- [X] Did you have fun?

### DIFF
--- a/frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
+++ b/frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
@@ -1,5 +1,6 @@
 package org.pytorch.serve.wlm;
 
+import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import java.io.File;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


agunapal (commented):
How do we make sure that someone running 0.9.0 TS with snapshot upgrading to the latest will not run into an issue.
Can we test this manually and add logs here?

mreso:
@ankithagunapal the new test test_empty_runtime covers that scenario by removing the runtime entry from the snapshot, but sure lets do this once manually.

mreso:
> How do we make sure that someone running 0.9.0 TS with snapshot upgrading to the latest will not run into an issue. Can we test this manually and add logs here?

Here is the log of the test:
https://gist.github.com/mreso/f3ba1b41489e8bb76856b295ecacfadf

agunapal (approved):
LGTM

Issue #2555, opened 2023-08-30
Torchserve M1 support


msaroufim:
### 🚀 The feature

Officially support torchserve on M1-M2 mac hardware. A bunch of people have asked for this, if you're interested please either upvote or volunteer to do this work, would be happy to review

### Motivation, pitch

I've seen a bunch of issues asking for this but we haven't managed to prioritize it yet so thought I'd highlight what needs to be done to support this

- [ ] Ensure `ts_scripts/install_dependencies.py` installs the right binaries for M1
- [ ] Update our docker build scripts to also work with M1 https://github.com/pytorch/serve/issues/1363#issuecomment-1112769676
- [ ] In our base handler check if mps is available and if it is push the model and its inputs to an mps device https://github.com/pytorch/serve/blob/master/ts/torch_handler/base_handler.py
- [ ] There will be many cases where mps might have some unsupported feature in which case it should be possible to also hardwire in config cpu fallback by putting in a flag in a model config like this one as an example https://github.com/pytorch/serve/blob/master/examples/large_models/inferentia2/model-config.yaml

cc @agunapal @mreso @malfet

Talbot3:
did it ok?i am  waiting ....

udaij12:
Hi @Talbot3 What issues are you still facing?

VladimirKadlec:
Hi, just FYI, `./build_image.sh` builds a working docker image just fine on Macbook Air M2 running Asahi Linux (Ubuntu 23.10)

Talbot3:
> Hi @Talbot3 What issues are you still facing?

https://github.com/pytorch/serve/issues/1363
I just want check pytorch container on my device (mac m1) can run

PR #2934, opened 2024-02-08
Adding M1 to Ci workflows

udaij12:
## Description
Adding M1 compatibility to CI jobs.
Tested: 
[Regression test](https://github.com/pytorch/serve/actions/runs/7836636911)
[Ci test](https://github.com/pytorch/serve/actions/runs/7836636910)

Issue #2555

## TODO
Change python back 3.8 when this [issue](https://github.com/actions/setup-python/issues/808) is resolved.


.github/workflows/ci_cpu.yml
```diff
@@ -21,9 +21,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04, macOS-latest]
+        os: [ubuntu-20.04, macOS-latest, macos-14]
     steps:
-      - name: Setup Python 3.8
+      - name: Setup Python for M1
```
msaroufim, line 26: Why python 3.10? It'll make releasing the binaries a bit annoying if each platform requires a different python version

msaroufim (changes requested):
So assuming CI passes we should be good to go but not sure I quite follow why we use python 3.10 vs 3.8 just for M1

.github/workflows/ci_cpu.yml
udaij12: Github `macos-14` does not work with python setup less than 3.10 https://github.com/actions/setup-python/issues/808

msaroufim: Reading through that issue it seems like the issue was fixed, mind retriggering with 3.8

msaroufim (approved):
Ok so assuming the macos-14 jobs pass we should be good to merge this, they are likely  not a required check so watch out and don't add this to merge queue before confirming

I'd also add a TODO referencing https://github.com/actions/setup-python/issues/808 in both CI jobs so we can remember to remove the python 3.10 stuff when ready

PR #3003, opened 2024-03-06
TorchServe quickstart chatbot example

agunapal:
## Description

This PR enables a new user of TorchServe to quickly launch a chatbot on Mac M1/M2 using TorchServe with 3 commands

```
# 1: Set HF Token as Env variable
export HUGGINGFACE_TOKEN=<Token> # get this from your HuggingFace account

# 2: Build TorchServe Image for Serving llama2-7b model with 4-bit quantization
./examples/llm/llama2/chat_app/docker/build_image.sh meta-llama/Llama-2-7b-chat-hf

# 3: Launch the streamlit app for server & client
docker run --rm -it --platform linux/amd64 -p 127.0.0.1:8080:8080 -p 127.0.0.1:8081:8081 -p 127.0.0.1:8082:8082 -p 127.0.0.1:8084:8084 -p 127.0.0.1:8085:8085 -v <model-store>:/home/model-server/model-store pytorch/torchserve:meta-llama---Llama-2-7b-chat-hf
```

Prerequsites:
1) HuggingFace token
2) Docker 

Fixes #(issue)

## Type of change

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

## Checklist:

- [ ] Did you have fun?


examples/LLM/llama2/chat_app/Readme.md
```diff
@@ -9,6 +9,33 @@ We are using [llama-cpp-python](https://github.com/abetlen/llama-cpp-python) in
 You can run this example on your laptop to understand how to use TorchServe
 
 
+## Quick Start Guide
```
msaroufim, line 12: We can be more ambitious and make this our new getting started

examples/LLM/llama2/chat_app/Readme.md
```diff
@@ -9,6 +9,33 @@ We are using [llama-cpp-python](https://github.com/abetlen/llama-cpp-python) in
...
+To get started with TorchServe, you need to run the following
+
+```
+# 1: Set HF Token as Env variable
+export HUGGINGFACE_TOKEN=<Token> # get this from your HuggingFace account
+
+# 2: Build TorchServe Image for Serving llama2-7b model with 4-bit quantization
+./examples/llm/llama2/chat_app/docker/build_image.sh meta-llama/Llama-2-7b-chat-hf
+
+# 3: Launch the streamlit app for server & client
```
msaroufim, line 23: I know it's not exactly what you might have in mind but I was thinking this would open a terminal based CLI

examples/LLM/llama2/chat_app/docker/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+RUN --mount=type=cache,id=apt-dev,target=/var/cache/apt \
+    apt-get update && \
+    apt-get install libopenmpi-dev git -y
+
+COPY $EXAMPLE_DIR/requirements.txt /home/model-server/chat_bot/requirements.txt
+RUN pip install -r /home/model-server/chat_bot/requirements.txt && huggingface-cli login --token $HUGGINGFACE_TOKEN
+RUN pip uninstall torchtext torchdata torch torchvision torchaudio -y
+RUN pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cpu --ignore-installed
+RUN pip uninstall torchserve torch-model-archiver torch-workflow-archiver -y
```
msaroufim, line 21: seems like a miss?

examples/LLM/llama2/chat_app/docker/Dockerfile
```diff
@@ -0,0 +1,30 @@
+ARG BASE_IMAGE=pytorch/torchserve:latest-gpu
+
+FROM $BASE_IMAGE as server
+ARG BASE_IMAGE
+ARG EXAMPLE_DIR
+ARG MODEL_NAME
+ARG HUGGINGFACE_TOKEN
+
+USER root
```
msaroufim, line 9: do you need root?

examples/LLM/llama2/chat_app/docker/torchserve_server_app.py
```diff
@@ -0,0 +1,193 @@
...
+
+MODEL_NAME = os.environ["MODEL_NAME"]
+MODEL = MODEL_NAME.split("---")[1]
+
+# App title
+st.set_page_config(page_title="TorchServe Server")
+
+
+def start_server():
+    os.system("torchserve --start --ts-config /home/model-server/config.properties")
```
msaroufim, line 16: This would show as success even if the the server failed to start, favor subprocess instead, check the return code and then query torchserve directly to see server health as opposed to using sleep

examples/LLM/llama2/chat_app/Readme.md
```diff
@@ -9,6 +9,33 @@ We are using [llama-cpp-python](https://github.com/abetlen/llama-cpp-python) in
...
+docker run --rm -it --platform linux/amd64 -p 127.0.0.1:8080:8080 -p 127.0.0.1:8081:8081 -p 127.0.0.1:8082:8082 -p 127.0.0.1:8084:8084 -p 127.0.0.1:8085:8085 -v <model-store>:/home/model-server/model-store pytorch/torchserve:meta-llama---Llama-2-7b-chat-hf
+```
+In step 3, `<model-store>` is a location where you want the model to be downloaded
+
+### What to expect
+This launches two streamlit apps
+1. TorchServe Server app to start/stop TorchServe, load model, scale up/down workers, configure dynamic batch_size ( Currently llama-cpp-python doesn't support batch_size > 1)
```
msaroufim, line 30: It's a bit painful to use llama-cpp here was hopign we could instead showcase an example with export or with mps in eager

msaroufim (changes requested):

examples/LLM/llama2/chat_app/Readme.md
agunapal: I was thinking about how to cover various scenarios..
So, my goal is to do a 3 part solution

1) chatbot quickstart with streamlit -> Because chatbots are popular
2) TS multi model app to show TS' full capability - Use this to create video series
3) quick start script for common use-cases with curl command -> This can be the getting started guide.

examples/LLM/llama2/chat_app/docker/Dockerfile
agunapal: You are right. This is not needed for this example. Will clean it

examples/LLM/llama2/chat_app/docker/torchserve_server_app.py
agunapal: Oh..good point..let me try. This command was returning immediately. I did try with ping but that was failing as the server was not up yet.

examples/LLM/llama2/chat_app/Readme.md
agunapal: I tried a few things
1) Use HF 7b models with quantization -> Only supported for CUDA
2) Use HF 7b models without quantization on CPU -> Extremely slow. No one would use this. 
3) Docker with MPS -> Seems like this is still not supported. Even pytorch supports only cpu in docker. https://github.com/pytorch/pytorch/issues/81224


So, currently this seems like the best solution. Seems like some people have tried mistral7b with llama-cpp-python..Its kind of mind blowing that most existing solutions are only targeted for the GPU rich.

examples/LLM/llama2/chat_app/docker/Dockerfile
agunapal: Done

examples/LLM/llama2/chat_app/docker/torchserve_server_app.py
agunapal: I changed the logic, but it still doesn't work as expected. There is a slight difference between when the command returns and when the server starts. Add a check with ping , but needs a sleep.

msaroufim (approved):

PR #3040, opened 2024-03-23
TorchServe quick start example

agunapal:
## Description

Fixes #(issue)

## Type of change

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

## Checklist:

- [ ] Did you have fun?


lxning:
This example contains Dockerfile and the config+handler for BERT, Resnet, and Desnet. It's not clear for cx to follow up this example since TS has dir for Dockerfile and model handlers for these models.

We need have a clear guidance and implementation to support model card.

examples/getting_started/index_mapping/berttg_index_to_name.json
```diff
@@ -0,0 +1,4 @@
+{
```
msaroufim, line 1: these labels have been confusing forever, can we do something else?

examples/getting_started/Transformer_handler_generalized.py
```diff
@@ -0,0 +1,538 @@
+import ast
```
msaroufim, line 1: is this copy pasted from the examples transformer handler? The duplication will hurt us would recommend having this in 1 place

examples/getting_started/Download_Transformer_models.py
```diff
@@ -0,0 +1,188 @@
+import argparse
```
msaroufim, line 1: same point on duplication here

examples/getting_started/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+    if echo "$MODEL_NAME" | grep -q "BERT"; then \
+        huggingface-cli login --token $HUGGINGFACE_TOKEN; \
+    fi
+
+COPY $EXAMPLE_DIR  /home/model-server/getting_started
+COPY $EXAMPLE_DIR/dockerd-entrypoint.sh /usr/local/bin/dockerd-entrypoint.sh
+COPY $EXAMPLE_DIR/config.properties /home/model-server/config.properties
+
+WORKDIR /home/model-server/getting_started
+RUN chmod +x /usr/local/bin/dockerd-entrypoint.sh \
```
msaroufim, line 30: cool!

msaroufim (changes requested):
personally I like the direction this is going in. Recommending docker early on to users feels like a natural thing since torchserve will just work on any device

I also like making the compiler flag an env variable will make trying things out quite simple

What I don't like is that bert examples still have these unsinspiring labels for "accepted" vs "not accepted" which is confusing because it might make people think it's a sytem level error instead of the expected output of the mode. I would also want to see the code duplication issues fixed

Otherwise once this is in I do expect this to be default way people first use torchserve, presumably you're doing llama7b and 70b in a future PR?

PR #3048, opened 2024-03-27
Adding mps support to base handler and regression test

udaij12:
## Description
Adding MPS as device type for Mac M1. 

When Model-config.yaml `deviceType: "gpu"` or deviceType is not specified then default to GPU: 
MPS enabled

When Model-config.yaml `deviceType: "cpu"`:
MPS not enabled 

## Test cases 

Added 4 test cases. 
Frontend:
1. test that number of gpu is greater then 0 when on Mac M1

Backend: 
1. Set deviceType to GPU and test that device == "mps"
3. Set deviceType to CPU and test that device != "mps" 
4. Dont set deviceType (remains at default) and test that device == "mps"


ts/torch_handler/base_handler.py
```diff
@@ -149,6 +149,8 @@ def initialize(self, context):
             self.device = torch.device(
                 self.map_location + ":" + str(properties.get("gpu_id"))
             )
+        elif hasattr(self, "model_yaml_config") and "mps" in self.model_yaml_config and self.model_yaml_config["mps"] == "enable":
```
agunapal, line 152: Can we add `mps` here https://github.com/pytorch/serve/blob/1a99de4675eaaad4ddf532e603ae2f175656462a/model-archiver/README.md?plain=1#L174

cc @lxning

agunapal: In terms of design, having one config for all `deviceType` probably makes sense?

lxning:
1. CX still can specify [deviceType](https://github.com/pytorch/serve/blob/master/frontend/archive/src/main/java/org/pytorch/serve/archive/model/ModelConfig.java#L433) (CPU or GPU) on Mac.
2. The PR should be able to support [getNumber of GPUs](https://github.com/pytorch/serve/blob/1a99de4675eaaad4ddf532e603ae2f175656462a/frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java#L829) on Mac.
3. The GPU device id is decided by frontend, which is as same as Linux.
4. Backend handler should be able to auto-detect Mac and apply MPS+deviceId. Essentially, there are no new config parameters in the model-config.yaml.

udaij12:
MPS is auto detected in the base handler and getAvailableGpu() now supports Mac M1 gpu cores.

agunapal (changes requested):
Please add the User experience for both the cases in the PR/ add this in a README and add a pytest for MPS on/off for a model

agunapal:
cc: @msaroufim

msaroufim (dismissed):

lxning (commented):
Can you add test cases?
1. define deviceType in model-config.yaml on Mac
1. cpu
2. gpu
* deviceId
2. define deviceType in model-config.yaml on gpu host

frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
```diff
@@ -835,6 +837,28 @@ private static int getAvailableGpu() {
                 for (String id : ids) {
                     gpuIds.add(Integer.parseInt(id));
                 }
+            } else if (System.getProperty("os.name").startsWith("Mac")) {
```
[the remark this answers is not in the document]
udaij12: Runtime.getRuntime().availableProcessors(); will return CPU cores not GPU cores

ts/torch_handler/base_handler.py
```diff
@@ -144,11 +144,18 @@ def initialize(self, context):
             self.model_yaml_config = context.model_yaml_config
 
         properties = context.system_properties
+        
         if torch.cuda.is_available() and properties.get("gpu_id") is not None:
             self.map_location = "cuda"
             self.device = torch.device(
                 self.map_location + ":" + str(properties.get("gpu_id"))
             )
+        elif torch.backends.mps.is_available() and properties.get("gpu_id") is not None:
```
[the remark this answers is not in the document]
udaij12: MPS is auto enabled if using Mac M1 and deviceType is GPU or undefined.

ts/torch_handler/base_handler.py
```diff
@@ -144,11 +144,18 @@ def initialize(self, context):
...
+            self.map_location = "mps"
+            self.device = torch.device(
+                self.map_location + ":" + str(properties.get("gpu_id"))
```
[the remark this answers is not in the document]
udaij12: changed

test/pytest/test_device_config.py
```diff
@@ -0,0 +1,181 @@
...
+    test_utils.reg_resp = test_utils.register_model_with_params(params)
+
+    yield model_name
+
+    test_utils.unregister_model(model_name)
+
+
+@pytest.mark.skipif(platform.machine() != "arm64", reason="Skip on Mac M1")
+@pytest.mark.parametrize("model_config_name", ["gpu"], indirect=True)
+def test_m1_device(model_name, model_config_name):
```
agunapal, line 153: How does this test mps?

The goal is to make sure mps is being used. 

nit: Please rename the tests, docs to apple_silicon . The same should work on m2 mac too?

agunapal (changes requested):

test/pytest/test_device_config.py
udaij12: Line 31 asserts for device type equal to mps. If this assert fails then the tests will have a 404 response code to describe model. 

As for m2 I am not sure as we have not tested that. None of the Github runners support m2 and have not tested locally on m2 machine to verify.

test/resources/model-config.yaml
```diff
@@ -0,0 +1 @@
+mps: "enable"
```
lxning, line 1: Why need this file and the parameter "mps". I assume mps can be auto-detected.

lxning:
@udaij12 in general, LGTM. I just left a minor comments https://github.com/pytorch/serve/pull/3048#discussion_r1556093071. Thanks.

lxning (approved):

agunapal (approved):
LGTM

PR #3073, opened 2024-04-03
Use weights_only for load

kit1980:
## Description

`torch.load` without `weights_only` parameter is unsafe. Explicitly set `weights_only` to False only if you trust the data you load and full pickle functionality is needed, otherwise set `weights_only=True`.

If `weights_only=True` doesn't work for some cases, then explicit `weights_only=False` should be used.

Found with https://github.com/pytorch-labs/torchfix/

## Type of change

This should be a no-op.


kit1980:
The actual changes are in the first commit.
The rest are to fix unrelated pre-existing issues in the files to pass lint.

agunapal:
CI should pass..not sure why it was failing.. Hopefully works this time

msaroufim:
@agunapal that's fine I think this is important to look at regardless

PR #3074, opened 2024-04-04
Update deprecated TorchVision and PyTorch APIs

kit1980:
## Description

1. For TorchVision models, pretrained parameters have been deprecated in favor of "Multi-weight support API" - see https://pytorch.org/vision/0.15/models.html
2. `torch.qr` updated to `torch.linalg.qr`

Found with https://github.com/pytorch-labs/torchfix/

Also applied pre-commit to pass linter.

## Type of change

This should be a no-op.


msaroufim (approved):
Thanks!

PR #3083, opened 2024-04-11
Fix GRPC address assignment to localhost by default

namannandan:
## Description
In the current implementation, the configured GRPC port is by default associated with the wildcard address

https://github.com/pytorch/serve/blob/cdba0fd449c2fd23dcf37c54c0784035541d5114/frontend/server/src/main/java/org/pytorch/serve/ModelServer.java#L450

```
$ torchserve --ncs --start --model-store ./model-store

$ netstat -an | grep 7070
Proto Recv-Q Send-Q  Local Address          Foreign Address        (state)
tcp46      0      0  *.7070                 *.*                    LISTEN

$ netstat -an | grep 7071
Proto Recv-Q Send-Q  Local Address          Foreign Address        (state)
tcp46      0      0  *.7071                 *.*                    LISTEN
```

Note that the grpc ports `7070` and `7071` are associated with the wildcard address `*`.

With the fix in this PR, the grpc ports are associated with localhost(`127.0.0.1`) by default.

```
$ torchserve --ncs --start --model-store ./model-store

$ netstat -an | grep 7070
Proto Recv-Q Send-Q  Local Address          Foreign Address        (state)
tcp46      0      0  127.0.0.1.7070                 *.*                    LISTEN

$ netstat -an | grep 7071
Proto Recv-Q Send-Q  Local Address          Foreign Address        (state)
tcp46      0      0  127.0.0.1.7071                 *.*                    LISTEN
```

In case of docker, the grpc address will need to be configured to `0.0.0.0` similar to the http management and inference addresses because otherwise, the grpc endpoint will not be accessible from outside the container:
https://github.com/pytorch/serve/blob/cdba0fd449c2fd23dcf37c54c0784035541d5114/docker/config.properties#L1-L2

Documentation already includes best practice to bind `localhost` ports to the ports exposed by docker to ensure access only to the host on which the container is running.
https://github.com/pytorch/serve/tree/master/docker#security-guideline

## Type of change

- [ ] CI

- [ ] Manual Test
```
$ torchserve --ncs --start --model-store ./model-store

$ python ts_scripts/torchserve_grpc_client.py register densenet161
## Check densenet161.mar in mar_set : set()
## Register marfile: https://torchserve.s3.amazonaws.com/mar_files/densenet161.mar

Model densenet161 registered successfully

$ python ts_scripts/torchserve_grpc_client.py infer densenet161 examples/image_classifier/kitten.jpg
{
  "tabby": 0.46661901473999023,
  "tiger_cat": 0.46449053287506104,
  "Egyptian_cat": 0.06614045053720474,
  "lynx": 0.0012924427865073085,
  "plastic_bag": 0.00022909742256160825
}

$ python ts_scripts/torchserve_grpc_client.py unregister densenet161
Model densenet161 unregistered successfully
```

### DIFF
--- a/docker/config.properties
+++ b/docker/config.properties
@@ -1,6 +1,8 @@
 inference_address=http://0.0.0.0:8080
 management_address=http://0.0.0.0:8081
 metrics_address=http://0.0.0.0:8082
+grpc_inference_address=0.0.0.0
+grpc_management_address=0.0.0.0
 number_of_netty_threads=32
 job_queue_size=1000
 model_store=/home/model-server/model-store
--- a/frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
+++ b/frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
@@ -359,20 +359,27 @@ public final class ConfigManager {
         return Connector.parse(binding, connectorType);
     }
 
-    public InetAddress getGRPCAddress(ConnectorType connectorType) throws UnknownHostException {
+    public InetAddress getGRPCAddress(ConnectorType connectorType)
+            throws UnknownHostException, IllegalArgumentException {
         if (connectorType == ConnectorType.MANAGEMENT_CONNECTOR) {
             return InetAddress.getByName(prop.getProperty(TS_GRPC_MANAGEMENT_ADDRESS, "127.0.0.1"));
-        } else {
+        } else if (connectorType == ConnectorType.INFERENCE_CONNECTOR) {
             return InetAddress.getByName(prop.getProperty(TS_GRPC_INFERENCE_ADDRESS, "127.0.0.1"));
+        } else {
+            throw new IllegalArgumentException(
+                    "Connector type not supporte dy gRPC: " + connectorType);
         }
     }
 
-    public int getGRPCPort(ConnectorType connectorType) {
+    public int getGRPCPort(ConnectorType connectorType) throws IllegalArgumentException {
         String port;
         if (connectorType == ConnectorType.MANAGEMENT_CONNECTOR) {
             port = prop.getProperty(TS_GRPC_MANAGEMENT_PORT, "7071");
-        } else {
+        } else if (connectorType == ConnectorType.INFERENCE_CONNECTOR) {
             port = prop.getProperty(TS_GRPC_INFERENCE_PORT, "7070");
+        } else {
+            throw new IllegalArgumentException(
+                    "Connector type not supporte dy gRPC: " + connectorType);
         }
         return Integer.parseInt(port);
     }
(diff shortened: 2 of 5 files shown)


frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
```diff
@@ -357,6 +359,14 @@ public Connector getListener(ConnectorType connectorType) {
         return Connector.parse(binding, connectorType);
     }
 
+    public InetAddress getGRPCAddress(ConnectorType connectorType) throws UnknownHostException {
+        if (connectorType == ConnectorType.MANAGEMENT_CONNECTOR) {
+            return InetAddress.getByName(prop.getProperty(TS_GRPC_MANAGEMENT_ADDRESS, "127.0.0.1"));
+        } else {
```
msaroufim, line 365: should we be more explicit in the if else that the second clause is only for inference?

docker/config.properties
```diff
@@ -1,6 +1,8 @@
 inference_address=http://0.0.0.0:8080
 management_address=http://0.0.0.0:8081
 metrics_address=http://0.0.0.0:8082
+grpc_inference_address=0.0.0.0
```
msaroufim, line 4: how is the port set for now?

namannandan: It uses the default values from here:
https://github.com/pytorch/serve/blob/89c5389d5dba23c32fbed7e7d58234fc30b690a5/frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java#L360-L368

And is called here: https://github.com/pytorch/serve/pull/3083/files#diff-45f9e03a9d3046ef0e2c18e18b0bb9c1889f9e3ff9918568870539cc926e3895R454

namannandan: Added configuration option to set grpc address separately from port to maintain backwards compatibility to retain the grpc port configuration:

https://github.com/pytorch/serve/blob/89c5389d5dba23c32fbed7e7d58234fc30b690a5/frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java#L104-L105

msaroufim (approved):

docker/config.properties
agunapal: can we keep this consistent?
`grpc_inference_address=http://0.0.0.0:7070`
`grpc_management_address=http://0.0.0.0:7071`

namannandan: It is possible to support configuration format as `grpc_inference_address=http://0.0.0.0:7070` but we've had support to configure the GRPC ports alone since `v0.3.0`:
https://github.com/pytorch/serve/blob/89c5389d5dba23c32fbed7e7d58234fc30b690a5/frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java#L104-L105

To maintain backwards compatibility, I believe we'll need to retain the grpc port configuration.

If we were to allow specifying the port along with the address as `grpc_inference_address=http://0.0.0.0:7070`, I believe we'll have two options:
1. Ignore the grpc port configuration ex `grpc_inference_port` if `grpc_inference_address` includes the port already.
2. Override the port specified in `grpc_inference_address` with the port specified `grpc_inference_port`.

Let me know your thoughts or suggestions on better ways to handle it.

PR #3116, opened 2024-04-26
Openvino integration into torchserve

suryasidd:
## Description

This PR adds initial support for integrating OpenVINO backend for torch.compile into torchserve. With this change users will be able to serve their pytorch models on torchserve and accelerate them using OpenVINO on Intel hardware.   

## Type of change

## Checklist:

- [ ] Did you have fun?


examples/pt2/torch_compile_openvino/stable_diffusion/requirements.txt
```diff
@@ -0,0 +1 @@
+diffusers==0.27.2
```
agunapal, line 1: Can we change this to  `>=0.27.2`

We have dependabot alerts which will keep flagging this otherwise

examples/pt2/torch_compile_openvino/stable_diffusion/model_config.yaml
```diff
@@ -0,0 +1,19 @@
+minWorkers: 1
+maxWorkers: 1
+maxBatchDelay: 200
+responseTimeout: 3600
+deviceType: "cpu"
```
agunapal, line 5: Are we testing with CPU or GPU?

examples/pt2/torch_compile_openvino/README.md
```diff
@@ -0,0 +1,146 @@
+
+# TorchServe inference with torch.compile with OpenVINO backend of vgg16 model
```
agunapal, line 2: It will be good if  this is changed to ResNet 50. We run nightly benchmarks for torch.compile with ResNet50. So, this will be an easy addition.

examples/pt2/torch_compile_openvino/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+This example shows how to take eager model of `vgg16`, configure TorchServe to use `torch.compile` and run inference using `torch.compile` with [OpenVINO backend](https://docs.openvino.ai/2024/openvino-workflow/torch-compile.html) and compare the performance
+
+This example has been tested on Intel Xeon Platinum 8469 CPU.
+
+### Pre-requisites
+
+- `PyTorch >= 2.0`
```
agunapal, line 10: Do all the options work with 2.0 or does it need 2.2/2.3?

examples/pt2/torch_compile_openvino/README.md
suryasidd: Sure will change this

examples/pt2/torch_compile_openvino/README.md
suryasidd: All the options work with 2.0 or higher

examples/pt2/torch_compile_openvino/stable_diffusion/model_config.yaml
suryasidd: If we add gpu to the device type, it's defaulting to nvidia execution. By keeping this to cpu and setting the device in openvino options to GPU, OpenVINO will internally run the model on GPU

agunapal: @suryasid This sounds like a bug in TorchServe. Expected behavior is to default to GPU if no config is specified or deviceType: "gpu"
Please file an issue for the same

setup.py
```diff
@@ -180,6 +180,7 @@ def run(self):
         extras_require={
             "onnx": ["numpy", "onnx", "onnx-runtime"],
             "ipex": ["intel_extension_for_pytorch"],
+            "openvino": ["openvino"],
```
agunapal, line 183: I'm not sure if we should be doing this.
Increasing size of docker images is a concern and we don't want to add to the list.
 
We want to follow what we did with TensorRT where the user needs to include tensorrt in requirements.txt. Can you please change this. https://github.com/pytorch/serve/blob/master/examples/torch_tensorrt/README.md

agunapal (changes requested):

setup.py
suryasidd: Made the change

agunapal (approved):
LGTM

PR #3128, opened 2024-05-01
Fix CI and Regression workflows for MAC Arm64

namannandan:
## Description
The Mac CI and regression failures are being caused by the following reasons:
1. The `macos-lastest` github runner tag now points to `arm64` based Mac.
2. PyTorch 2.3 has dropped support for `x86` Mac.
3. The M1 mac github runners do not expose the GPU(MPS hardware)
https://github.com/actions/runner-images/issues/7085
https://github.com/actions/runner-images/issues/9254#issuecomment-1968570672

This PR fixes the above issues by:
1. Fixing the GPU detection logic when running on M1 Mac to fall back to CPU. Note that the Mac CI and Regression only runs on CPU, when running on the github runner. To run on MPS, a self hosted runner will need to be created.
2. Update CI and regression tests to run on both CPU only and MPS on M1 Mac

## Type of change

- [ ] Mac `arm64` CI: https://github.com/pytorch/serve/actions/runs/8976048758/job/24652034356?pr=3128
- [ ] Mac `arm64` Regression: https://github.com/pytorch/serve/actions/runs/8976048741/job/24652042161?pr=3128


mreso (approved):
LGTM

PR #3146, opened 2024-05-21
Asynchronous worker communication and vllm integration

mreso:
## Description
This PR adds a new asynchronous communications mode to torchServe in order to accommodate inferences engines like vLLM.

Context:
These engines usually include techniques like pages attention which improve memory utilization by fitting as many requests into the engine as possible and evacuate requests if space gets tight. This means the classical synchronized communication model of TorchServe (batch of N requests in -> batch of N requests out) does not work well with these engines. This feature now offer to forward all incoming requests to the backend where they are fed to the engine which start producing responses which are streamed out asynchronously.

Multi-worker note:
While this in theory works with multiple workers it would distribute the incoming requests in a round robin fashion which might lead to non optimal worker/hardware utilization. It is therefore advised to only use a single worker and utilize tensor parallelism to distribute the model over multiple GPUs.

Fixes #(issue)

## Type of change

- [X] pytest test/pytest/test_async_worker_comm.py test/pytest/test_example_vllm.py
 ```
====================================================================================================== test session starts ======================================================================================================
platform linux -- Python 3.10.12, pytest-7.3.1, pluggy-1.3.0
rootdir: /home/ubuntu/serve
plugins: mock-3.10.0, anyio-4.3.0, cov-4.1.0
collected 2 items

test/pytest/test_async_worker_comm.py .                                                                                                                                                                                   [ 50%]
test/pytest/test_example_vllm.py .                                                                                                                                                                                        [100%]

====================================================================================================== 2 passed in 31.83s =======================================================================================================
```

## Checklist:

- [X] Did you have fun?

### CODE REVIEW

File: examples/large_models/vllm/base_vllm_handler.py [hunk 1]
```diff
@@ -21,59 +24,64 @@ def __init__(self):
...
-                ],
+        stop_time = time.time()
+        metrics.add_time(
+            "HandlerTime", round((stop_time - start_time) * 1000, 2), None, "ms"
+        )
+        return output
+
+    async def preprocess(self, requests):
+        input_batch = []
+        assert len(requests) == 1, "Expecting batch_size = 1"
```
namannandan, line 53: Is this to ensure that the async request handling in the frontend immediately forwards the request to the backend without waiting to form a batch?

frontend/server/src/main/java/org/pytorch/serve/wlm/AsyncBatchAggregator.java
```diff
@@ -0,0 +1,196 @@
...
+                            prediction.getResp(),
+                            prediction.getContentType(),
+                            prediction.getStatusCode(),
+                            prediction.getReasonPhrase(),
+                            prediction.getHeaders());
+                } else {
+                    logger.warn(
+                            "Drop response for inference request {} due to client timeout",
+                            job.getPayload().getRequestId());
+                }
```
namannandan, line 109: For my understanding, when a job expires, the behavior we expect here is that the backend continues to stream responses until the stopping criteria is reached but the frontend does not send it back to the client? Do we have a way to signal to backend about expired jobs and clean up `jobs_in_backend` here?

frontend/server/src/main/java/org/pytorch/serve/wlm/AsyncBatchAggregator.java
```diff
@@ -0,0 +1,196 @@
...
+            }
+            if (!jobs_in_backend.isEmpty()) {
+                // cleanJobs();
+                logger.error("Not all jobs got an error response.");
+            }
+        } else {
+            // Send the error message to all the jobs
+            for (Map.Entry<String, Job> j : jobs_in_backend.entrySet()) {
+                String jobsId = j.getValue().getJobId();
+                Job job = jobs_in_backend.get(jobsId);
```
namannandan, line 165:
```suggestion
                Job job = jobs_in_backend.remove(jobsId);
```
Should we update state of `jobs_in_backend` here as well, like above?

examples/large_models/vllm/base_vllm_handler.py
mreso: Correct, in engines like the vllm there is no classical setting as the batch size but number of concurrent requests is influenced by the available memory as well as request length. Therefore, send the request to the engine asap is the best option in my opinion.

frontend/server/src/main/java/org/pytorch/serve/wlm/AsyncBatchAggregator.java
mreso: Yes, it would be great to have this kind of functionality. But it will require some kind of command signaling into the backend which we can leverage. E.g. allowing for a C in the otf protocol which we discussed in the past also regarding up and downscaling of thread workers. We can add this functionality in a follow-up PR. For now the behavior will be as you described.

ts/async_service.py
```diff
@@ -0,0 +1,156 @@
...
+    return create_predict_response(
+        ret, req_id_map, "Prediction success", 200, context=context
+    )
+
+
+class AsyncService(object):
+    def __init__(self, service):
+        self.service = service
+        self.service.predict = types.MethodType(predict, self.service)
+        self.in_queue = Queue()
```
namannandan, line 94: Curious why we need a thread safe queue for the `in_queue` or we could safely use the non thread-safe async queue, similar to `out_queue` here as well?

ts/async_service.py
```diff
@@ -0,0 +1,156 @@
...
+        while True:
+            st = time.time()
+            batch = []
+            try:
+                request = self.in_queue.get()
+                batch += request
+                while len(batch) < BATCH_SIZE and (time.time() - st) < MAX_WAIT:
+                    timeout = max(0, MAX_WAIT - (time.time() - st))
+                    request = self.in_queue.get(timeout=timeout)
+                    batch += request
```
namannandan, line 125: Nit: We may need to check here if we actually got a `request` or we timed out before adding to `batch`

ts/async_service.py
```diff
@@ -0,0 +1,156 @@
...
+                AsyncQueue(loop=self.loop)
+                if sys.version_info <= (3, 9)
+                else AsyncQueue()
+            )
+            fetch = Thread(target=self.fetch_batches)
+            fetch.start()
+            receive = Thread(target=self.receive_requests)
+            receive.start()
+            send = Thread(target=self.send_responses)
+            send.start()
```
namannandan, line 149: If one or more of these threads exit/terminate and the others continue to run, what would the recovery look like? Would we have to restart the thread or somehow trigger an entire worker restart to ensure simple recovery?

namannandan (approved):
Few follow up questions on backend async service implementation, otherwise looks good to me!

ts/async_service.py
mreso: Yes, thats a good point. We need the thread safe queue here as async queue does not provide a timeout. The timeout is actually not strictly necessary for the vllm use case but I wanted to recreate the same batching functionality as in the frontend for future reference. If e.g. someone wants to run multiple workers with batching in the backend.

ts/async_service.py
mreso: When .get() times out which is the only alternative to getting a request an Empty expection is thrown which is caught in line 126 and will just restart the batch window.

ts/async_service.py
mreso: Good point, let me see if I can make this more fail resistant.

examples/large_models/vllm/Readme.md
```diff
@@ -13,3 +19,25 @@ vLLM [EngineArgs](https://github.com/vllm-project/vllm/blob/258a2c58d08fc7a24255
...
 vLLM [SamplingParams](https://github.com/vllm-project/vllm/blob/258a2c58d08fc7a242556120877a89404861fbce/vllm/sampling_params.py#L27) is defined in the JSON format, for example, [prompt.json](lora/prompt.json).
+
+### Distributed Inference
+All examples can be easily distributed over multiple GPUs by enabling tensor parallelism in vLLM.
+To enable distributed inference the following additions need to made to the model-config.yaml of the examples where 4 is the number of desired GPUs to use for the inference:
+
+```yaml
+# TorchServe frontend parameters
+...
+parallelType: "custom"
```
agunapal, line 30: What does `parallelType` mean? and what is `custom` . What are the other options?
We should perhaps talk about these options in a README under docs?

ts/model_service_worker.py
```diff
@@ -201,6 +203,51 @@ def handle_connection(self, cl_socket):
...
+            else:
+                raise ValueError(
+                    "Unexpected command (Only expecting LOAD command): {}".format(cmd)
+                )
+
+        service = AsyncService(service)
+
+        service.run()
+
+        # while True:
```
agunapal, line 236: Should this code block be removed?

examples/large_models/vllm/mistral/model-config.yaml
```diff
@@ -1,11 +1,10 @@
 # TorchServe frontend parameters
 minWorkers: 1
 maxWorkers: 1
-batchSize: 16
 maxBatchDelay: 100
 responseTimeout: 1200
 deviceType: "gpu"
-continuousBatching: true
+asyncCommunication: true
```
agunapal, line 7: What happens if a user specifies batchSize ? i.e Does this take care of ignoring dynamic batching in frontend?

examples/large_models/vllm/Readme.md
mreso: parallelType was already there, options are [tp, pp, and pptp](https://github.com/pytorch/serve/blob/079ff7b5d31d79627fa73d247c4f86a12893f8c5/examples/large_models/tp_llama/model-config.yaml#L6). Its basically the torchrun integration. "Custom" lets users customize the way they do parallelism. Will see if there is more doc on this and link here or add an issue to improve docs on parallelType.

ts/model_service_worker.py
mreso: yes, let me clean that up

examples/large_models/vllm/mistral/model-config.yaml
mreso: Choosing batch_size != 1 will currently trigger an assertion in the handler and there is not point in supporting it as it only delays processing of the request.

agunapal (approved):
LGTM

PR #3152, opened 2024-05-21
Pinning setuptools version

udaij12:
## Description
Pinning setuptools version due to newer version causing errors with pkg_resources


agunapal (approved):
LGTM

PR #3160, opened 2024-05-24
Support continuous batching in sequence batch streaming case

lxning:
## Description

Fixes #(issue)

## Type of change

```
 pytest test_example_stateful_sequence_continuous_batching_http.py
============================== test session starts ===============================
platform linux -- Python 3.10.14, pytest-7.3.1, pluggy-1.4.0
rootdir: /home/ubuntu/serve
plugins: cov-4.1.0, mock-3.12.0
collected 3 items

test_example_stateful_sequence_continuous_batching_http.py ...             [100%]

=============================== 3 passed in 19.81s ===============================
```

## Checklist:

- [ ] Did you have fun?


test/pytest/test_example_stateful_sequence_continuous_batching_http.py
```diff
@@ -0,0 +1,363 @@
...
+            elif sequence_id == "seq_1":
+                headers_seq_1 = {
+                    "ts_request_sequence_id": s_id,
+                }
+            start = False
+        else:
+            response = requests.post(
+                url=f"http://localhost:8080/predictions/{model_name}",
+                headers=headers_seq_0 if sequence_id == "seq_0" else headers_seq_1,
+                data=str(idx + 1).encode(),
```
mreso, line 312: For seq_0 we expect the last value to stay constant as we send 0, but we also send 0 to seq_1. Why do we expecty the last value to be 30 there?

test/pytest/test_example_stateful_sequence_continuous_batching_http.py
```diff
@@ -0,0 +1,363 @@
...
+    for idx in range(5):
+        if idx == 2:
+            cancel = True
+        if sequence_id == "seq_0":
+            idx = 2 * idx
+        elif sequence_id == "seq_1":
+            idx = 2 * idx + 1
+
+        if cancel is True and sequence_id == "seq_0":
+            response = requests.post(
```
mreso, line 333: This test does not cover the continuous batching part. We should add that in the test.

mreso (changes requested):
Left some comments. Please add continuous batching to the test to make sure this part works as well.

examples/stateful/sequence_continuous_batching/stateful_handler.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+
+        return data
+
+    def clean_up_seq(self, seq_id):
+        # clean up
+        del self.cache[seq_id]
+        del self.context.cache[seq_id]
+
+    def clean_up_req(self, seq_id, req_id):
```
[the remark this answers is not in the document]
lxning: they are used to clean up cache in 2 cases:
1. session end: clean this session in context cache
2. cancel: clean req in context cache for this session

frontend/server/src/main/java/org/pytorch/serve/util/messages/RequestInput.java
```diff
@@ -75,10 +79,10 @@ public void setClientExpireTS(long clientTimeoutInMills) {
     }
 
     public String getSequenceId() {
-        if (sequenceId == null) {
+        if (sequenceId.isEmpty()) {
```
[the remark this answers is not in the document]
lxning: I keep the code of setting requestinput as the original in http handler part. this function is only called by sequence batching case. If the sequence is already set as non-empty value, we don't need to check hash table

frontend/server/src/main/java/org/pytorch/serve/wlm/SequenceContinuousBatching.java
```diff
@@ -0,0 +1,138 @@
...
+                    jobGroup.setFinished(true);
+                }
+            }
+        }
+    }
+
+    @Override
+    protected void pollInferJob() throws InterruptedException {
+        // TBD: Temporarily hard code the continuous batch size is 2 * batchSize
+        model.pollInferJob(jobs, model.getBatchSize() * 2 - jobs.size(), jobsQueue);
```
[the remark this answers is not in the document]
lxning: No, it is not necessary for backend handler to sort b/c backend handler preprocessing is able to mark the previous running request as "cancel"

test/pytest/test_example_stateful_sequence_continuous_batching_http.py
lxning: b/c seq_1 call infer_stateful, not infer_stateful_end

test/pytest/test_example_stateful_sequence_continuous_batching_http.py
lxning: it is complicate to match the final result b/c it is dynamically in continuous batching case. Here, the client side does not enable chunked encode, which mean skipping the rest of response, to get the fixed the result.

frontend/server/src/main/java/org/pytorch/serve/wlm/SequenceContinuousBatching.java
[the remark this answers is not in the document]
mreso: @lxning So you mean the preprocessing in the handler will need to sort this out as there can be multiple requests from multiple sequences and in no specific pattern. In the end the preprocessing in the handler will need to go though the batch and see if besides the previous request another cancel request is in the batch as well and than cancel it/clean it up.

test/pytest/test_example_stateful_sequence_continuous_batching_http.py
mreso: You can just use sleep in the inference() call to make sure we get a cancellation request in after we get a certain response through the chunked response.

test/pytest/test_example_stateful_sequence_continuous_batching_http.py
mreso: got it, missed that, thanks

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkLoadManager.java
```diff
@@ -229,12 +229,10 @@ private void addThreads(
 
             BatchAggregator aggregator;
 
-            if (model.isSequenceBatching()) {
-                if (model.isSequenceContinuousBatch()) {
-                    aggregator = new SequenceContinuousBatching(model);
-                } else {
-                    aggregator = new SequenceBatching(model);
-                }
+            if (model.isSequenceContinuousBatch()) {
```
mreso, line 232: This is incorrect as it tests for isSequenceContinuousBatch in both cases. I also meant that we should not need to create the separate configuration keyword sequenceContinuousBatching but we can just rely on sequenceBatching + continuousBatching being set to True.

mreso (changes requested):
Please see comments.

frontend/server/src/main/java/org/pytorch/serve/wlm/SequenceContinuousBatching.java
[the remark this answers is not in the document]
lxning: No, there is no sort. it is based on "[num_requests](https://github.com/pytorch/serve/blob/4b43233381beec5034f4c2c65dc162d11e833fa5/examples/stateful/sequence_continuous_batching/stateful_handler.py#L185)"

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkLoadManager.java
lxning: remove model config "sequenceContinuousBatching"

examples/stateful/sequence_continuous_batching/stateful_handler.py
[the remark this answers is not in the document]
lxning: combine the 2 function as one.

mreso (approved):
LGTM

Issue #3157, opened 2024-05-23
Enable token authentication as default


lxning:
### 🚀 The feature

- By default, enable plugin token authentication.
- Allow cx to disable token authentication via config.properties.

### Motivation, pitch

Current token authentication is enabled via option "--plugins-path /path/to/the/jar/files".

Issue #3066, opened 2024-04-02
[RFC]: Deprecation notice for TorchText installation in TorchServe


agunapal:
### 🚀 The feature

TorchText release is being deprecated by PyTorch starting PyTorch 2.4 https://github.com/pytorch/text/issues/2250

TorchServe will follow the same. The release of TorchServe with PyTorch 2.4 will not have TorchText installed.

### Motivation, pitch

TorchText is not being maintained and hence TorchText Release is being deprecated.

### Alternatives

We would need to copy paste these functions being referenced in `text_handler.py` and `text_classifier.py`
https://github.com/pytorch/text/issues/2250#issuecomment-2033227578

ayushraj2349:
hello sir, i'm also facing the same issue.
can i still use it now till the release of 2.4?

can you please explain the working of alternatives again?

agunapal:
Hi @ayushraj2349 We merged this PR which removes the dependency we had in TorchServe https://github.com/pytorch/serve/pull/3076

PR #3163, opened 2024-05-29
Enable Token Authorization by default

udaij12:
## Description
Setting token authorization by default #3157

Users can disable using environment variables, command line, or with config.properties. 
Check Documentation for examples 

## Tests
Updating all existing tests to disable token authorization except for token authorization tests in pytest


docs/token_authorization_api.md
```diff
@@ -1,11 +1,14 @@
...
 
-## Setup
-1. Download the jar files from [Maven](https://mvnrepository.com/artifact/org.pytorch/torchserve-endpoint-plugin) 
-2. Enable token authorization by adding the `--plugins-path /path/to/the/jar/files` flag at start up with the path leading to the downloaded jar files.
+Torchserve now supports token authorization by default.
+
+## How to disable Token Authorization
+1. Set global environment variable `TS_DISABLE_TOKEN_AUTHORIZATION=true`
+2. Add `--disable-token` to command line when running TorchServe.
+3. Add `disable_token_authorization=true` to config.properties file
```
namannandan, line 8: Should we clarify that any one of these options can be used to disable token authorization?
```suggestion
## How to disable Token Authorization
1. Set global environment variable `TS_DISABLE_TOKEN_AUTHORIZATION=true` or
2. Add `--disable-token` to command line when running TorchServe or
3. Add `disable_token_authorization=true` to config.properties file
```

Also for `Set global environment variable TS_DISABLE_TOKEN_AUTHORIZATION=true` I think it may be good to link to https://github.com/pytorch/serve/blob/master/docs/configuration.md and mention that `enable_envvars_config=true` must be set to ensure that the environment variable takes effect.

docs/token_authorization_api.md
```diff
@@ -45,9 +48,7 @@
 ## Customization
 Torchserve offers various ways to customize the token authorization to allow owners to reach the desired result.
```
namannandan, line 49: Could we double check that this statement on customizability is still valid. In case this is not intended, we can delete it or test using a plugin to see if `token` implementation provided in the frontend can be overridden by a plugin.

lxning (commented):
Why not enable the token authentication plugin by default to avoid big code changes and  copying/pasting Token class?

frontend/server/src/main/java/org/pytorch/serve/http/TokenAuthorizationHandler.java
```diff
@@ -44,36 +60,46 @@ public void handleRequest(
...
                 checkTokenAuthorization(req, "inference");
+                chain.handleRequest(ctx, req, decoder, segments);
             }
+        } else {
+            if (tokenType == TokenType.MANAGEMENT && req.toString().contains("/token")) {
+                throw new ResourceNotFoundException();
+            }
+            chain.handleRequest(ctx, req, decoder, segments);
         }
-        chain.handleRequest(ctx, req, decoder, segments);
```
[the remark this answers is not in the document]
udaij12: Noted

docs/token_authorization_api.md
udaij12: Will remove customization section and move time to expiration to notes.  Plugin implementation no longer works

frontend/server/src/main/java/org/pytorch/serve/http/TokenAuthorizationHandler.java
```diff
@@ -44,36 +60,46 @@ public void handleRequest(
...
                 logger.info("TOKEN CLASS IMPORTED SUCCESSFULLY");
             }
-        } catch (NoSuchMethodException
-                | IllegalAccessException
-                | InstantiationException
-                | InvocationTargetException
-                | ClassNotFoundException e) {
+        } catch (Exception e) {
             e.printStackTrace();
             logger.error("TOKEN CLASS IMPORTED UNSUCCESSFULLY");
```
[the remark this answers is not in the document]
udaij12: Kept it so that when TorchServe starts users can see if the keyfile was created successfully if enabled.

udaij12:
> Why not enable the token authentication plugin by default to avoid big code changes and copying/pasting Token class?

I feel that if we are implementing token authorization as default it makes sense to integrate it into the frontend rather than as an external plugin as there is no need for that. It would also make future changes painful since we would have to build the plugin jar and publish to maven. It makes sense to integrate and remove the plugin completely so rather then duplication of code its just moving.

lxning (commented):
Can you add test for the priority b/w config file vs cmd, especially for the negative test cases (eg. conflict setting)?

udaij12:
> Can you add test for the priority b/w config file vs cmd, especially for the negative test cases (eg. conflict setting)?

Added the test and added more clarification in the docs.

docs/token_authorization_api.md
```diff
@@ -1,11 +1,28 @@
 # TorchServe token authorization API
 
-## Setup
-1. Download the jar files from [Maven](https://mvnrepository.com/artifact/org.pytorch/torchserve-endpoint-plugin) 
-2. Enable token authorization by adding the `--plugins-path /path/to/the/jar/files` flag at start up with the path leading to the downloaded jar files.
+Torchserve now supports token authorization by default.
```
agunapal, line 3: @udaij12 Please create an RFC for this if not already done, before the PR lands

agunapal (changes requested):

docs/token_authorization_api.md
```diff
@@ -41,14 +58,7 @@
...
-2. The token authorization code is consolidated in the plugin and thus can be changed without impacting the frontend or end result. The only thing the user cannot change is:
-    1. The urlPattern for the plugin must be 'token' and the class name must not change
-    2. The `generateKeyFile`, `checkTokenAuthorization`, and `setTime` functions return type and signature must not change. However, the code in the functions can be modified depending on user necessity.
-
 ## Notes
 1. DO NOT MODIFY THE KEY FILE. Modifying the key file might impact reading and writing to the file thus preventing new keys from properly being displayed in the file.
-2. 3 tokens allow the owner with the most flexibility in use and enables them to adapt the tokens to their use. Owners of the server can provide users with the inference token if users should not mess with models. The owner can also provide owners with the management key if owners want users to add and remove models.
+2. Time to expiration is set to default at 60 minutes but can be changed in the config.properties by adding `token_expiration_min`. Ex:`token_expiration_min=30`
```
agunapal, line 63: How is this going to work in a production setting? What would be the workflow?

docs/token_authorization_api.md
agunapal, line 3: I suggest we specify this in multiple places: docker, kubernetes, kserve etc.
Its better to over-communicate this basic behavior change.

agunapal (changes requested):

docs/token_authorization_api.md
udaij12: What is an RFC Im not aware of what that is.

docs/token_authorization_api.md
udaij12: The workflow remains the same for production settings. Customers can disable token authorization to keep things the same or can change the token expiration depending on their use case. If a token is expired users have to reset the token and then use the updated token for future API calls.

docs/token_authorization_api.md
udaij12: Agreed. Was planning on creating a PR of documentation updates with both security related changes to inform of new behavior to TorchServe.

agunapal: Example: https://github.com/pytorch/serve/issues/3066

agunapal (approved):
LGTM. Please create an RFC like https://github.com/pytorch/serve/issues/3066 before merging

namannandan (approved):

frontend/server/src/main/java/org/pytorch/serve/http/TokenAuthorizationHandler.java
```diff
@@ -44,36 +60,44 @@ public void handleRequest(
...
+                        throw new InvalidKeyException(
+                                "Token Authentication failed. Token either incorrect, expired, or not provided correctly");
+                    }
                 } else {
                     checkTokenAuthorization(req, "management");
                 }
             } else if (tokenType == TokenType.INFERENCE) {
                 checkTokenAuthorization(req, "inference");
             }
+        } else {
```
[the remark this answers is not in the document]
udaij12: This is to prevent the token api from being called when token authorization is disabled.

frontend/server/src/main/java/org/pytorch/serve/http/TokenAuthorizationHandler.java
```diff
@@ -84,20 +108,208 @@ public static void setupTokenClass() {
...
     }
 }
+
+class Token {
+    private static String apiKey;
+    private static String managementKey;
+    private static String inferenceKey;
+    private static Instant managementExpirationTimeMinutes;
+    private static Instant inferenceExpirationTimeMinutes;
+    private static Double timeToExpirationMinutes;
```
mreso, line 129: Lets get rid if the double bookkeeping here and clean out the code. Token is now part of TS so it can access ConfigManager. We can just read the time from there to have a single source of truth.

frontend/server/src/main/java/org/pytorch/serve/http/TokenAuthorizationHandler.java
```diff
@@ -24,7 +40,7 @@ public class TokenAuthorizationHandler extends HttpRequestHandlerChain {
     private static final Logger logger = LoggerFactory.getLogger(TokenAuthorizationHandler.class);
     private static TokenType tokenType;
     private static Boolean tokenEnabled = false;
-    private static Class<?> tokenClass;
+    private static Token tokenClass;
     private static Object tokenObject;
     private static Double timeToExpirationMinutes = 60.0;
```
mreso, line 45: See comment on ```class Token``` and lets assume source of through to be config manager

frontend/server/src/main/java/org/pytorch/serve/http/TokenAuthorizationHandler.java
```diff
@@ -84,20 +108,208 @@ public static void setupTokenClass() {
...
+    private static String inferenceKey;
+    private static Instant managementExpirationTimeMinutes;
+    private static Instant inferenceExpirationTimeMinutes;
+    private static Double timeToExpirationMinutes;
+    private SecureRandom secureRandom = new SecureRandom();
+    private Base64.Encoder baseEncoder = Base64.getUrlEncoder();
+    private String fileName = "key_file.json";
+    private String filePath = "";
+
+    public String updateKeyFile(FullHttpRequest req) throws IOException {
```
mreso, line 135: Lets not pull HttpRequest into this function but instead call generateKeyFile directly and parse request outside of Token. This class is highly critical, so we should keep it as lean as possible.

frontend/server/src/main/java/org/pytorch/serve/http/TokenAuthorizationHandler.java
```diff
@@ -84,20 +108,208 @@ public static void setupTokenClass() {
...
+        } else if ("inference".equals(queryResponse)) {
+            generateKeyFile("inference");
+        } else {
+            test = "{\n\t\"Error\": " + queryResponse + "\n}\n";
+        }
+        return test;
+    }
+
+    // parses query and either returns management/inference or a wrong type error
+    public String parseQuery(FullHttpRequest req) {
```
mreso, line 149: Same here. This should be part of the handler not token.

frontend/server/src/main/java/org/pytorch/serve/http/TokenAuthorizationHandler.java
```diff
@@ -84,20 +108,208 @@ public static void setupTokenClass() {
...
+        return baseEncoder.encodeToString(randomBytes);
+    }
+
+    public Instant generateTokenExpiration() {
+        long secondsToAdd = (long) (timeToExpirationMinutes * 60);
+        return Instant.now().plusSeconds(secondsToAdd);
+    }
+
+    public void setFilePath(String path) {
+        filePath = path;
```
mreso, line 175: Lets read this directly from the config as well.

frontend/server/src/main/java/org/pytorch/serve/http/TokenAuthorizationHandler.java
```diff
@@ -84,20 +108,208 @@ public static void setupTokenClass() {
...
+            Set<PosixFilePermission> permissions = PosixFilePermissions.fromString("rw-------");
+            Files.setPosixFilePermissions(path, permissions);
+        } catch (Exception e) {
+            return false;
+        }
+        return true;
+    }
+
+    // checks the token provided in the http with the saved keys depening on parameters
+    public boolean checkTokenAuthorization(FullHttpRequest req, String type) {
```
mreso, line 251: Same, lets move this up to handler and only get type + token as parameter

frontend/server/src/main/java/org/pytorch/serve/http/TokenAuthorizationHandler.java
```diff
@@ -84,20 +108,208 @@ public static void setupTokenClass() {
...
+
+    public String getInferenceKey() {
+        return inferenceKey;
+    }
+
+    public String getKey() {
+        return apiKey;
+    }
+
+    public Instant getInferenceExpirationTime() {
```
mreso, line 304: Unused?

frontend/server/src/main/java/org/pytorch/serve/http/TokenAuthorizationHandler.java
```diff
@@ -84,20 +108,208 @@ public static void setupTokenClass() {
...
+        return inferenceExpirationTimeMinutes;
+    }
+
+    public Instant getManagementExpirationTime() {
```
mreso, line 308: Unused?

frontend/server/src/main/java/org/pytorch/serve/http/TokenAuthorizationHandler.java
```diff
@@ -84,20 +108,208 @@ public static void setupTokenClass() {
...
+        return managementExpirationTimeMinutes;
+    }
+
+    public void setTime(Double time) {
```
mreso, line 312: See above. Use config manager.

frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
```diff
@@ -46,6 +46,7 @@
 import org.apache.commons.cli.Options;
 import org.apache.commons.io.IOUtils;
 import org.pytorch.serve.archive.model.Manifest;
+import org.pytorch.serve.http.TokenAuthorizationHandler;
```
mreso, line 49: See above

frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
```diff
@@ -438,6 +445,14 @@ public boolean isOpenInferenceProtocol() {
         return Boolean.parseBoolean(prop.getProperty(TS_OPEN_INFERENCE_PROTOCOL, "false"));
     }
 
+    public boolean setupToken() {
```
mreso, line 448: See above

frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
```diff
@@ -1203,6 +1230,10 @@ public String getWorkflowStore() {
             return workflowStore;
         }
 
+        public String isTokenDisabled() {
```
mreso, line 1233: Why do we use a string here instead of boolean?

test/pytest/test_token_authorization.py
```diff
@@ -11,26 +9,9 @@
...
-    subprocess.run(["./gradlew", "formatJava"])
-    result = subprocess.run(["./gradlew", "build"])
-    jar_path = os.path.join(plugin_folder, "endpoints/build/libs")
-    jar_files = [file for file in os.listdir(jar_path) if file.endswith(".jar")]
-    for jar_file in jar_files:
-        shutil.move(
-            os.path.join(jar_path, jar_file), os.path.join(new_folder_path, jar_file)
-        )
-    os.chdir(REPO_ROOT)
+REPO_ROOT = os.path.dirname(os.path.abspath(__file__))
```
mreso, line 12: nit: Better to rename as its not repo root anymore.

test/pytest/test_utils.py
```diff
@@ -59,6 +59,7 @@ def start_torchserve(
     no_config_snapshots=False,
     gen_mar=True,
     plugin_folder=None,
+    token=False,
```
mreso, line 62: lets make this less an a logic puzzle and name is disable_token so we do not need to invert below.

test/pytest/test_token_authorization.py
```diff
@@ -215,3 +193,25 @@ def test_token_expiration_time(setup_torchserve_expiration):
...
+    )
+    test_utils.start_torchserve(
+        snapshot_file=config_file_priority, no_config_snapshots=True
+    )
+
+    response = requests.get(f"http://localhost:8081/models")
+
+    test_utils.stop_torchserve()
+
+    assert response.status_code == 200, "Token check failed"
```
mreso, line 217: These is a test missing for the env variable and the different priorities with it.

ts_scripts/tsutils.py
```diff
@@ -48,6 +48,7 @@ def start_torchserve(
     config_file="",
     log_file="",
     gen_mar=True,
+    token=False,
```
mreso, line 51: Same here, change to disable_token

mreso (changes requested):
The tests are not sufficient. We need to add disable through env variable and the different priorities. We also need to test updating the expired and not expired tokens (of both types).
I've left a bunch of other comments which are more related to a refactoring as we're e.g. introducing dependency cycles here. I'll create a follow up issue where we need to address these on a short timeline. We can ignore these for now as this is time sensitive but the testing needs to be improved before we can merge this.

udaij12:
Added new tests 1. generating new tokens once expired 2. priority with env, cmd, and config properties 3. priority with env and cmd

mreso (approved):
Lets go ahead with this with the expectation that https://github.com/pytorch/serve/issues/3185 is addressed in the short-term.

frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
udaij12: Using this function to set prop and prop takes strings not booleans.

Issue #3158, opened 2024-05-23
Support "model-control-mode" in configuration


lxning:
### 🚀 The feature

- Support model-control-mode:
1. none:  Models are loaded at start up and can not be modified. (default).
2. explicit: Models can be loaded and unloaded via the model load APIs.

- Support model-control-mode is configurable via config.properties 

### Motivation, pitch

For security reason, it is important to prevent malicious code is uploaded in model server.

udaij12:
Model API control has been added. Default is model API disabled preventing registering and deleting models once TorchServe starts.

PR #3165, opened 2024-05-30
Adding model-control-mode

udaij12:
## Description
Adding a model-control-mode to TorchServe #3158

Default is `none` which prevents users from registering and unregistering models after startup. 
Can be set to `enabled` which allows users to use register and unregister APIs

Check documentation for how to set the model-control-mode

## Tests
Setting model mode to "enabled" for all current tests. 
Added a new pytest [test_model_control_mode.py](https://github.com/pytorch/serve/pull/3165/files#diff-ca05115302bff77e9d174711f74dd947c611fc5ffb7c8fdf939bedf178637e33) that covers default and model api enabled


docs/model_control_mode.md
```diff
@@ -0,0 +1,152 @@
+# Model Control Mode
+
+TorchServe now supports model control mode with two settings "none"(default) and "explicit"
+
+## How to set Model Control to explicit
+1. Add `--model-mode-explicit` to command line when running TorchServe.
+2. Add `model_control_mode=explicit` to config.properties file
```
lxning, line 7: What's the priority b/w the two options? We can only offer this config.properties if it is too complicate for you to implement.

docs/model_control_mode.md
```diff
@@ -0,0 +1,152 @@
...
+    },
+    {
+      "modelName": "squeezenet1_1",
+      "modelUrl": "https://torchserve.pytorch.org/mar_files/squeezenet1_1.mar"
+    }
+  ]
+}
+ubuntu@ip-172-31-11-32:~/serve$ torchserve --stop
+TorchServe has stopped.
+```
```
lxning, line 152: it is not necessary to post the log in the readme. You only need to demo accepting register/unregister curl response at most.

frontend/server/src/main/java/org/pytorch/serve/http/api/rest/ManagementRequestHandler.java
```diff
@@ -129,6 +130,11 @@ private boolean isManagementReq(String[] segments) {
                 || endpointMap.containsKey(segments[1]);
     }
 
+    private boolean checkModelControl() {
+        ConfigManager configManager = ConfigManager.getInstance();
+        return (configManager.getModelControlMode().equals("explicit"));
+    }
+
```
lxning, line 137: 1. it is not efficient to compare string each time. 
2. better to rename as isModelControlled
3. move it to ConfigManager.
4. and also it is kind of redundancy getModelControlMode

docs/model_control_mode.md
udaij12: Priority follows TorchServe standard as documented here https://github.com/pytorch/serve/blob/master/docs/configuration.md, will add a note in the docs that links to the priority

docs/model_control_mode.md
udaij12: Will clean up the logs so it shows the register/unregister curl and output

frontend/server/src/main/java/org/pytorch/serve/http/api/rest/ManagementRequestHandler.java
udaij12: Noted will move the check into ConfigManager

lxning (commented):
Can you add test for the priority b/w config file vs cmd, especially for the negative test cases (eg. conflict setting)?

frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
```diff
@@ -478,6 +484,11 @@ public int getNumberOfGpu() {
         return getIntProperty(TS_NUMBER_OF_GPU, 0);
     }
 
+    public boolean getModelControlMode() {
+        String mode = getProperty(MODEL_CONTROL_MODE, "none");
+        return mode.equals("explicit") ? true : false;
+    }
+
```
lxning, line 491: As I commented before, it is not efficient to compare every time. it should be only time job, return true/false directly starting from 2nd time

frontend/server/src/main/java/org/pytorch/serve/http/api/rest/ManagementRequestHandler.java
lxning: issue 1 is still not addressed.

frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
lxning: I mean "mode.equals("explicit")" is not efficient.

udaij12: It is only being compared one time now, like you mentioned that its not efficient every time. Are you saying to remove the string comparison all together? Currently it is only being compared once instead of during every check as before.

frontend/server/src/main/java/org/pytorch/serve/http/api/rest/ManagementRequestHandler.java
udaij12: Issue 1 has been resolved as the string comparison only happens once and then is a true/false directly starting from 2nd time

frontend/server/src/main/java/org/pytorch/serve/http/api/rest/ManagementRequestHandler.java
```diff
@@ -129,6 +132,14 @@ private boolean isManagementReq(String[] segments) {
...
     }
 
+    private boolean isModeExplicit() {
+        if (Modelmode == null) {
+            ConfigManager configManager = ConfigManager.getInstance();
+            Modelmode = configManager.getModelControlMode();
+        }
+        return Modelmode;
+    }
+
```
namannandan, line 142: This can be moved to the config manager `getModelControlMode`, so we can just query the model control mode here from the config manager and not have to store the configuration value.

test/pytest/test_model_control_mode.py
```diff
@@ -0,0 +1,115 @@
...
+    params = (
+        ("model_name", "resnet-18"),
+        ("url", "resnet-18.mar"),
+        ("initial_workers", "1"),
+        ("synchronous", "true"),
+    )
+    response = requests.post("http://localhost:8081/models", params=params)
+
+    assert response.status_code == 200, "model control check failed"
+    response = requests.get("http://localhost:8081/models/resnet-18")
```
namannandan, line 91: Please extend test to include model deletion with `mode:explicit` to ensure we cover that use case as well.

namannandan (changes requested):

docs/model_control_mode.md
```diff
@@ -0,0 +1,66 @@
...
+  "code": 405,
+  "type": "MethodNotAllowedException",
+  "message": "Requested method is not allowed, please refer to API document."
+}
+```
+
+## Model Control Mode EXPLICIT
+Setting model control to `explicit` allows users to load and unload models using the model load APIs.
+
+### Example using cmd line to set mode to explicit
```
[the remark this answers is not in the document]
udaij12: The default mode example already exists with no CLI input or config properties.

frontend/server/src/main/java/org/pytorch/serve/http/api/rest/ManagementRequestHandler.java
```diff
@@ -48,6 +49,8 @@
  */
 public class ManagementRequestHandler extends HttpRequestHandlerChain {
 
+    private Boolean Modelmode = null;
```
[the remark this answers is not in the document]
udaij12: Noted and moved

test/pytest/test_model_control_mode.py
udaij12: Noted and added

frontend/server/src/main/java/org/pytorch/serve/http/api/rest/ManagementRequestHandler.java
```diff
@@ -74,7 +75,7 @@ public void handleRequest(
                     if (HttpMethod.GET.equals(method)) {
                         handleListModels(ctx, decoder);
                         return;
-                    } else if (HttpMethod.POST.equals(method)) {
+                    } else if (HttpMethod.POST.equals(method) && isModeExplicit()) {
```
namannandan, line 78: We'll need to include this check for gRPC endpoints as well
https://github.com/pytorch/serve/blob/dd1f8dc36efe21f036be1e65043310dec74af28e/frontend/server/src/main/java/org/pytorch/serve/grpcimpl/ManagementImpl.java#L104

frontend/server/src/main/java/org/pytorch/serve/http/api/rest/ManagementRequestHandler.java
```diff
@@ -93,7 +94,7 @@ public void handleRequest(
                     } else {
                         handleScaleModel(ctx, decoder, segments[2], modelVersion);
                     }
-                } else if (HttpMethod.DELETE.equals(method)) {
+                } else if (HttpMethod.DELETE.equals(method) && isModeExplicit()) {
```
namannandan, line 97: We'll need to include this check for gRPC as well
https://github.com/pytorch/serve/blob/dd1f8dc36efe21f036be1e65043310dec74af28e/frontend/server/src/main/java/org/pytorch/serve/grpcimpl/ManagementImpl.java#L196

frontend/server/src/main/java/org/pytorch/serve/http/api/rest/ManagementRequestHandler.java
```diff
@@ -129,6 +130,11 @@ private boolean isManagementReq(String[] segments) {
                 || endpointMap.containsKey(segments[1]);
     }
 
+    private boolean isModeExplicit() {
+        ConfigManager configManager = ConfigManager.getInstance();
```
namannandan, line 134: Nit: `ConfigManager` is singleton so we can fetch this only once and save it in a class variable and just do `return configManager.getModelControlMode()` here.

frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
```diff
@@ -254,6 +256,11 @@ private ConfigManager(Arguments args) throws IOException {
             prop.setProperty(TS_LOAD_MODELS, String.join(",", models));
         }
 
+        boolean mode = args.isModeExplicit();
```
namannandan, line 259: Nit: 
```suggestion
        boolean modelControlMode = args.isModeControlModeExplicit();
```

frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
```diff
@@ -1117,6 +1134,7 @@ public static final class Arguments {
         private boolean snapshotDisabled;
         private String workflowStore;
         private String cppLogConfigFile;
+        private boolean modelExplicit;
```
namannandan, line 1137:
```suggestion
        private boolean modelControlModeExplicit;
```

frontend/server/src/main/java/org/pytorch/serve/http/api/rest/ManagementRequestHandler.java
udaij12: Created issue for gRPC

frontend/server/src/main/java/org/pytorch/serve/http/api/rest/ManagementRequestHandler.java
udaij12: Created issue for gRPC

frontend/server/src/main/java/org/pytorch/serve/http/api/rest/ManagementRequestHandler.java
```diff
@@ -48,9 +49,12 @@
  */
 public class ManagementRequestHandler extends HttpRequestHandlerChain {
 
+    ConfigManager configManager;
+
     /** Creates a new {@code ManagementRequestHandler} instance. */
     public ManagementRequestHandler(Map<String, ModelServerEndpoint> ep) {
         endpointMap = ep;
+        configManager = ConfigManager.getInstance();
```
namannandan, line 57: Make member variable private
```suggestion
        private configManager = ConfigManager.getInstance();
```

frontend/server/src/main/java/org/pytorch/serve/http/api/rest/ManagementRequestHandler.java
namannandan: https://github.com/pytorch/serve/issues/3180

frontend/server/src/main/java/org/pytorch/serve/http/api/rest/ManagementRequestHandler.java
namannandan: https://github.com/pytorch/serve/issues/3180

test/pytest/sanity/coverage.xml
namannandan, line 1: Please remove this file from your PR.

namannandan (approved):
Few minor comments to address. Looks good otherwise!

docs/model_control_mode.md
```diff
@@ -0,0 +1,68 @@
...
+    Result: Explicit mode enabled
+* Example 2:
+  * Config file: `model_control_mode=explicit`
+
+    cmd line: `torchserve --start --ncs --model-store model_store`
+
+    Result: Mode is explicit (no way to enable "none" through cmd)
+
+## Model Control Mode Default
+At startup TorchServe loads only those models specified explicitly with the `--models` command-line option. After startup users will be unable to register or delete models in this mode.
```
mreso, line 24: The naming of the modes (none, explicit) does not make sense for TorchServe. Using the wording "explicitly" in the doc for the default (none) mode will confuse users even more.

The difference between other solutions (where this naming makes sense) is that TorchServe does not load all models in model_store by default but none. Only if we give --models we load the specified. So for TorchServe this PR is more about locking the api after startup so no models can be loaded afterwards. This should also be reflected in the parameter. 

To make clear what the parameter does we will need to name it something like: --model-control-api-enabled

frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
```diff
@@ -478,6 +485,14 @@ public int getNumberOfGpu() {
         return getIntProperty(TS_NUMBER_OF_GPU, 0);
     }
 
+    public boolean getModelControlMode() {
+        String mode = getProperty(MODEL_CONTROL_MODE, "none");
+        if (modelMode == null) {
```
mreso, line 490: To avoid getting the property getProperty should be inside the condition.

test/pytest/test_utils.py
```diff
@@ -59,6 +59,8 @@ def start_torchserve(
     no_config_snapshots=False,
     gen_mar=True,
     plugin_folder=None,
+    models=None,
+    mode=None,
```
mreso, line 63: The default mode is called "none" in the docs. But here None selects explicit. Better to revert the logic to no confuse people.

mreso (changes requested):
The chosen parameter name is not reflecting the actual functionality in TorchServe. Please see comments.

docs/model_control_mode.md
udaij12: changed it to --model-api-enabled and changed explicit to enabled in docs and code

frontend/server/src/main/java/org/pytorch/serve/util/ConfigManager.java
udaij12: done

test/pytest/test_utils.py
udaij12: The logic for test is to keep enabled as default so that it does not break existing test structure for now. Will require complete test overhaul at a later date.

docs/model_control_mode.md
```diff
@@ -0,0 +1,68 @@
+# Model Control Mode
+
+TorchServe now supports model control mode with two settings "none"(default) and "enabled"
```
namannandan, line 3: Please add a short description here of what the two modes do and which APIs they disable/enable.

docs/model_control_mode.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Two ways to set Model Control
+1. Add `--model-api-enabled` to command line when running TorchServe to switch from none to enabled mode. Command line cannot be used to set mode to none, can only be used to set to enabled
+2. Add `model_control_mode=enabled` or `model_control_mode=none` to config.properties file
```
namannandan, line 7: To simplify this configuration, please change it to `model_api_enabled` or `model_control_api_enabled` and accept boolean values.

docs/model_control_mode.md
```diff
@@ -0,0 +1,68 @@
...
+
+Priority between cmd and config file follows the following [TorchServer standard](https://github.com/pytorch/serve/blob/c74a29e8144bc12b84196775076b0e8cf3c5a6fc/docs/configuration.md#advanced-configuration)
+* Example 1:
+  * Config file: `model_control_mode=none`
+
+    cmd line: `torchserve --start --ncs --model-store model_store --model-api-enabled`
+
+    Result: Model api mode enabled
```
namannandan, line 15: Describe here the values that were set in the config file and command line and which took precedence to make it clear to the reader.

docs/model_control_mode.md
```diff
@@ -0,0 +1,68 @@
...
+* Example 2:
+  * Config file: `model_control_mode=enabled`
+
+    cmd line: `torchserve --start --ncs --model-store model_store`
+
+    Result: Mode is enabled (no way to enable "none" through cmd)
```
namannandan, line 21: Same as above: Describe here the values that were set in the config file and command line and which took precedence to make it clear to the reader.

test/pytest/test_utils.py
mreso: You can just set the default value to model the old behavior to not break any test.

udaij12: We want TorchServe default to be model api enabled as false meaning that users will not be able to register or delete models once it starts. That change will break the tests

mreso: Udai, this is not about the way people will setup TorchServe, this is only about the interface of start_torchserve which is not user facing.

test/pytest/test_model_control_mode.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+@pytest.fixture(scope="module")
+def setup_torchserve():
+    MODEL_STORE = os.path.join(ROOT_DIR, "model_store/")
+
+    Path(test_utils.MODEL_STORE).mkdir(parents=True, exist_ok=True)
+
+    test_utils.start_torchserve(
+        no_config_snapshots=True, models="mnist=mnist.mar", mode="none"
```
mreso: The interface should be 
```
def start_torchserve(..., model_control_enabled=True):
```
so when you setup a test that want the new behavior you use:
```
test_utils.start_torchserve(..., model_control_enabled=False)
```

mreso (approved):
Addressed my own comments to move on with this. LGTM now

PR #3175, opened 2024-06-03
Changing mar file for Bert torch compile

udaij12:
## Description

Changing mar file for Bert torch compile so that it includes `torch.inference_mode()`


agunapal (approved):
LGTM

Issue #3176, opened 2024-06-05
ImportError: cannot import name 'packaging' from 'pkg_resources'


Sieltek:
### 🐛 Describe the bug

Bug of the import of packaging from setuptools.

### Error logs

2024-06-05T10:24:37,046 [INFO ] W-9000-crochet_yolov8s_512_240207_1.0-stdout MODEL_LOG - Backend worker process died.
2024-06-05T10:24:37,047 [INFO ] W-9000-crochet_yolov8s_512_240207_1.0-stdout MODEL_LOG - Traceback (most recent call last):
2024-06-05T10:24:37,047 [INFO ] W-9000-crochet_yolov8s_512_240207_1.0-stdout MODEL_LOG -   File "/home/torch/.local/lib/python3.11/site-packages/ts/model_loader.py", line 108, in load
2024-06-05T10:24:37,047 [INFO ] W-9000-crochet_yolov8s_512_240207_1.0-stdout MODEL_LOG -     module, function_name = self._load_handler_file(handler)
2024-06-05T10:24:37,047 [INFO ] W-9000-crochet_yolov8s_512_240207_1.0-stdout MODEL_LOG -                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-06-05T10:24:37,047 [INFO ] W-9000-crochet_yolov8s_512_240207_1.0-stdout MODEL_LOG -   File "/home/torch/.local/lib/python3.11/site-packages/ts/model_loader.py", line 153, in _load_handler_file
2024-06-05T10:24:37,047 [INFO ] W-9000-crochet_yolov8s_512_240207_1.0-stdout MODEL_LOG -     module = importlib.import_module(module_name)
2024-06-05T10:24:37,047 [INFO ] W-9000-crochet_yolov8s_512_240207_1.0-stdout MODEL_LOG -              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-06-05T10:24:37,048 [INFO ] W-9000-crochet_yolov8s_512_240207_1.0-stdout MODEL_LOG -   File "/usr/lib64/python3.11/importlib/__init__.py", line 126, in import_module
2024-06-05T10:24:37,048 [INFO ] W-9000-crochet_yolov8s_512_240207_1.0-stdout MODEL_LOG -     return _bootstrap._gcd_import(name[level:], package, level)
[...]

PR #3177, opened 2024-06-05
Fix Inf2 benchmark

namannandan:
## Description
Inf2 benchmarks have been failing due to the following reasons:
1. The latest version of `torch-neuronx` requires torch `2.1.2` and torch was pinned to `1.13`
2. Although `setuptools` is [pinned](https://github.com/pytorch/serve/pull/3152), it was getting upgraded when loading the benchmark models due to installation of custom dependencies ex: `torch-neuronx`. Related issue: https://github.com/pytorch/serve/issues/3176

Fix:
1. Pin torch to a know supported version for neuronx
2. Update Inf2 model artifacts to not use custom dependencies since these are already setup by [install_dependencies.py](https://github.com/pytorch/serve/blob/master/ts_scripts/install_dependencies.py) script before running the benchmark.

## Type of change

- [ ] Passing benchmark run: https://github.com/pytorch/serve/actions/runs/9376610797/job/25816825345

### DIFF
--- a/.github/workflows/benchmark_nightly.yml
+++ b/.github/workflows/benchmark_nightly.yml
@@ -1,14 +1,16 @@
 name: Benchmark torchserve nightly
 
 on:
-  push
+  # run every day at 2:15am
+  schedule:
+    - cron:  '15 02 * * *'
 
 jobs:
   nightly:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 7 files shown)


udaij12:
LGTM

udaij12 (approved):
LGTM

PR #3179, opened 2024-06-06
Fixing torch compile benchmark 

udaij12:
## Description
Torch compile nightly tests are running for 22+ hours and are terminating due to extended use. Root cause is that torchtext installation is causing torch cpu to be installed rather then cu121 which causes the tests to run on the CPU rather then GPU. 

This can be verified through ec2 stats showing 99% CPU usage during the failed tests and `nvidia-smi` showing no running processes during the tests. 

Solution to add `no-deps` flag to the installation of torchtext to prevent it from installing a different version of torch.

## Logs 
https://github.com/pytorch/serve/actions/runs/9395103010/job/25873916833 
Can verify torch version and can also see 
```
GPU models and configuration: 
GPU 0: NVIDIA A10G
GPU 1: NVIDIA A10G
GPU 2: NVIDIA A10G
GPU 3: NVIDIA A10G
```

### DIFF
--- a/ts_scripts/install_dependencies.py
+++ b/ts_scripts/install_dependencies.py
@@ -141,7 +141,7 @@ class Common:
                 f"pip3 install numpy --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/{pt_nightly}"
             )
             os.system(
-                f"pip3 install --pre torchtext --index-url https://download.pytorch.org/whl/nightly/cpu"
+                f"pip3 install --pre torchtext --index-url https://download.pytorch.org/whl/nightly/cpu --no-deps"
             )
         else:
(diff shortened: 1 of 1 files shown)


agunapal (changes requested):
This is the prescribed way to install torchtext

https://github.com/pytorch/serve/pull/3011

As a work around you can skip installing torchtext nightlies.

We will be removing this when 2.4 is released

udaij12:
> #3011

There is no reason to do this anymore as the torchtext in https://download.pytorch.org/whl/nightly/cpu and the torchtext in https://download.pytorch.org/whl/nightly/cu121 link to the same thing.

agunapal (approved):
LGTM

PR #3194, opened 2024-06-14
Fixing docker CI 

udaij12:
## Description
Current Docker CI is using latest TorchServe release rather then the branch that is being merged in. 
Fix by using the `-s ` and `-r` flag that will build TorchServe from the source and use the local branch. 

Also adding '--disable-token` flag for the docker tests to run properly.


agunapal (approved):
LGTM

PR #3197, opened 2024-06-20
Benchmark numpy fix

udaij12:
## Description
Numpy release of 2.0 has caused issues in the benchmark script. 
Solution is to add the same numpy pin in the benchmark requirements file

## Testing 
https://github.com/pytorch/serve/actions/runs/9592493747/job/26451152735


agunapal (approved):
LGTM

Issue #3044, opened 2024-03-26
Kserve nightly CPU tests failure


namannandan:
### 🐛 Describe the bug

Kserve nightly CPU test failure: https://github.com/pytorch/serve/actions/runs/8431231162/job/23088317085

### Error logs

Likely mismatch between supported `kubectl` and `kserve` versions:
```
Install Kserve
error: unknown flag: --short
See 'kubectl version --help' for usage.
./hack/quick_install.sh: line 41: [: -lt: unary operator expected
```

Import error
```
Traceback (most recent call last):
  File "/home/model-server/kserve_wrapper/__main__.py", line 6, in <module>
    import kserve
  File "/home/venv/lib/python3.9/site-packages/kserve/__init__.py", line 18, in <module>
    from .model_server import ModelServer
  File "/home/venv/lib/python3.9/site-packages/kserve/model_server.py", line 27, in <module>
    from ray.serve.handle import RayServeHandle
ImportError: cannot import name 'RayServeHandle' from 'ray.serve.handle' (/home/venv/lib/python3.9/site-packages/ray/serve/handle.py)
```

Inference service failure:
```
 Warning  Unhealthy  1s (x3 over 3s)  kubelet            Readiness probe failed: Get "http://10.244.0.17:8012/": context deadline exceeded (Client.Timeout exceeded while awaiting headers)

Wait for inference service to be ready
Timeout waiting for inference service to come up.
Delete cluster
* Deleting "minikube" in docker ...
* Deleting container "minikube" ...
* Removing /home/ubuntu/.minikube/machines/minikube ...
* Removed all traces of the "minikube" cluster.
Error: Process completed with exit code 1
```

### Installation instructions

Workflow: https://github.com/pytorch/serve/actions/runs/8431231162/workflow

### Model Packaing

Workflow: https://github.com/pytorch/serve/actions/runs/8431231162/workflow

### Versions

master branch

### Repro instructions

Workflow: https://github.com/pytorch/serve/actions/runs/8431231162/workflow

agunapal:
https://github.com/kserve/kserve/issues/3541

PR #3205, opened 2024-06-25
Kserve fix ray & setuptools dependency issue

udaij12:
## Description
Current pinned version of ray is causing dependency issues with newer version of setuptools. Solution is to unpin ray and test to see if the newer release of ray will work.


agunapal (approved):
approving to see if this resolves the problem

udaij12:
This change fixed the dependency issue but https://github.com/pytorch/serve/issues/3044 has shown up

PR #3209, opened 2024-06-26
Add single command LLM deployment

mreso:
## Description

This PR adds a feature to TorchServe to deploy LLM with a single command.
Is uses a new ts.launcher interface to start and stop torchserve and adds ts.llm_launcher to launch an LLM model given its hugginface hub identifier using our new vllm integration.
It also adds a docker image based on our gpu image to easily run this without installing torchserve.

## Type of change

- [X] Test A
```
docker run --rm -ti --gpus all -e HUGGING_FACE_HUB_TOKEN=$token -p 8080:8080 -v data:/data ts/llm --model_id meta-llama/Meta-Llama-3-8B-Instruct --disable_token
curl -X POST -d '{"prompt":"Hello, my name is", "max_new_tokens": 50}' --header "Content-Type: application/json" "http://localhost:8080/predictions/model"
```
Logs for Test A
```
{"text": " [", "tokens": 510}{"text": "Your", "tokens": 7927}{"text": " Name", "tokens": 4076}{"text": "].", "tokens": 948}{"text": " I", "tokens": 358}{"text": " am", "tokens": 1097}{"text": " a", "tokens": 264}{"text": " [", "tokens": 510}{"text": "Your", "tokens": 7927}{"text": " Profession", "tokens": 50311}{"text": "/", "tokens": 14}{"text": "Student", "tokens": 14428}{"text": "]", "tokens": 60}{"text": " and", "tokens": 323}{"text": " I", "tokens": 358}{"text": " am", "tokens": 1097}
```

## Checklist:

- [X] Did you have fun?


docker/Dockerfile.llm
```diff
@@ -0,0 +1,9 @@
+FROM pytorch/torchserve-nightly:latest-gpu as server
+
+USER root
+
+RUN mkdir /data && chown -R model-server /data
+
+USER model-server
+
+ENTRYPOINT [ "python", "-m", "ts.llm_launcher", "--vllm_engine.download_dir", "/data" ]
```
agunapal, line 9: Can we make this generic. What if we use a script and pass args . Ex: What do we do if we want to reuse this for TRT-LLM?

ts/llm_launcher.py
```diff
@@ -0,0 +1,162 @@
...
+        "maxWorkers": 1,
+        "batchSize": 1,
+        "maxBatchDelay": 100,
+        "responseTimeout": 1200,
+        "deviceType": "gpu",
+        "asyncCommunication": True,
+        "parallelLevel": torch.cuda.device_count() if torch.cuda.is_available else 1,
+        "handler": {
+            "model_path": args.model_id,
+            "vllm_engine_config": {
```
agunapal, line 32: Same as above, Can we make this generic to support other engines?

frontend/server/src/main/java/org/pytorch/serve/wlm/AsyncWorkerThread.java
```diff
@@ -75,6 +77,14 @@ public void run() {
                 try {
                     backendChannel.get(0).writeAndFlush(req).sync();
                     logger.debug("Successfully flushed req");
+
+                    if (loadingFinished == false) {
```
agunapal, line 81: Is this specific to async workers only?

docker/Dockerfile.llm
```diff
@@ -0,0 +1,9 @@
+FROM pytorch/torchserve-nightly:latest-gpu as server
+
+USER root
+
+RUN mkdir /data && chown -R model-server /data
+
+USER model-server
+
```
agunapal, line 8: How do we pass any optional 3rd party dependencies/scripts/commands to execute?

docker/Dockerfile.llm
mreso: plan is to use fire instead of arparse for the cli interface which will allow for arbitrary arguments and we can then select the engine and parse e.g. --vllm_engine.download_dir and --trt_llm.foo_bar based on the type.
Current interface will not be affected so we can add this in follow up prs.

ts/llm_launcher.py
mreso: see above, when we can add an --engine command in follow up pr with default vllm we can add support for other engines without changing the current interface

frontend/server/src/main/java/org/pytorch/serve/wlm/AsyncWorkerThread.java
mreso: Yes, sync worker is waiting for model LOAD command to complete but async worker does not.

docker/Dockerfile.llm
mreso: Not sure if I follow. What kind of dependencies? The quickstart launcher uses vllm which brings everything we need. What do you have in mind?

docker/Dockerfile.llm
agunapal: Sounds good!

ts/llm_launcher.py
agunapal: Cool. That makes sense

docker/Dockerfile.llm
agunapal: I am thinking beyond a basic deployment. Ex: If I want to use this with langchain and want to pip install langchain and derive a custom handler from the VLLM handler and add the langchain specific code in there, how do I do this. Or may be this can just in the guide as instructions?

mreso: Got it, not sure if we can integrate this into the quickstart/llm_launcher. But users can refer to examples/large_models/vllm for this to customize the handler. Will add a reference to that.

mreso: added

docs/llm_deployment.md
```diff
@@ -0,0 +1,73 @@
...
+
+Models are usually loaded from the HuggingFace hub and are cached in a [docker volume](https://docs.docker.com/storage/volumes/) for faster reload.
+If you want to access gated models like the Meta-Llama3 model you need to provide a HuggingFace hub token:
+```bash
+export token=<HUGGINGFACE_HUB_TOKEN>
+```
+
+You can then go ahead and launch a TorchServe instance serving your selected model:
+```bash
+docker run --rm -ti --gpus all -e HUGGING_FACE_HUB_TOKEN=$token -p 8080:8080 -v data:/data ts/llm --model_id meta-llama/Meta-Llama-3-8B-Instruct --disable_token
```
agunapal, line 25: Is this needed `-e HUGGING_FACE_HUB_TOKEN=$token` Why not directly set `export HUGGING_FACE_HUB_TOKEN= <HUGGINGFACE_HUB_TOKEN>` 

Also from the security POV,  does the existing command print the token?

agunapal (approved):
Looks good! 
Had one comment on the token. Please check and change if you think it makes sense

docs/llm_deployment.md
mreso: AFAIK the docker will not pick up env variables from the calling environment. So you would still have 
```
export HUGGING_FACE_HUB_TOKEN= <HUGGINGFACE_HUB_TOKEN>
docker run --rm -ti --gpus all -e HUGGING_FACE_HUB_TOKEN=$HUGGING_FACE_HUB_TOKEN ...
```
which is even longer and it comes down to the same process. For a RL deployment the token variable would be set through a secret and so the token would not show up in any of the logs.

PR #3211, opened 2024-06-27
Update documentation for vgg16 to use torch.compile

ijkilchenko:
Update image_classifier/vgg-16 to include torch.compile

Add model config file
Update README

## Type of change

## Testing

Verified output for a sample inference call makes sense before and after using a model config file


mreso (approved):
LGTM, thanks for the contribution!

PR #3213, opened 2024-06-27
Copy remote branch entrypoint to compile and production image stages

lanxih:
## Description

Torchserve docker build with remote branch needs to copy the remote version of entrypoint (`dockerd-entrypoint.sh`) to the image to keep it consistent with the `--branch_name` specified for the build.

Fixes #(issue)
We encountered this issue where the torchserve image is expected from --branch_name v0.11.0 in order to work with pinned torchserve pypi package version, but the current entrypoint is out-of-sync and uses additional flag `--disable-token` from master.

## Type of change

Re-run `bash ./build_image.sh --branch_name v0.11.0 --pythonversion 3.10 --tag test-image` to verify that the newly built image now has `/usr/local/bin/dockerd-entrypoint.sh` from branch v0.11.0 without `--disable-token` set. 
```
% docker run --rm -it --entrypoint bash test-image
model-server@465e40c1ead4:~$ cat /usr/local/bin/dockerd-entrypoint.sh
#!/bin/bash
set -e

if [[ "$1" = "serve" ]]; then
    shift 1
    torchserve --start --ts-config /home/model-server/config.properties
else
    eval "$@"
fi

# prevent docker exit
tail -f /dev/null
```
Also ran without branch name to verify the entrypoint has the new flag.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -78,6 +78,8 @@ RUN \
 
 WORKDIR "serve"
 
+RUN cp docker/dockerd-entrypoint.sh /usr/local/bin/dockerd-entrypoint.sh
+
 RUN \
     if echo "$BASE_IMAGE" | grep -q "cuda:"; then \
         # Install CUDA version specific binary when CUDA version is specified as a build arg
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -78,6 +78,8 @@ RUN \
 
 WORKDIR "serve"
 
+RUN cp docker/dockerd-entrypoint.sh /usr/local/bin/dockerd-entrypoint.sh
```
agunapal, line 81: Can you please change this to `COPY` instead of `RUN cp`

agunapal (changes requested):

docker/Dockerfile
lanxih: great question, I also tried `COPY` but seems that would copy dockerd-entrypoint.sh from the volume (which is the older version prior to cloning from specific branch) to the image, so I suppose that does not work...

agunapal: oh..Interesting

agunapal (approved):
LGTM

PR #3224, opened 2024-07-03
Updating examples for security tags

udaij12:
## Description
adding `--disable-token-auth` and `--enable-model-api` to examples


agunapal:
Hi @udaij12 I think we should be adding `--enable-model-api` only when we are registering models after TorchServe has started. Otherwise , it would confuse the users.

agunapal (approved):
LGTM

PR #3228, opened 2024-07-03
TensorRT-LLM Engine integration

agunapal:
## Description

Fixes #(issue)

## Type of change

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

## Checklist:

- [ ] Did you have fun?

### DIFF
--- a/examples/large_models/trt_llm/llama/README.md
+++ b/examples/large_models/trt_llm/llama/README.md
@@ -1,0 +1,84 @@
+# Llama TensorRT-LLM Engine integration with TorchServe
+
+[TensorRT-LLM](https://github.com/NVIDIA/TensorRT-LLM) provides users with an option to build TensorRT engines for LLMs that contain state-of-the-art optimizations to perform inference efficiently on NVIDIA GPUs.
+
+## Pre-requisites
+
+TRT-LLM requires Python 3.10
+Once TorchServe is installed, install TensorRT-LLM using the following.
+This will downgrade the versions of PyTorch & Triton but this doesn't cause any issue.
+```
+pip install tensorrt_llm  --extra-index-url https://pypi.nvidia.com
+python -c "import tensorrt_llm"
+```
+shows
+```
+[TensorRT-LLM] TensorRT-LLM version: 0.10.0
+```
+
+## Download model from HuggingFace
+```
+huggingface-cli login
+# or using an environment variable
+huggingface-cli login --token $HUGGINGFACE_TOKEN
+```
+```
+python ../../utils/Download_model.py --model_path model --model_name meta-llama/Meta-Llama-3-8B-Instruct
+```
+
+## Create TensorRT-LLM Engine
+Clone TensorRT-LLM which will be used to create the TensorRT-LLM Engine
+
+```
+git clone -b v0.10.0 https://github.com/NVIDIA/TensorRT-LLM.git
+```
+
+Compile the model into a TensorRT engine with model weights and a model definition written in the TensorRT-LLM Python API.
+
+```
+python TensorRT-LLM/examples/llama/convert_checkpoint.py --model_dir model/models--meta-llama--Meta-Llama-3-8B-Instruct/snapshots/e1945c40cd546c78e41f1151f4db032b271faeaa/ --output_dir ./tllm_checkpoint_1gpu_bf16 --dtype bfloat16
+```
+```
+trtllm-build --checkpoint_dir tllm_checkpoint_1gpu_bf16 --gemm_plugin bfloat16 --gpt_attention_plugin bfloat16 --output_dir ./llama-3-8b-engine
+```
+
+You can test if TensortRT-LLM Engine has been compiled correctly by running the following
+```
+python TensorRT-LLM/examples/run.py --engine_dir ./llama-3-8b-engine  --max_output_len 100 --tokenizer_dir model/models--meta-llama--Meta-Llama-3-8B-Instruct/snapshots/e1945c40cd546c78e41f1151f4db032b271faeaa/ --input_text "How do I count to nine in French?"
+```
+
+You should see an output as follows
+```
+Input [Text 0]: "<|begin_of_text|>How do I count to nine in French?"
+Output [Text 0 Beam 0]: " Counting to nine in French is easy and fun. Here's how you can do it:
+One: Un
+Two: Deux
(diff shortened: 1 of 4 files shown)


examples/large_models/trt_llm/llama/README.md
```diff
@@ -0,0 +1,85 @@
...
+```
+mkdir model_store
+torch-model-archiver --model-name llama3-8b --version 1.0 --handler trt_llm_handler.py --config-file model-config.yaml --archive-format no-archive --export-path model_store -f
+mv model model_store/llama3-8b/.
+mv llama-3-8b-engine model_store/llama3-8b/.
+```
+
+## Start TorchServe
+```
+torchserve --start --ncs --model-store model_store --models llama3-8b --disable-token
```
mreso, line 78: Needs to updated to --disable-token-auth after #3224

examples/large_models/trt_llm/llama/README.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+## Run Inference
+```
+pip install orjson
```
mreso, line 83: We should either kill orjson in that utility or make it a ts dependency. I favor option 1 as we don't really benefit from its speed here.

examples/large_models/trt_llm/llama/model-config.yaml
```diff
@@ -0,0 +1,12 @@
+# TorchServe frontend parameters
+minWorkers: 1
+maxWorkers: 1
+maxBatchDelay: 100
+responseTimeout: 1200
+deviceType: "gpu"
+asyncCommunication: true
```
mreso, line 7: Can TRT-LLM handle multi-gpu inference easily? Then we should demonstrate that we can easily integrate that with 
```
parallelType: "custom"
parallelLevel: 4
```

examples/large_models/trt_llm/llama/trt_llm_handler.py
```diff
@@ -0,0 +1,125 @@
...
+                batch_input_ids=input_ids_batch,
+                temperature=temperature,
+                max_new_tokens=max_new_tokens,
+                end_id=self.tokenizer.eos_token_id,
+                pad_id=self.tokenizer.pad_token_id,
+                output_sequence_lengths=True,
+                streaming=True,
+                return_dict=True,
+            )
+        torch.cuda.synchronize()
```
mreso, line 99: What is the synchronization for?

examples/large_models/trt_llm/llama/trt_llm_handler.py
```diff
@@ -0,0 +1,125 @@
...
+            output_ids = inference_output["output_ids"]
+            sequence_lengths = inference_output["sequence_lengths"]
+
+            batch_size, num_beams, _ = output_ids.size()
+            input_lengths = [x.size(0) for x in input_batch]
+            for batch_idx in range(batch_size):
+                for beam in range(num_beams):
+                    output_begin = input_lengths[batch_idx]
+                    output_end = sequence_lengths[batch_idx][beam]
+                    outputs = output_ids[batch_idx][beam][
```
mreso, line 114: Are we sure we're doing the right thing here? Because output_begin is never used but instead `output_end-1`.

examples/large_models/trt_llm/llama/trt_llm_handler.py
```diff
@@ -0,0 +1,125 @@
...
+                        output_end - 1 : output_end
+                    ].tolist()
+                    output_text = self.tokenizer.decode(outputs)
+                    send_intermediate_predict_response(
```
mreso, line 118: I we send N=num_beams intermediate results back without any order information, can we assign each partial response to its beam sequence? Better to send one response per batch entry (will will be =1) with updates for all beams included as a list?

mreso (changes requested):
Solid overall, left some comments. Especially the postprocessing needs some rework I think as client side does not now how many beams to expect and so it will be hard to make sense of the returned streaming chunks.

examples/large_models/trt_llm/llama/trt_llm_handler.py
agunapal: good catch, we don't need output_begin. The example code uses this

examples/large_models/trt_llm/llama/trt_llm_handler.py
agunapal: it seems like this is not needed for llama as num_beams > 1 is not working for llama Removed the inner for loop

examples/large_models/trt_llm/llama/trt_llm_handler.py
agunapal: Copy paste from the example code , but seems like its not needed. Works fine without it

examples/large_models/trt_llm/llama/model-config.yaml
agunapal: I tried standalone multi gpu inference, it didn't work for me. Although the model loaded on the 4 GPUs, the inference was hanging

examples/large_models/trt_llm/llama/README.md
agunapal: agreed. I removed orjson

examples/large_models/trt_llm/llama/README.md
agunapal: done. Thanks. But seems like there is a bug which is accepting the prefix.

examples/large_models/trt_llm/llama/README.md
```diff
@@ -10,6 +10,7 @@ This will downgrade the versions of PyTorch & Triton but this doesn't cause any
 
 ```
 pip install tensorrt_llm==0.10.0 --extra-index-url https://pypi.nvidia.com
+pip install tensorrt-cu12==10.1.0
```
mreso, line 13: Is this CUDA 12 exclusive? In that case we should inform people to install torch with CUDA 12 as well.

mreso (approved):
Lets figure out the CUDA 12 situation, then this LGTM

examples/large_models/trt_llm/llama/README.md
agunapal: It doesn't mention it but all their docs are pointing to CUDA 12.x . Let me mention that its tested with CUDA 12.1

PR #3232, opened 2024-07-09
System metrics command environment variable

namannandan:
## Description
Enable configuring system metrics command via environment variable


agunapal (commented):
@namannandan Can you please add in the PR, how this would work with an example

lxning:
system metrics cmd is better to be set in config.property, not in env var. For Meta Fair demo, the session stats collected via system metrics is a short term solution. In the long term, it should be

lxning:
it might trigger security concern b/c user can input any command via env. That's why the original code only allows this setting via config.property.

For Meta Fair, the customized container is required. So I don't think setting via env is necessary.

PR #3233, opened 2024-07-09
Updating docker cuda and github branch

udaij12:
## Description
Updating cuda to 121 and updating github branch extraction to now also be able to get all branches rather than just when trying to merge.


agunapal (approved):
LGTM

PR #3230, opened 2024-07-05
Fix/llm launcher disable token

mreso:
## Description

This PR fixes the llm deployment after changes in disable-token-auth parameter

Fixes #(issue)
#3229 

## Type of change

Please delete options that are not relevant.

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] This change requires a documentation update

## Feature/Issue validation/testing

Please describe the Unit or Integration tests that you ran to verify your changes and relevant result summary. Provide instructions so it can be reproduced.
Please also list any relevant details for your test configuration.

- [X] Test A
```
docker build . -f docker/Dockerfile.llm -t ts/llm
docker run --rm -ti --gpus all --shm-size 1g -e HUGGING_FACE_HUB_TOKEN=$token -p 8080:8080 -v data:/data ts/llm --model_id meta-llama/Meta-Llama-3-8B-Instruct --disable_token_auth
```
Logs for Test A
```
...
2024-07-05T14:32:43,689 [INFO ] W-9000-model_1.0-stdout MODEL_LOG - INFO 07-05 14:32:43 model_runner.py:965] Graph capturing finished in 2 secs.
2024-07-05T14:32:43,784 [INFO ] epollEventLoopGroup-5-1 org.pytorch.serve.wlm.AsyncBatchAggregator - Predictions is empty. This is from initial load....
2024-07-05T14:32:43,785 [INFO ] epollEventLoopGroup-5-1 org.pytorch.serve.wlm.AsyncWorkerThread - Worker loaded the model successfully
2024-07-05T14:32:43,785 [DEBUG] epollEventLoopGroup-5-1 org.pytorch.serve.wlm.WorkerThread - W-9000-model_1.0 State change WORKER_STARTED -> WORKER_MODEL_LOADED
2024-07-05T14:32:43,785 [INFO ] epollEventLoopGroup-5-1 TS_METRICS - WorkerLoadTime.Milliseconds:19486.0|#WorkerName:W-9000-model_1.0,Level:Host|#hostname:16c99e392b9c,timestamp:1720189963
2024-07-05T14:32:43,785 [INFO ] W-9000-model_1.0 org.pytorch.serve.wlm.AsyncBatchAggregator - Getting requests from model: org.pytorch.serve.wlm.Model@4115ab4f
```
Run
```
curl -X POST -d '{"prompt":"Hello, my name is", "max_new_tokens": 50, "temperature": 0}' --header "Content-Type: application/json" "http://localhost:8080/predictions/model"
```
Log
```
[...]

PR #3242, opened 2024-07-15
HPU requirements config file

RafLit:
## Description

In #3230 vllm was added to requirements/common.txt. 
Installation of vllm automatically installs an exact version of pytorch and overwrites our version.
We propose to add a parameter to install dependencies from a separate configuration for hpu.
Alternatively, moving vllm away from the common requirements file should also be considered.

## Type of change

- requirements/hpu.txt with gaudi dependencies was added,
- `--hpu` flag was added to ts_scripts/install_dependencies.py instead of `--skip_torch_install` as torch was eventually installed by vllm anyway.

## Checklist:

- [x] Did you have fun?


RafLit:
@agunapal could you please review this?

agunapal:
> @agunapal could you please review this?

Hi @RafLit We'll have to discuss internally where to place vllm. We'll have a similar challenge with TensorRT(trt-llm)

But I like this idea of having a separate requirements file, at least for different GPU HWs. But the question that would still remain is..who will update this file? We get alerts when new versions of these libraries are released. Would you be updating them?.

agunapal:
Hi @RafLit Closing this PR since we removed the dependency on vllm. Please check and confirm

RafLit:
@agunapal Hi, we might be interested in the future in having a separate file for dependencies for our hardware and we could update it. For now this can remain closed, Thank you!

Issue #3261, opened 2024-07-20
Startup timeout should be configurable separately


Isalia20:
### 🚀 The feature

Right now to increase startup timeout of the model, you have to increase request timeout, it would be nice if it could be configurable separately so you don't have to increase the response timeout

### Motivation, pitch

There are scenarios where startup time takes considerably more amount of time than the usual requests. One example is when you are torch compiling the model at the start of the torchserve and run a "warmup" forward pass so it actually compiles the model and the real request isn't slowed down by compilation

### Alternatives

Only alternative now(as far as I'm aware) is to increase response timeout, however increasing this also means increasing regular response timeout which isn't a desirable behaviour, startup might take for example 3 minutes but real inference might be under 1 minute.

### Additional context

I would like to contribute this feature if it's possible. It would be my first contribution to torchserve and I already have a minimum working example, would be glad to hear your thoughts on this

agunapal:
Thanks @Isalia20 This makes sense. Sure, if you have a working solution, please send the PR and we can discuss.

Isalia20:
Thanks for the reply, I've added the PR. Have some questions which I have added in the PR

Isalia20:
It's ready for a review

agunapal:
Thanks @Isalia20 We will try to review this next week

mreso:
Got merged with #3262

Issue #2452, opened 2023-07-08
how does `default_response_timeout` work?


sreeprasannar:
### 📚 The doc issue

I set the value of `default_response_timeout` to 4 i.e. 4 seconds. At the start of the model load, this happens after 4 (ish) seconds:
```
org.pytorch.serve.wlm.WorkerInitializationException: Backend worker did not respond in given time
```

My guess is because the model takes a while to load (more than 4 seconds), the worker gets killed.
Is there a way to set a larger initial delay i.e. differentiate these two scenarios:
* account for the initial model load with a number different from `default_response_timeout`
* if model doesn't response in `default_response_timeout` after the initial load, then kill the worker

sreeprasannar:
I know I can set a higher `default_response_timeout` but that doesn't solve my problem. I want my model to go into `UNLOADING` state so that my kubernetes cluster can restart the pod. I want this to happen asap not after 120seconds hence the need to reduce the `default_response_timeout`

sreeprasannar:
and a follow-up to this: is there a way to reduce the number of auto-recovery tries? Considering torchserve isn't really good at autorecovery anyway, I'd rather let kubernetes restart the pod

arnavmehta7:
Agreed, yes, auto-recovery just stucks for me. Whereas restarting pod will be better.

lxning:
@sreeprasannar @arnavmehta7 v0.8.0 allows for model level config (see [model config yaml](https://github.com/pytorch/serve/tree/master/model-archiver#config-file)). 
- increase `responseTimeout` in model config yaml file to handle long model loading latency.
- add `maxRetryTimeoutInSec` in model config yaml file to make ping return 500 and trigger pod restart. (see quota from v0.8.0 release notes => New Features => 10. Updated ping endpoint default behavior)

arnavmehta7:
@lxning Does the responseTimeout consider the waiting time in queue? i.e., if my responseTimeout is 100 seconds, and a request is in queue for last 101 seconds, will it get cancelled?

arnavmehta7:
Waiting time ns: 992559626073, Backend time ns: 14687643525
 I am experiencing enormous Waiting times like 900seconds in this example. But quite less backend time :(

PR #3262, opened 2024-07-22
Configurable startup time

Isalia20:
## Description

Adds feature from  #3261  and from #2452

## Type of change

- [x] Test for long startup while having small response timeout. Test can be found in test/pytest/test_startup_timeout.py

================================================================================ test session starts =================================================================================
platform linux -- Python 3.10.14, pytest-7.3.1, pluggy-1.0.0
rootdir: /home/isalia/Desktop/torchserve
plugins: mock-3.14.0, timeout-2.3.1, cov-4.1.0, typeguard-4.3.0
collected 1 item                                                                                                                                                                     

test/pytest/test_startup_timeout.py .                                                                                                                                          [100%]

================================================================================= 1 passed in 15.02s =================================================================================

## Checklist:

- [x] Did you have fun?


test/pytest/test_startup_timeout.py
```diff
@@ -0,0 +1,116 @@
...
+            if model_status == "READY":
+                break
+        time.sleep(1)
+
+    end_time = time.time()
+    elapsed_time = end_time - start_time
+
+    assert response.status_code == 200, "Model startup failed"
+    assert model_status == "READY", f"Unexpected model status: {model_status}"
+    assert 10 <= elapsed_time < 20, f"Unexpected startup time: {elapsed_time} seconds"
```
mreso, line 116: Currently, this test only tests the successful side of registering a model. Lets also test the situation where we run into the timeout and the worker gets killed. We can just configure a 5 seconds timeout and let the init sleep 10 as its now. We can parameterize the test for this. Then we can check in the stdout of TorchServe for the right message. See [here](https://github.com/pytorch/serve/blob/3f401806742a9430f85aabba7f2fa90a7be8c56d/test/pytest/test_continuous_batching.py#L92) how to get torchserve output in a test (The tests actually ignored it but the torchserve fixture is actually a Queue that you can .get() log messaged from). Let me know if you need help with this.

mreso (changes requested):
Thanks for the contribution. Please see my comment regarding testing. Otherwise this look good to go after adjusting the test.

frontend/server/src/main/java/org/pytorch/serve/wlm/WorkerThread.java
```diff
@@ -292,8 +297,8 @@ public void run() {
                 logger.debug("Shutting down the thread .. Scaling down.");
             } else {
                 logger.debug(
-                        "Backend worker monitoring thread interrupted or backend worker process died., responseTimeout:"
-                                + responseTimeout
+                        "Backend worker monitoring thread interrupted or backend worker process died., startupTimeout:"
```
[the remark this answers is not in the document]
Isalia20: makes sense, added the conditional

test/pytest/test_startup_timeout.py
```diff
@@ -0,0 +1,116 @@
...
+            model_status = response.json()[0]["workers"][0]["status"]
+            if model_status == "READY":
+                break
+        time.sleep(1)
+
+    end_time = time.time()
+    elapsed_time = end_time - start_time
+
+    assert response.status_code == 200, "Model startup failed"
+    assert model_status == "READY", f"Unexpected model status: {model_status}"
```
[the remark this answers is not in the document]
Isalia20: oops, missed that. Added the default value for it

test/pytest/test_startup_timeout.py
Isalia20: I've added the tests for it but not sure if it's the best way to do it. It reads the value from torchserve stdout and expects an error message to pop up and passes the test in that case.

mreso: Great! That will do, thanks!

ts_scripts/spellcheck_conf/wordlist.txt
```diff
@@ -218,6 +218,7 @@ preflight
 readthedocs
 req
 responseTimeout
+startupTimeout
```
mreso, line 221:
```suggestion
startupTimeout
timeframe
```

Isalia20:
I see some checks are failing but I'm not sure what the issue really is, to fix it

mreso:
> I see some checks are failing but I'm not sure what the issue really is, to fix it

Yeah, these regression tests can be a pain. Taking a look

Isalia20:
One of the tests was failing due to startup_timeout not being cast to int in `test/pytest/test_gRPC_management_apis.py`. Added it to the list of params that should be casted to int

Isalia20:
Seems like tests have passed(required ones). Docker ones are failing due to:
0.811 fatal: Remote branch configurable-startup-time not found in upstream origin
which could be because it can't clone from my branch 🤔

mreso:
Hi @Isalia20 yes, that should be the docker build script which only supports pytorch/serve as remote. We can proceed with merging this PR. Thanks again for your contribution!

mreso (approved):
LGTM now

PR #3281, opened 2024-08-05
Added cu125 image to support Ubuntu 24.04

### LINKED ISSUE #3278: Support for a CUDA 12.5.x
### 🚀 The feature

I pushed the container and saw over 500 CVEs when I scanned the container. Many, if not all, would be addressed if we support Ubuntu 24.04. NVIDA has a docker image running 24.04 as it's base and it runs 12.5.1. I would like to update the build scripts to support this imae

### Motivation, pitch

500+ CVEs related to the base image using 20.04

### Alternatives

_No response_

### Additional context

_No response_

jasonsmithio:
## Description

I pushed the existing container and saw over 500 CVEs when I scanned the container. Many, if not all, would be addressed if we support Ubuntu 24.04. NVIDA has a docker image running 24.04 as it's base and it runs 12.5.1. I went ahead and added build support for it. As far as I can tell, nothing is broken in this version. 

Fixes #3278

## Type of change

- [ x] New feature (non-breaking change which adds functionality)
- [x ] This change requires a documentation update (maybe)

No formal unit test. I built the container without error. This one has only 116 vulnerabilities vs 596 in previous version. 2 CVEs are high and he remaining are Medium (97) and Low (17). 

Used the container and didn't notice any issues differing from the previous version.  I am open to doing more tests if need be. 

## Checklist:

- [x ] Did you have fun?
- [x ] Have you added tests that prove your fix is effective or that this feature works?

### DIFF
--- a/model-archiver/model_archiver/model_packaging_utils.py
+++ b/model-archiver/model_archiver/model_packaging_utils.py
@@ -43,6 +43,8 @@ MODEL_ARCHIVE_VERSION = "1.0"
 MANIFEST_FILE_NAME = "MANIFEST.json"
 MAR_INF = "MAR-INF"
 
+logger = logging.getLogger(__file__)
+
 
 class ModelExportUtils(object):
     """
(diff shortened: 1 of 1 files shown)


mreso:
@ankithagunapal do you recall what kind of issues you were facing when updating to 24.04? Otherwise it would be great to move forward with this.

agunapal:
@mreso CI / regressions with ubuntu 22 were flaky. Hence we didn't upgrade. Haven't tried that in a while.

jasonsmithio:
What exactly were the regression issues? I tested and didn't notice anything out of the ordinary when I deployed this on Ubuntu 24.04

agunapal:
> What exactly were the regression issues? I tested and didn't notice anything out of the ordinary when I deployed this on Ubuntu 24.04

@jasonsmithio The workflows you see in this file, you can try updating them to ubuntu 22 or 24 (atleast the CPU ones) . For CUDA versions, we follow PyTorch, So first step would be to support 12.4 , Python version too, we could try updating to 3.10 first

PR #3306, opened 2024-09-09
Example to demonstrate building a custom endpoint plugin

namannandan:
## Description
Add an example to demonstrate building a custom endpoint plugin which reports model ready status to serve inference requests.

## Type of change

- [ ] Readme includes the steps and the outputs seen as a result

### DIFF
--- a/examples/custom_endpoint_plugin/ModelReady.java
+++ b/examples/custom_endpoint_plugin/ModelReady.java
@@ -1,0 +1,54 @@
+package org.pytorch.serve.plugins.endpoint;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
+import org.pytorch.serve.servingsdk.Context;
+import org.pytorch.serve.servingsdk.Model;
+import org.pytorch.serve.servingsdk.ModelServerEndpoint;
+import org.pytorch.serve.servingsdk.Worker;
+import org.pytorch.serve.servingsdk.annotations.Endpoint;
+import org.pytorch.serve.servingsdk.annotations.helpers.EndpointTypes;
+import org.pytorch.serve.servingsdk.http.Request;
+import org.pytorch.serve.servingsdk.http.Response;
+
+@Endpoint(
+        urlPattern = "model-ready",
+        endpointType = EndpointTypes.INFERENCE,
+        description = "Endpoint indicating registered model/s ready to serve inference requests")
+public class ModelReady extends ModelServerEndpoint {
+    private boolean modelsLoaded(Context ctx) {
+        Map<String, Model> modelMap = ctx.getModels();
+
+        for (Map.Entry<String, Model> entry : modelMap.entrySet()) {
+            boolean workerReady = false;
+            for (Worker w : entry.getValue().getModelWorkers()) {
+                if (w.isRunning()) {
+                    workerReady = true;
+                    break;
+                }
+            }
+            if (!workerReady) {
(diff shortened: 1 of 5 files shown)


examples/custom_endpoint_plugin/README.md
```diff
@@ -0,0 +1,112 @@
...
+  ```
+
+  ```bash
+  $ curl -X GET http://localhost:8080/model-ready
+  {
+    "Status": "Model/s ready"
+  }
+  ```
+
+  The `model-ready` endpoint reports that the model is now ready since there is atleast one worker that has loaded the model and is ready to serve inference requests.
```
agunapal, line 105: Can you please update the example to have 2 mar files. The API should return success when all the models have atleast one worker

agunapal (changes requested):

examples/custom_endpoint_plugin/README.md
namannandan: Updated the example to include two models.

examples/custom_endpoint_plugin/README.md
```diff
@@ -0,0 +1,145 @@
...
+  ```bash
+  $ python ts_scripts/install_dependencies.py --environment=dev
+  ```
+
+- Step 2: Copy [ModelReady.java](ModelReady.java) to the endpoint plugins directory
+
+  ```bash
+  $ cp examples/custom_endpoint_plugin/ModelReady.java plugins/endpoints/src/main/java/org/pytorch/serve/plugins/endpoint
+  ```
+  Review the utilization of the [TorchServe SDK API](https://github.com/pytorch/serve/tree/master/serving-sdk) in [ModelReady.java](ModelReady.java) to implement the necessary functionality for the HTTP API endpoint.
```
agunapal, line 24: Its not clear what this is for?

examples/custom_endpoint_plugin/ModelReady.java
```diff
@@ -0,0 +1,54 @@
...
+                if (w.isRunning()) {
+                    workerReady = true;
+                    break;
+                }
+            }
+            if (!workerReady) {
+                return false;
+            }
+        }
+        return true;
```
agunapal, line 35: if no model is loaded, should it return 200?

agunapal: This should be false?

namannandan: > if no model is loaded, should it return 200?

Makes sense, we can have a separate check to see if no models have been registered and return `false` in that case.

namannandan: > This should be false?

This should still be `true` because if every registered model has atleast on worker that has loaded the model, we should return `true`.

examples/custom_endpoint_plugin/README.md
namannandan: We can make this more specific and point it to the serving SDK interfaces here: https://github.com/pytorch/serve/tree/master/serving-sdk/src/main/java/org/pytorch/serve/servingsdk and also link to the corresponding implementation: https://github.com/pytorch/serve/tree/master/frontend/server/src/main/java/org/pytorch/serve/servingsdk/impl. As to why this is required, it is useful when writing a custom plugin, see the implementation in the plugin file: examples/custom_endpoint_plugin/ModelReady.java

agunapal (approved):
LGTM Thanks!

PR #3328, opened 2024-09-20
fix for snapshot variables missing/null

udaij12:
## Description

Adding check for model variables to set to default if missing when using snapshot


frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
```diff
@@ -190,12 +190,35 @@ public JsonObject getModelState(boolean isDefaultVersion) {
...
-        minWorkers = modelInfo.get(MIN_WORKERS).getAsInt();
-        maxWorkers = modelInfo.get(MAX_WORKERS).getAsInt();
-        maxBatchDelay = modelInfo.get(MAX_BATCH_DELAY).getAsInt();
-        responseTimeout = modelInfo.get(RESPONSE_TIMEOUT).getAsInt();
-        startupTimeout = modelInfo.get(STARTUP_TIMEOUT).getAsInt();
-        batchSize = modelInfo.get(BATCH_SIZE).getAsInt();
+        minWorkers =
+                modelInfo.has(MIN_WORKERS) && !modelInfo.get(MIN_WORKERS).isJsonNull()
+                        ? modelInfo.get(MIN_WORKERS).getAsInt()
+                        : 1; // default value for minWorkers
```
agunapal, line 196: Aren't the default values for these defined somewhere. Why are we hardcoding here?

agunapal: Also, shouldn't this be handled inside modelInfo.get() ?

agunapal (changes requested):

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
udaij12: They are private variables in modelConfig and modelinfo.get() is a json get function that just returns the value as json.

mreso: They are part of the model config but get overwritten by the value loaded from the yaml file. How about we store defaults separate in the model config as public constants, read the value to init the value but then they are available for later as well?

```
public class ModelConfig {
    public static final defaultStartupTimeout = 120; // unit: sec
    // ...
    /** the timeout in sec of a specific model's startup. */
    private int startupTimeout = defaultStartupTimeout; 

```

Then you should be able to do this here:
```
startupTimeout =
                modelInfo.has(STARTUP_TIMEOUT) && !modelInfo.get(STARTUP_TIMEOUT).isJsonNull()
                        ? modelInfo.get(STARTUP_TIMEOUT).getAsInt()
                        : modelArchive.getModelConfig().defaultStartupTimeout; // default value for startupTimeout
```
Then we would at least have a single source of truth.

udaij12: That sounds good... I did that for the four variables that have a default

agunapal: wondering whether default value of min_workers should be 1 ? or 0 ? By default a model can have no worker and we can use the mgmt API to increase this correct? To be safe , shall we just set default values for startup time & response time ?

mreso: To be honest I do not see the reason for min_worker being 0 and more than once I was frustrated as I did not set the parameter and then had to look up how to use the management API to increase the number of workers. Its also [not possible to set 0 inside](https://github.com/pytorch/serve/blob/master/frontend/archive/src/main/java/org/pytorch/serve/archive/model/ModelConfig.java#L284) the yaml file. You can only omit setting the parameter to achieve this state which is kind of unintuitive. @agunapal Do you see a use case for worker being 0? If someone does not want to serve the model they can remove it through management API.

udaij12: Done... only setting default values for startup and response time

agunapal: @mreso I'm not sure if there is a good reason. But this is something that I observed as well. I'm not sure why this was done.  and with this PR, if we set it to 1, we'll be changing the behavior which users are used to. We could change it in another PR in the next release.

agunapal (approved):
LGTM, pending nightly test

frontend/server/src/main/java/org/pytorch/serve/wlm/Model.java
mreso: Sounds good to me @agunapal

PR #3332, opened 2024-10-01
allow python version=3.12 in docker script build_image.sh

fabridamicelli:
## Description
Docker images are built using the script `build_image.sh` which takes the argument `pythonversion`.
Currently only versions 3.8, 3.9, 3.10 and 3.11 are allowed.
To the best of my knowledge there is no reason why we should not allow python 3.12 (please correct me if I am wrong).
This PR extends that list to include 3.12.

PS: If it is indeed the case that we should not allow 3.12, then it would be great to update the readme for consistency (currently only specifies `python >= 3.8`)

## Type of change

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,10 +65,10 @@ Refer to [torchserve docker](docker/README.md) for details.
 #### VLLM Engine
 ```bash
 # Make sure to install torchserve with pip or conda as described above and login with `huggingface-cli login`
-python -m ts.llm_launcher --model_id meta-llama/Meta-Llama-3.1-8B-Instruct --disable_token_auth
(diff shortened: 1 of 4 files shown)


fabridamicelli:
Hi @mreso 

any ideas/comments on this? :)
Since python 3.8 has reached end-of-life and 3.13 was released, it'd be great to be able to bump the versions in the docker scripts.
Thanks in advance
F

sapphire008:
Current numpy version 1.24.3 in the requirements.txt will prevent proper building in Python 3.12. Upgrading to 1.26.4 will solve the issue (ref: https://stackoverflow.com/questions/77364550/attributeerror-module-pkgutil-has-no-attribute-impimporter-did-you-mean)

PR #3382, opened 2025-01-22
Improve docker image UX and fix MNIST serving example

jurca:
## Description

Improve ease of use for the docker image and fix the MNIST serving example.

I haven't filed an issue, and this does not seem to be reported (or I missed it - sorry!).

## Type of change

I've tested the changes only manually, as they are very small.

## Checklist:

- [ ] Did you have fun?
  - Sorry, no, it has been frustrating finding out what's wrong and how to fix it - but that's why I created this PR, so that others and future me will instead have fun.
  - I think it might not be necessary in this case.

### DIFF
--- a/docker/config.properties
+++ b/docker/config.properties
@@ -3,6 +3,7 @@ management_address=http://0.0.0.0:8081
 metrics_address=http://0.0.0.0:8082
 grpc_inference_address=0.0.0.0
 grpc_management_address=0.0.0.0
+enable_envvars_config=true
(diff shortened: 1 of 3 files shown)


jurca:
Additionally, if the `tail -f /dev/null` in `docker/dockerd-entrypoint.sh` was present only to prevent container termination when torchserve was running as daemon in background, I would like to propose removing it completely, as I find it detrimental to UX when using custom command.

